PR #3, opened 2020-04-16
Drop things from the queue when completed

dramich:
Issue: https://github.com/rancher/rancher/issues/26633
https://github.com/rancher/rancher/issues/26577


dramich:
@ibuildthecloud Can you take a look at this? I can't request reviewers in this repo.

ibuildthecloud:
Do you have more context on why the enqueueAfter change?

dramich:
To try again, if you want to just mark the object as done regardless I can do that.

ibuildthecloud:
Ah, so the memory leak is that we don't ever call done? I don't think we want to reenque. I'd just always call done.

PR #6, opened 2020-07-24
[WIP] add serviceAccount accessStore support

aiwantaozi:


ibuildthecloud:
@aiwantaozi I haven't tested this PR but would https://github.com/rancher/steve/pull/7 work for you.  I think all that is required is that the user indexer dynamically builds the user.Name field and not just assume it is subject.Name.

aiwantaozi:
@ibuildthecloud Hi, I just try PR #7, user indexer works well, could access by user.Name and serviceaccount:namespace:name, thank you

PR #15, opened 2021-04-02
Add ability to support registryOverride

StrongMonkey:
Add ability to support cluster-level private registry support. https://github.com/rancher/rancher/issues/30735

### DIFF
--- a/pkg/podimpersonation/podimpersonation.go
+++ b/pkg/podimpersonation/podimpersonation.go
@@ -122,6 +122,7 @@ type PodOptions struct {
 	ConfigMapsToCreate []*v1.ConfigMap
 	SecretsToCreate    []*v1.Secret
 	Wait               bool
+	ImageOverride      string
 }
 
 // CreatePod will create a pod with a service account that impersonates as user. Corresponding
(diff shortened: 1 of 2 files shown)


pkg/podimpersonation/podimpersonation.go
```diff
@@ -131,7 +131,7 @@ type PodOptions struct {
 //      namespace used only for the purpose of running impersonated pods. This is to ensure
 //      proper protection for the service accounts created.
 //   2. The pod must KUBECONFIG env var set to where you expect the kubeconfig to reside
-func (s *PodImpersonation) CreatePod(ctx context.Context, user user.Info, pod *v1.Pod, podOptions *PodOptions) (*v1.Pod, error) {
+func (s *PodImpersonation) CreatePod(ctx context.Context, user user.Info, pod *v1.Pod, podOptions *PodOptions, registryOverride string) (*v1.Pod, error) {
```
ibuildthecloud, line 134: I think a better approach would be to just add ImageName as a field on PodOptions that if set it will use that instead of calling the `func() string`.

PR #20, opened 2021-06-21
[2.4] Add default api server and fix invalid open-api arbitrary type

aiwantaozi:
Problems:
1. allow steve to customize the api-ui response writer for the air-gap environment users
2. failed to view steve v1 API when its CRDs schemas contain the arbitrary schema type of the openAPI.

These problems already fixed in 2.5, pick [PR](https://github.com/rancher/steve/pull/14) from the master branch to fix problems for 2.4


guangbochen:
LGTM

PR #21, opened 2021-06-29
Don't cancel context of subscribe on method exit

smallteeths:
The ws channels created by `project-member` are cancelled early.

Reproduction steps

1. Create the cluster cluster1
2. Create the user test-user and add it as a member of cluster1 with the role ClusterMember, then add it to the Default project with the role ProjectMember
3. Log in with test-user and create a deployment from the command line. 
4. Navigate to the dashboard page, The ws connection with socketId = 2 will report a recurring error, as shown in the figure

Jietu20210625-140001

https://github.com/cnrancher/pandaria/issues/1549

### DIFF
--- a/pkg/server/store/proxy/rbac_store.go
+++ b/pkg/server/store/proxy/rbac_store.go
@@ -182,7 +182,6 @@ func (r *RBACStore) Watch(apiOp *types.APIRequest, schema *types.APISchema, w ty
 	}
 
 	ctx, cancel := context.WithCancel(apiOp.Context())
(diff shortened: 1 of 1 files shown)


aiwantaozi:
LGTM

guangbochen:
@ibuildthecloud can u PTAL, this is for Rancher 2.4 fix of an existing customer, thanks
A similar master-head fix is done with https://github.com/rancher/steve/commit/8b9bea44c6b26e39690c5b84443450db03fbf907

PR #45, opened 2022-04-05
Dependencies bumps

macedogm:
Signed-off-by: Guilherme Macedo <guilherme.macedo@suse.com>


.gitignore
```diff
@@ -5,3 +5,4 @@
 /dist
 *.swp
 .idea
+steve
```
maxsokolovsky, line 8: Why is this being ignored? Is this a directory with some generated/built artifacts?

macedogm: It's the build artifact when building locally.

maxsokolovsky (approved):

MbolotSuse (approved):
Looks like it builds and `go mod tidy` was run properly.

LGTM.

PR #47, opened 2022-04-13
[2.5] Prevent write on closed channel

jakefhyde:
Closing a channel in a separate goroutine than the one it is being written from can cause a write on closed channel panic. https://github.com/rancher/rancher/issues/37353


rmweir (approved):
LGTM

thedadams (approved):

PR #49, opened 2022-05-20
Make watches debuggable

cmurphy:
# Return websocket error and add logging for watches

Add debug logs and send websocket messages when the watch is closed
unexpectedly.

In addition to being helpful for debugging, the dashboard specifically
looks for a `resource.error` event containing the string "too old" in
order to trigger the watch to be resynced with a refreshed revision
number.  Without this error returned, the dashboard will only see
`resource.stop` events and never change its behavior, continuing to try
to restart the watch with an incorrect resource version.

# Make watch timeout configurable

By default, a watch times out after 30 minutes. For debugging purposes,
it's convenient if this can be decreased. Add an environment variable
CATTLE_WATCH_TIMEOUT_SECONDS to enable setting the timeout in seconds.

https://github.com/rancher/rancher/issues/37627

### DIFF
--- a/pkg/stores/proxy/proxy_store.go
+++ b/pkg/stores/proxy/proxy_store.go
@@ -318,6 +318,9 @@ func (s *Store) listAndWatch(apiOp *types.APIRequest, client dynamic.ResourceInt
 				obj, err := s.byID(apiOp, schema, rel.Namespace, rel.Name)
 				if err == nil {
 					result <- s.toAPIEvent(apiOp, schema, watch.Modified, obj)
+				} else {
+					logrus.Debugf("notifier watch error: %v", err)
+					returnErr(errors.Wrapf(err, "notifier watch error: %v", err), result)
 				}
 			}
 			return fmt.Errorf("closed")
@@ -327,6 +330,12 @@ func (s *Store) listAndWatch(apiOp *types.APIRequest, client dynamic.ResourceInt
 	eg.Go(func() error {
 		for event := range watcher.ResultChan() {
 			if event.Type == watch.Error {
+				if status, ok := event.Object.(*metav1.Status); ok {
+					logrus.Debugf("event watch error: %s", status.Message)
+					returnErr(fmt.Errorf("event watch error: %s", status.Message), result)
+				} else {
+					logrus.Debugf("event watch error: could not decode event object %T", event.Object)
+				}
 				continue
 			}
 			result <- s.toAPIEvent(apiOp, schema, event.Type, event.Object)
--- a/pkg/stores/proxy/proxy_store.go
+++ b/pkg/stores/proxy/proxy_store.go
@@ -7,8 +7,10 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"reflect"
 	"regexp"
+	"strconv"
 
 	"github.com/pkg/errors"
 	"github.com/rancher/apiserver/pkg/types"
@@ -31,6 +33,8 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 )
+
+const watchTimeoutEnv = "CATTLE_WATCH_TIMEOUT_SECONDS"
 
 var (
 	lowerChars  = regexp.MustCompile("[a-z]+")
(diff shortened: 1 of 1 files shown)


cbron (approved):
lgtm, interesting that there aren't even logged errors for many of these things already.

pkg/stores/proxy/proxy_store.go
```diff
@@ -327,6 +342,12 @@ func (s *Store) listAndWatch(apiOp *types.APIRequest, client dynamic.ResourceInt
 	eg.Go(func() error {
 		for event := range watcher.ResultChan() {
 			if event.Type == watch.Error {
+				if status, ok := event.Object.(*metav1.Status); ok {
+					logrus.Debugf("event watch error: %s", status.Message)
+					returnErr(errors.Wrapf(err, "event watch error: %+v", event), result)
```
Jason-ZW, line 347: There may be a situation where `err` is `nil`. If `err` is `nil` we don't get any useful error msg from the UI(websocket). It is best to add an `if` statement. When `err` is `nil`, we should use `errors.Errorf`.

Jason-ZW: In addition, we need to pay attention that  [rancher/apiserver](https://github.com/rancher/apiserver/blob/master/pkg/subscribe/watcher.go#L30)) will send `resource.stop` event notifications, combined with the current logic of [rancher/dashboard resource.error](https://github.com/rancher/dashboard/blob/master/shell/plugins/steve/subscribe.js#L437) & [rancher/dashboard resource.stop](https://github.com/rancher/dashboard/blob/master/shell/plugins/steve/subscribe.js#L460), possibly sending `resource.error` event will have other effects.

ly5156:
This pr will cause the backend to send error and stop events
1. As the UI will re-watch the resource (possibly with the wrong resourceVersion) 5 seconds after the stop event is fired: https://github.com/rancher/dashboard/blob/master/shell/plugins/steve/subscribe.js#L460

2.It will also determine if there is a `too old` string in the message in the error event, and if so it will resyncWatch the resource again: https://github.com/rancher/dashboard/blob/master/shell/plugins/steve/subscribe.js#L437

 So the above issue may not be resolved

rmweir (dismissed):
LGTM

cmurphy:
I've updated this to address the nil error issue that @Jason-ZW pointed out. I need more time to see how emitting both resource.stop and resource.error will behave with the UI. It may be that we should only emit one or the other, or change the UI to only respond to one.

cmurphy:
@Jason-ZW I was able to test out these changes again and found that with the error message properly emitted the dashboard successfully [calls resyncWatch](https://github.com/rancher/dashboard/blob/master/shell/plugins/steve/subscribe.js#L437) like you pointed out, and this actually seems to fix the issue on its own because the resource list is called again and the new watch uses the correct resource version from that fresh list. So it's possible we might not need a dashboard change.

rmweir (approved):
LGTM

Jason-ZW:
> @Jason-ZW I was able to test out these changes again and found that with the error message properly emitted the dashboard successfully [calls resyncWatch](https://github.com/rancher/dashboard/blob/master/shell/plugins/steve/subscribe.js#L437) like you pointed out, and this actually seems to fix the issue on its own because the resource list is called again and the new watch uses the correct resource version from that fresh list. So it's possible we might not need a dashboard change.

@ly5156 Cloud you please confirm this information?

ly5156:
> > @Jason-ZW I was able to test out these changes again and found that with the error message properly emitted the dashboard successfully [calls resyncWatch](https://github.com/rancher/dashboard/blob/master/shell/plugins/steve/subscribe.js#L437) like you pointed out, and this actually seems to fix the issue on its own because the resource list is called again and the new watch uses the correct resource version from that fresh list. So it's possible we might not need a dashboard change.
> 
> @ly5156 Cloud you please confirm this information?
LGTM
Because an error message is pushed, containing the `too old` character, it will trigger `resyncWatch`

Jason-ZW:
LGTM

cbron (dismissed):

cmurphy:
Updated one of the commit messages and PR description since we are now using this to fix the issue instead of just for debugging.

cbron (approved):

rmweir (approved):
LGTM

PR #50, opened 2022-05-25
v1.24.0

ebauman:
Bumped client-go and other packages to v1.24.0. 

Part of 1.24 appears to be a breaking change in Webhook setup (requiring a rest.Config now instead of kubeconfig string). 
I took the liberty of setting that client up in webhookcli. 

Please let me know if changes are needed, etc.


jiaqiluo (approved):
LGTM

kinarashah (approved):

PR #52, opened 2022-06-14
[2.5] Return websocket error and add logging for watches

cmurphy:
Add debug logs and send websocket messages when the watch is closed
unexpectedly.

In addition to being helpful for debugging, the dashboard specifically
looks for a `resource.error` event containing the string "too old" in
order to trigger the watch to be resynced with a refreshed revision
number.  Without this error returned, the dashboard will only see
`resource.stop` events and never change its behavior, continuing to try
to restart the watch with an incorrect resource version.

Backport of functional commit in https://github.com/rancher/steve/pull/49
https://github.com/rancher/rancher/issues/37986


rmweir (approved):
LGTM

PR #54, opened 2022-06-28
wait for svc account secret to be populated with token

kinarashah:


rosskirkpat (approved):

jiaqiluo (approved):

pkg/podimpersonation/podimpersonation.go
```diff
@@ -348,7 +348,19 @@ func (s *PodImpersonation) createPod(ctx context.Context, user user.Info, role *
 	if err != nil {
 		return nil, err
 	}
-
+	if _, ok := tokenSecret.Data["token"]; !ok {
+		for {
```
prachidamle, line 352: can use const v1.ServiceAccountTokenKey instead of "token"

pkg/podimpersonation/podimpersonation.go
```diff
@@ -348,7 +348,19 @@ func (s *PodImpersonation) createPod(ctx context.Context, user user.Info, role *
...
+			logrus.Debugf("wait for svc account secret to be populated with token %s", tokenSecret.Name)
+			time.Sleep(2 * time.Second)
```
prachidamle, line 354: Also wondering if we should keep the wait interval configurable across rancher - but something for later

kinarashah: yep maybe we could for rancher controllers, not sure if passing to steve would be easily possible

prachidamle (approved):
LGTM

pkg/podimpersonation/podimpersonation.go
rosskirkpat: If we could make the wait interval configure via an opt-in environment variable without adding the requirement for additional cycles, I believe that would be a great value add. That way, we take a "don't touch this unless you are told to touch this" approach

rosskirkpat (approved):

jiaqiluo (approved):

PR #55, opened 2022-07-29
Adding logic to limit number of cached schemas

MbolotSuse:
Adds logic to ensure that only one schema/access set
is cached for each user. This should improve memory
consumption

### DIFF
--- a/pkg/accesscontrol/access_store.go
+++ b/pkg/accesscontrol/access_store.go
@@ -14,6 +14,7 @@ import (
 
 type AccessSetLookup interface {
 	AccessFor(user user.Info) *AccessSet
+	PurgeUserData(id string)
 }
 
 type AccessStore struct {
@@ -61,6 +62,10 @@ func (l *AccessStore) AccessFor(user user.Info) *AccessSet {
 	}
 
 	return result
+}
+
+func (l *AccessStore) PurgeUserData(id string) {
+	l.cache.Remove(id)
 }
 
 func (l *AccessStore) CacheKey(user user.Info) string {
--- a/pkg/schema/collection.go
+++ b/pkg/schema/collection.go
@@ -36,6 +36,7 @@ type Collection struct {
 	byGVR      map[schema.GroupVersionResource]string
 	byGVK      map[schema.GroupVersionKind]string
 	cache      *cache.LRUExpireCache
+	userCache  *cache.LRUExpireCache
 	lock       sync.RWMutex
 
 	ctx     context.Context
@@ -84,6 +85,7 @@ func NewCollection(ctx context.Context, baseSchema *types.APISchemas, access acc
 		byGVR:      map[schema.GroupVersionResource]string{},
 		byGVK:      map[schema.GroupVersionKind]string{},
 		cache:      cache.NewLRUExpireCache(1000),
+		userCache:  cache.NewLRUExpireCache(1000),
 		notifiers:  map[int]func(){},
 		ctx:        ctx,
 		as:         access,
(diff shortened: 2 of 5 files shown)


MbolotSuse:
## Background 

Related to rancher/rancher#38330. When steve calculates schemas, it uses a cache to ensure that schemas are not being constantly re-calculated for each user. The key to the cache is based on the user's access set (with the value being the entire schema). This key is formed based on the user's permissions as defined through their role bindings and cluster role bindings. This means that if a user (or a group that the user is in) gets bound to a new permission, a new entry is added to the cache. However, the old entry (which is only valid for a very unique set of rolebindings/cluster role bindings) is still in the cache. This causes the cache to grow over time unnecessarily and can result in heavy memory usage since the schemas are such large objects.

## Changes
- Steve now maintains a separate LRU cache (same size as the schema cache + same eviction time) mapping the user id to the most recent access id
- When a user's schema is requested, we now check if the access id matches what we have in the cache. If it doesn't we purge the cache (the one used by access set and the one used by the schemas) of the old entry. This ensures that only one schema is maintained per user

## Notes
- Current memory leak is capped at 1000 entries, since that's the size on the LRU cache.

MbolotSuse:
Marked this as WIP since it needs some form of testing. Would like to get thoughts on this approach first before going through the effort to write tests for it.

cmurphy:
I think this is a good direction. Did you test this to ensure it fully fixes the memory issue? This code is the core of the issue for SURE-4188 as well. I tried something similar before and it did not fully fix that issue so I stopped pursuing it. This still does not completely fix it but it seems to improve it.

MbolotSuse:
@cmurphy I did test it out. Unfortunately, I think that you are correct that this doesn't fully fix the memory issue. 

However, it does improve the overall memory case and seemed to fix the memory issues that are based in this section of the code (i.e. the section related to schema creation/handling). A before/after pprof showed this section as using a comparable amount of memory before/after cluster creation and deletion (with a 5 min buffer after deletion) (available on request). I still think there's other items to fix, but this is seems to be a good self-contained change that fixes one of the issues in this area.

For larger setups (i.e. with a ton of users) I am a bit skeptical that this change will help - given that the LRU cache (even in the current state) only holds 1000 entries. I would expect that for setups with 1k+ users, this will use the same memory before/after the change (given that the cache seems to have a 24 hr ttl).

In short - tests say that this improves the overall issue and particularly the issue in this area. But it definitely doesn't fix everything, especially for large setups.

MbolotSuse:
Removing WIP - tests have been added. There are two commits here - one for tests, and one for the actual functionality.

rmweir (approved):
LGTM

pkg/accesscontrol/access_store.go
```diff
@@ -63,6 +64,10 @@ func (l *AccessStore) AccessFor(user user.Info) *AccessSet {
 	return result
 }
 
+func (l *AccessStore) PurgeUserData(id string) {
```
maxsokolovsky, line 67: What does this other access store cache store and why? And why is its size 50?

MbolotSuse: This caches a key representing the aggregate permissions given to a user to the access set created for that set of permissions. We expose a purge method for this so that when we have determined that a user's permissions have changed and that we should no longer cache the old permissions, we can purge that from the cache here as well.

I have no idea why this is size 50 - I can't think of a discernible reason for that number as opposed to another number.

maxsokolovsky (approved):

PR #56, opened 2022-08-05
Adding user authentication failed context value

MbolotSuse:


MbolotSuse:
Build has been cancelled for now until the PR for steve goes in. This will require steve to be updated to take advantages of the changes made there.

KevinJoiner (approved):
LGTM

pkg/auth/filter.go
```diff
@@ -153,6 +156,7 @@ func ToMiddleware(auth Authenticator) Middleware {
 						"system:cattle:error",
 					},
 				}
+				ctx = request.WithValue(ctx, CattleAuthFailed, "true")
 			} else if !ok {
 				info = &user.DefaultInfo{
```
cbron, line 161: So all this PR does is add a single key/val to the context if the authentication failed. But then below if authentication worked but user is unauthenticated we don't care, because they are using a rancher token ?

rmweir (approved):
LGTM

cbron (approved):

PR #58, opened 2022-08-24
Changing count watch to only return changed counts

MbolotSuse:
Related to rancher/rancher#36681.

## Overview

This changes the counts schema to only send counts when the count for a given schema has changed. This is to allow the UI to more easily handle changing counts for various types of resources. In addition, this changes the "debounce" logic to aggregate the individual resource count changes into an overall event. There was also a slight change in input/output for the debounce logic. 

## Changes
- Events now stream only changes for a specific schema, rather than returning the entire count
- Buffer logic now takes in an Channel of `Counts`, and converts that to an output Channel of type API Event
- Debounce logic now sends the first count that it sees immediately

### DIFF
--- a/pkg/resources/counts/buffer.go
+++ b/pkg/resources/counts/buffer.go
@@ -6,35 +6,60 @@ import (
 	"github.com/rancher/apiserver/pkg/types"
 )
 
-func buffer(c chan types.APIEvent) chan types.APIEvent {
+// debounceDuration determines how long events will be held before they are sent to the consumer
+var debounceDuration = 5 * time.Second
+
+// countsBuffer creates an APIEvent channel with a buffered response time (i.e. replies are only sent once every second)
+func countsBuffer(c chan Count) chan types.APIEvent {
 	result := make(chan types.APIEvent)
 	go func() {
 		defer close(result)
-		debounce(result, c)
+		debounceCounts(result, c)
 	}()
 	return result
 }
 
-func debounce(result, input chan types.APIEvent) {
-	t := time.NewTicker(time.Second)
+// debounceCounts converts counts from an input channel into an APIEvent, and updates the result channel at a reduced pace
+func debounceCounts(result chan types.APIEvent, input chan Count) {
+	// counts aren't a critical value. To avoid excess UI processing, only send updates after debounceDuration has elapsed
+	t := time.NewTicker(debounceDuration)
 	defer t.Stop()
 
-	var (
-		lastEvent *types.APIEvent
-	)
+	var currentCount *Count
+
+	firstCount, fOk := <-input
+	if fOk {
+		// send a count immediately or we will have to wait a second for the first update
+		result <- toAPIEvent(firstCount)
+	}
 	for {
 		select {
-		case event, ok := <-input:
-			if ok {
-				lastEvent = &event
-			} else {
+		case count, ok := <-input:
+			if !ok {
 				return
 			}
+			if currentCount == nil {
+				currentCount = &count
+			} else {
+				itemCounts := count.Counts
+				for id, itemCount := range itemCounts {
+					// our current count will be outdated in comparison with anything in the new events
+					currentCount.Counts[id] = itemCount
+				}
+			}
 		case <-t.C:
-			if lastEvent != nil {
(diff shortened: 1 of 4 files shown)


MbolotSuse:
Some additional notes here:
- Currently the UI is set up to expect us to send the full count of every resource each time. With this change in place, this does break the UI's sidebar which contains resource counts. The UI will need to adjust to apply the ups/downs in the resource and not remove the current count for resources which aren't mentioned. They've requested that I open an issue in rancher/dashboard (with the appropriate milestone) when this is merged so that they can adjust.
- rancher/rancher#36682 is also related to this. Since it's controlled by the buffer that I am editing here, I'm also going to adjust it here to 5 seconds, as requested by the UI team.

rmweir (approved):
LGTM

pkg/resources/counts/buffer.go
```diff
@@ -6,35 +6,61 @@ import (
...
+		debounceCounts(result, c)
 	}()
 	return result
 }
 
-func debounce(result, input chan types.APIEvent) {
-	t := time.NewTicker(time.Second)
+// debounceCounts converts counts from an input channel into an APIEvent, and updates the result channel at a reduced pace
+func debounceCounts(result chan types.APIEvent, input chan Count) {
+	// counts aren't a critical value. To avoid excess UI processing, only send updates every 5 seconds
```
KevinJoiner, line 24: Nit: 
In case debounceDuration is changed in the future.
```suggestion
	// counts aren't a critical value. To avoid excess UI processing, only send updates after debounceDuration has elapsed.

File: pkg/resources/counts/buffer.go [hunk 2]
```diff
@@ -6,35 +6,61 @@ import (
...
+	count, ok := <-input
+	if ok {
+		// send a count immediately or we will have to wait a second for the first update
+		result <- toAPIEvent(count)
+	}
 	for {
 		select {
-		case event, ok := <-input:
+		case count, ok := <-input:
 			if ok {
```
KevinJoiner, line 38: Nit: 
Early return to flatten things out. 
```suggestion
			if !ok {
			    return
			 }
```

pkg/resources/counts/buffer_test.go
```diff
@@ -0,0 +1,95 @@
...
+			debounceDuration = 100 * time.Millisecond
+			countsChannel := make(chan Count, 100)
+			outputChannel := countsBuffer(countsChannel)
+
+			// first event is not buffered, so get it out of the way to test the debounce
+			countsChannel <- Count{
+				ID:     "count",
+				Counts: map[string]ItemCount{"test": createItemCount(1)},
+			}
+			<-outputChannel
```
KevinJoiner, line 44: Nit: 
You could use `receiveWithTimeout` here to verify that the first result is returned immediately. This would be primarily to stop someone from introducing a regression if buffer is ever updated.

KevinJoiner (approved):
I like the solution of adding an aggregation of all changed counts.

Nit:
If you are able can you add some comments to `counts.go`

pkg/resources/counts/buffer_test.go
MbolotSuse: I'll have to re-implement it since this in a different package than counts_test.go (I can test public interfaces with the other one but not with this one), but I think that's ok since it's just a helper/utility method.

pkg/resources/counts/buffer.go
MbolotSuse: Done - good call.

pkg/resources/counts/buffer.go
MbolotSuse: Done Good call.

pkg/resources/counts/buffer_test.go
MbolotSuse: Re-implemented this. I've also changed the other part of the test to use this method too.

MbolotSuse:
@KevinJoiner I've added comments where I felt that I had an adequate understanding to do so (pretty minor, but it's something).

MbolotSuse:
Note: Don't merge this until after 2.7.0. This will break the UI, and we need to coordinate so that we don't ship a broken rancher.

PR #59, opened 2022-10-13
Steve Pipeline, Go Version updates

MbolotSuse:
This PR aims to add some better CI for steve (test and validate in CI phases), and to update go to version 1.19.


MbolotSuse:
## Overview

Bumps steve's go version to 1.19, adds a pipeline for build/test/lint, and fixes a CI bug causing fossa to be run on every PR.

## Changes
- Bumps the version in go.mod to 1.19
- Changes the build image used in the CI to use sle-bci golang 1.19
- Changes the run image used in steve's image to use bci-micro instead of alpine
- Adds a new `build-bin` make command which builds the steve binary
- Adds a new `test` make command which tests all go files
- Adds a new `validate` make command which runs golangci-lint
- Adds a new .golangci.json file which provides configuration for golangci-lint (mostly copied from rancher/rancher, see notes)
- Adds a new CI phase which runs the `make build-bin` command
- Adds a new CI phase which runs the `make test` command
- Adds a new CI phase which runs the `make validate` command after installing dependencies
- Fixes small linter issues in a variety of go files which were causing `make validate` to fail

## Notes
- The golangci-lint config file (`.golangci.json`) is mostly the same as the one in `rancher/rancher`. The one difference is that it excludes the lint directive from revive which fails when you have an exported function which stutters - like `request.RequestForm` (since it has request in package name and in name of the function). I couldn't fix these issues because it would change exported functions, which is a break in behavior and would require a major release. So I turned off the validation for these until we get steve in a place where this is possible.

scripts/build-bin.sh
```diff
@@ -0,0 +1 @@
+CGO_ENABLED=0 go build -ldflags "-extldflags -static -s" -o /steve
```
cmurphy, line 1: I wouldn't be able to run this directly without root. Could we create a `bin/` directory and do `-o ./bin/steve` instead? The docker container can copy it to `/` if it needs to.

cmurphy: Should probably include the shebang at the top

scripts/validate.sh
```diff
@@ -0,0 +1,15 @@
...
+golangci-lint run
+go mod tidy
+go mod verify
+if unclean=$(git status --porcelain --untracked-files=no); then
+  if [ -n "$unclean" ]; then
+    echo "Encountered dirty repo!";
+    echo "$unclean";
+    exit 1;
+  fi
+else
```
cmurphy, line 12: You could use `set -e` at the top instead of adding an else condition for this

pkg/server/handler/handlers.go
```diff
@@ -9,11 +9,6 @@ import (
 
 func k8sAPI(sf schema.Factory, apiOp *types.APIRequest) {
 	vars := mux.Vars(apiOp.Request)
-	group := vars["group"]
-	if group == "core" {
-		group = ""
-	}
```
cmurphy, line 15: Seems like a significant functionality change? Did this fail the linter?

cmurphy (commented):
> we will now log a couple of errors that we would have otherwise completely ignored. I think this is ok, but am open to discussing it.

I didn't catch this anywhere in the Go code, could you point it out to me?

pkg/server/handler/handlers.go
MbolotSuse: It did. Group actually isn't used anywhere - we get it, override it, and then never use it.

MbolotSuse: I'm open to being wrong and having missed something - if you see somewhere it is being used LMK and I can revert.

cmurphy: You're right, I missed that it wasn't being used. It was dropped a while ago.

scripts/validate.sh
MbolotSuse: Just tried this - I see the error code but the exit code is still 0. Not sure if I am doing something wrong, but I need the exit code to persist too so that the stage fails. 

For reference, I added `set -e` to the top of the file, then I change the `git status` command to `g status` (to produce a command not found error). I saw the error message, but `echo $?` reported a 0 exit code - which makes me think that this doesn't work as desired. Did I miss something?

scripts/build-bin.sh
MbolotSuse: Good call, will adjust. I was a bit lazy in how I copied this from docker.

MbolotSuse:
> I didn't catch this anywhere in the Go code, could you point it out to me?

Sorry, I misremembered. I originally had other linter fixes which didn't make them in (the modified configuration from rancher apparently doesn't look for ignored errors). Since the CI doesn't care about those, I didn't end up pushing them to reduce blast radius of the changes. I'll update my comment to be more accurate - sorry for the misinformation.

scripts/build-bin.sh
MbolotSuse: Change made - thanks for the suggestion.

scripts/validate.sh
cmurphy: You could use the extra variable like you're currently doing
```
$ cat set-e.sh 
#!/bin/bash
set -e
unclean=$(false && echo 'hello')
if [ -n "$unclean" ] ; then
    echo 'Encountered dirty repo!'
    exit 1
fi
$ bash -x set-e.sh 
+ set -e
++ false
+ unclean=
$ echo $?
1
$ sed -i -e 's/false/true/' set-e.sh
$ bash -x set-e.sh 
+ set -e
++ true
++ echo hello
+ unclean=hello
+ '[' -n hello ']'
+ echo 'Encountered dirty repo!'
Encountered dirty repo!
+ exit 1
$ echo $?
1
```
Is the presence of `git` the specific worry? What about the presence of `go` or `golangci-lint`? It doesn't seem to be something we're concerned about in rancher/rancher's validate script. You could use `which git` to check before proceeding. `set -e` should still be used in any case to halt execution and return an error if any of the linters or `go mod` commands fail.

MbolotSuse: > Is the presence of git the specific worry?

No that part isn't a big deal, I install all of the dependencies in-pipeline in before using the validate. When testing on my local, I had an issue with git where the command exited because it couldn't determine repo ownership or some such (was because I had the directory mounted as a volume). I realized that if this command failed for whatever reason, we would pass the CI. We've had issues like this before - and it created scenarios where people were able to merge in files without running go generate and the like which caused further issues for others down the road (can point you to such a scenario if you are curious). 

I thought I would take a shot at getting a more robust script together that protects us from such issues. I'll try what you've put above - thank you for elaborating.

MbolotSuse: Change made, thank you for helping me with that.

cmurphy (approved):

pkg/stores/partition/parallel.go
```diff
@@ -170,17 +170,16 @@ func (p *ParallelPartitionLister) feeder(ctx context.Context, state listState, l
...
-					capacity -= len(list.Objects)
-					if list.Continue == "" {
-						return nil
-					}
-					// loop again and get more data
-					state.Continue = list.Continue
-					state.PartitionName = partition.Name()
-					state.Offset = 0
 				}
+				result <- list.Objects
```
KevinJoiner, line 174: ❤️

pkg/stores/proxy/rbac_store.go
```diff
@@ -166,14 +166,13 @@ func isPassthroughUnconstrained(apiOp *types.APIRequest, schema *types.APISchema
...
 			return nil, true
-		} else {
-			return []partition.Partition{
-				Partition{
-					Namespace: apiOp.Namespace,
-					Names:     resources[apiOp.Namespace].Names,
-				},
-			}, false
 		}
+		return []partition.Partition{
```
KevinJoiner, line 170: ❤️

KevinJoiner (approved):
Big fan of the linter changes.

PR #61, opened 2022-10-20
Updates Schemas watch logic.

KevinJoiner:
### Problem 1
For some schemas, Steve would send a change event even though an object did not change.
Steve already performed a deep equals check for changes to each schema since the last event [here](https://github.com/rancher/steve/blob/06584d207dac89091e5601884282bf7e109e19f0/pkg/resources/schemas/template.go#L109).
This check failed for schemas that have an access attribute set because when the new scheme are converted to apiObjects the access attribute is nationally stripped off [here](https://github.com/rancher/apiserver/blob/fe12f2183e7112becec1633eb2ea728a09cece47/pkg/store/schema/schema_store.go#L21). 

The fix for this also removes the access attribute from the old schema that we are comparing against.

### Problem 2
While testing, it was discovered schema updates were triggered on both user access changes and schema onChange calls. These two notification triggers were performed in separate go routines that would duplicate change events since the last sent state was not shared between the two routines.  This separation of notification also caused the stream to continue even if one of the two routines returns early due to an error.

Both notifications are now checked in the same go routine, and shemaState is shared to avoid these duplicate updates.

### Other Changes
Code comments were added.
generated mocks were created using go-mock for testing.


pkg/resources/schemas/template.go
```diff
@@ -42,82 +45,108 @@ type Store struct {
...
 		return oldSchemas
 	}
 
 	inNewSchemas := map[string]bool{}
-	for _, apiObject := range schemastore.FilterSchemas(apiOp, schemas.Schemas).Objects {
+
+	// Convert the schemas for the given user to a flat list of APIObjects.
+	apiObjects := schemastore.FilterSchemas(apiOp, schemas.Schemas).Objects
+	for i := range apiObjects {
+		apiObject := apiObjects[i]
```
cmurphy, line 118: This might be a nitpick but wouldn't it be more idiomatic to
```suggestion
	for _, apiObject := range apiObjects {
```

pkg/resources/schemas/template.go
```diff
@@ -42,82 +45,108 @@ type Store struct {
...
 			ResourceType: "schema",
 			Object:       apiObject,
 		}
 	}
 
-	for _, oldSchema := range schemastore.FilterSchemas(apiOp, oldSchemas.Schemas).Objects {
+	// Identify all of the oldSchema APIObjects that have been removed and send Remove APIEvents.
+	oldSchemaObjs := schemastore.FilterSchemas(apiOp, oldSchemas.Schemas).Objects
+	for i := range oldSchemaObjs {
+		oldSchema := oldSchemaObjs[i]
```
cmurphy, line 149: same comment

pkg/resources/schemas/template_test.go
```diff
@@ -0,0 +1,375 @@
...
+				})
+
+				expectedEvents := []types.APIEvent{}
+				return testValues{expectedEvents, factory, eventsReady}
+			},
+		},
+	}
+
+	for i := range tests {
+		test := tests[i]
```
cmurphy, line 292:
```suggestion
	for _, test := range tests {
```

pkg/resources/schemas/template_test.go
```diff
@@ -0,0 +1,375 @@
...
+				updateSchemas.AddSchema(testSchema)
+
+				// return updated schemas for the second request
+				factory.EXPECT().Schemas(userInfo).DoAndReturn(func(_ user.Info) (*types.APISchemas, error) {
+					// signal that schemas were requested
+					close(eventsReady)
+					return updateSchemas, nil
+				})
+
+				expectedEvents := []types.APIEvent{}
```
cmurphy, line 285: I'm surprised by this, can you help me understand why this is expected behavior? `schemasForSubject` is all about regenerating the schemas if the user's access changed. If the user lost access to an API type, I would expect the schemas to need a refresh.

pkg/resources/schemas/template_test.go
```diff
@@ -0,0 +1,375 @@
...
+				time.Sleep(setupTimeout)
+			case <-time.After(setupTimeout):
+				// When we continue here then the test will fail due to missing mock calls not being called.
+			}
+
+			// verify correct results are sent
+			hasExpectedResults(t, values.expectedChanges, resultChan, setupTimeout)
+		})
+	}
+}
```
cmurphy, line 341: Would it be reasonable to add a test for Problem 2 checking for duplicate notifications?

pkg/resources/schemas/template.go
KevinJoiner: The suggested way is easier to read, but some scenarios cause unintended behavior. For instance, in this case, `apiObjects` is a list of `types.APIObject` not `*types.APIObject`, so every iteration using the suggested range format would make a new shallow copy at the start of the loop of `types.APIObject` instead of just referencing the existing object.

If you are interested, there is currently some discussion of changing the behavior of range loops because of these idiosyncrasies. 
https://github.com/golang/go/discussions/56010

pkg/resources/schemas/template_test.go
KevinJoiner: This is similar to my comment above, except the issue with using `_, test` here is that we are using t.Parallel(), which causes a race condition because the same `test` variable is scoped to the loop, not each iteration. The alternative is to add `test := test` at the top of the loop, but I don't like shadowing previously declared variables.

pkg/resources/schemas/template_test.go
KevinJoiner: This was the cause of the original bug. When we create the schemas with `schema.Factory.Schemas(user)` we add the specific verbs a user has on each schema in the attributes section of the schema under the key `access` [here](https://github.com/rancher/steve/blob/06584d207dac89091e5601884282bf7e109e19f0/pkg/schema/factory.go#L134).
When returning each schema as an APIEvent over the websocket we must first convert the Schema into an APIObject in apiserver [here](https://github.com/rancher/apiserver/blob/fe12f2183e7112becec1633eb2ea728a09cece47/pkg/store/schema/schema_store.go#L19). For reason I am not positive on we deliberately strip off the `access` key from the attributes during this conversion. Thus when Steve was comparing the APIObject schema and the previously recorded Schema one would be missing the `access` verbs and a change would always be detected.

I ended up stripping the `access` key out of attributes on the old Schema as well before the comparison so we could get accurate change detection. I strayed away from removing the delete during conversion in apiserver for 2 reasons.
1. I am not sure if Steve is the only consumer of `apiserver.APISchemas`
2. The events sent to the UI previously did not include the `access` verbs in the attributes, so I didn't want to add new data that may not be used.

pkg/resources/schemas/template_test.go
KevinJoiner: Yeah, I think those would be good tests for me to add.

pkg/resources/schemas/template.go
KevinJoiner: Same as above to prevent unneeded copies of the object.

pkg/resources/schemas/template_test.go
KevinJoiner: Also, if the user's access changes, we regenerate the available schemas for the user via `userChangeNotify` in `template.go`. The above test just demonstrates that we do not send the verbs that changed for each schema.

KevinJoiner:
I added another commit that runs go generate in the drone validate step so that our generated mocks stay up to date.

pkg/resources/schemas/template_test.go
cmurphy: Spent some time playing with this. It looks like if I add or remove a role for a user, the `resource.create` and `resource.remove` events go through (because there's nothing to call DeepEqual on). If I change the role's verbs from, e.g. [create, update, get, list] to just [get, list], the `resource.change` event also goes through because the object's ResourceMethods and CollectionMethods change and the DeepEqual returns false. On the other hand, if I change it from [get, list] to just [get], the methods do not change, so no event goes through. But this seems fine, since as you pointed out the UI isn't using the access verbs anyways. The schema change event doesn't affect the user's actual access level. So I think this makes sense :+1:

cmurphy (approved):

maxsokolovsky (approved):

PR #62, opened 2022-10-25
Reducing number of generated schemas

MbolotSuse:
Changes steve to only return top level schema definitions rather than a full defined schema for every field


MbolotSuse:
Related to rancher/rancher#38468 - attempts to address the large size of schemas. Still needs tests.

MbolotSuse:
## Overview

Steve generates schemas today in an attempt to provide a full api definition. This approach is flawed for the reasons outlined in [this comment](https://github.com/rancher/rancher/issues/38468#issuecomment-1289463085) (basically, it produces poor performance and incomplete information). This PR aims to change the schemas so that they are only generated for top-level resources (meaning, there's only a schema for an object like `Pods` that you can create, and not a schema for a sub-field like `ObjectMetadata` which is only a member of other objects that you can actually create). 

## Changes
- Changes steve to not recursively produce schemas for objects
- Adds comments to several functions in `pkg/schema/converter` (main package of this PR - responsible for producing much of the schemas)
- Fixes a bug where we used the openAPI schema for the first version of a CRD for all versions of that CRD
- Removes parameters which are no longer used from several functions
- Adds tests for the `ToSchemas` function of `k8stonorman.go`
- Adds tests for the `AddCustomResources` function of `crd.go`
- Adds tests for the `AddDiscovery` function of `discovery.go`

MbolotSuse:
@cmurphy 

> But the speed of the request, when we ignore the time needed to render the result, is only slightly improved

I'm not sure that I agree with you here. The data that you have provided indicates a 30% speedup. That's non-trivial, and if it holds in environments with a worse network connection (i.e. where the bigger schemas are more of a problem and take longer to retrieve), then I think it is valuable.

>  So my question is, does the dashboard actually need to render this schema, apart from when we actually hit /v1/schemas in the browser?

Yes - the UI does need this endpoint for parts of its functionality. The UI uses these resources to control button visibility. For example, who has access to see the "Create Namespace" button by a project? Just because a user can see the project doesn't mean they can see the button (i.e. users who have read-only access to the project). That's one example - in general, the UI uses the resourceVerbs and collectionVerbs parts of the schema (as well as the existence of the schema) to check if the user can access a given resource (and if so, how they can access the resource).

> Is this slight speed improvement actually impactful for day-to-day dashboard use?

Overall impact is tough to judge but I believe so. The UI loads the schemas as part of it's base functionality - if you hit refresh and look at the console, you can see that it hits the schemas endpoint (to load these in and control element visibility). 

I think this matters more for environments that have degraded performance (i.e. rancher behind a proxy, slow network connection), where the larger data of the original schemas could have a larger impact. I think that the UI is also storing these resources on their side, so reducing the volume might help with memory/performance outside of the original request (as a note, we also cache these resources on the steve side).

> I also noticed that the new schema list removes the [status subresources](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource) from the list. The status subresource can have its own access control separate from the main resource, noted [here](https://kubernetes.io/docs/reference/using-api/api-concepts/):

Thank you for pointing that out, I did not know about that. I will look at how steve handled this before the change. If these resources did include verbs, I will adjust my solution to include them. If they did not, I won't make any changes here, but I will get a new issue to include these sub-resources as part of the schema in the future.

pkg/schema/converter/k8stonorman.go
```diff
@@ -38,13 +40,11 @@ func GVRToPluralName(gvr schema.GroupVersionResource) string {
 	return fmt.Sprintf("%s.%s", gvr.Group, gvr.Resource)
 }
 
+// ToSchemas creates the schemas for a K8s server, using client to discover groups/resources, and crd to potentially
+// add additional information about new fields/resources. Mostly ties together AddDiscovery and AddCustomResources
 func ToSchemas(crd v1.CustomResourceDefinitionClient, client discovery.DiscoveryInterface) (map[string]*types.APISchema, error) {
 	result := map[string]*types.APISchema{}
 
-	if err := AddOpenAPI(client, result); err != nil {
```
cmurphy, line 44: Can you clarify why this was removed - I guess so that resourceFields would be empty? It looks like now the description is also missing because of this, that seems like a slight drawback.

pkg/schema/converter/crd_test.go
```diff
@@ -0,0 +1,459 @@
...
+										JSONPath: "TestPath",
+										Type:     "TestType",
+										Format:   "TestFormat",
+									},
+								},
+								Schema: &v1.CustomResourceValidation{
+									OpenAPIV3Schema: &v1.JSONSchemaProps{
+										Description: "Test Resource for unit tests",
+										Required:    []string{"required"},
+										Properties: map[string]v1.JSONSchemaProps{
```
cmurphy, line 54: Are all these properties required for testing, or could they be condensed into one or two for the sake of readability?

pkg/schema/converter/crd.go
```diff
@@ -27,6 +27,8 @@ var (
 	}
 )
 
+// AddCustomResources uses the openAPISchema defined on CRDs to provide field definitions to previously discovered schemas
+// Note that this function does not create new schemas - it only adds details to resources already present in the schemas map
```
cmurphy, line 31: (nit)
```suggestion
// AddCustomResources uses the openAPISchema defined on CRDs to provide field definitions to previously discovered schemas.
// Note that this function does not create new schemas - it only adds details to resources already present in the schemas map.
```

pkg/schema/converter/discovery.go
```diff
@@ -22,6 +22,8 @@ var (
 	}
 )
 
+// AddDiscovery uses a k8s discovery client to create very basic schemas for all registered groups/resources. Other
+// functions, such as AddCustomResources are used to add more details to these schemas later on
```
cmurphy, line 26: (nit)
```suggestion
// functions, such as AddCustomResources are used to add more details to these schemas later on.
```

pkg/schema/converter/k8stonorman.go
```diff
@@ -1,3 +1,5 @@
+// Package converter is responsible for converting the types registered with a k8s server to schemas which can be used
+// by the UI (and other consumers) to get a rough idea of the resources available and their current user's permissions
```
cmurphy, line 2:
```suggestion
// Package converter is responsible for converting the types registered with a k8s server to schemas which can be used
// by the UI (and other consumers) to discover the resources available and their current user's permissions.
```

pkg/schema/converter/k8stonorman.go
```diff
@@ -38,13 +40,11 @@ func GVRToPluralName(gvr schema.GroupVersionResource) string {
 	return fmt.Sprintf("%s.%s", gvr.Group, gvr.Resource)
 }
 
+// ToSchemas creates the schemas for a K8s server, using client to discover groups/resources, and crd to potentially
+// add additional information about new fields/resources. Mostly ties together AddDiscovery and AddCustomResources
```
cmurphy, line 44: (nit)
```suggestion
// ToSchemas creates the schemas for a K8s server, using client to discover groups/resources, and crd to potentially
// add additional information about new fields/resources. Mostly ties together AddDiscovery and AddCustomResources.
```

cmurphy (commented):
>> So my question is, does the dashboard actually need to render this schema, apart from when we actually hit /v1/schemas in the browser?

> Yes - the UI does need this endpoint for parts of its functionality.

Sorry for being unclear in my question, I was actually wondering about whether it needs to *render*, as in format and display the data to the user, not about whether it uses the attributes in the schema. I did some testing in the browser and found that the actual start-to-finish time of loading /v1/schemas (meaning, navigating to that path rather than loading it as a byproduct of a regular dashboard page) takes around 4 seconds, and it looks like it's because there's so much content that it takes the UI javascript around 3 seconds to finish processing it after it receives it:

image

so my thinking was that as long as this was not the typical use case, then receiving so much data may not be much of a problem. But now I'm seeing that there is a noticeable difference in the time it takes to download the reduced size content regardless of rendering it - around 130ms compared to around 220ms previously - so it does appear to be providing value to the browser.

> Because of this, I would say that steve does not currently support subresources.

It looks like you're right, I can't seem to get the PodStatus schema to add method verbs even by adding a role and rolebinding for pods/status, so I guess we shouldn't hold this up if we're not using it anyway.

Generally fine with this change, left a couple of questions and minor suggestions inline.

pkg/schema/converter/k8stonorman.go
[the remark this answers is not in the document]
MbolotSuse: Good call, I'll fix that.

pkg/schema/converter/k8stonorman.go
MbolotSuse: This part mostly gave recursive definition for the core types from what I can see on testing (i.e. what types are in a pod spec).  Since we no longer need to define those values, I think it's safe to remove, though I can look at it again if there's anything that you think it had that we need (I can also send over a with/without diff that may help if you want to see what fields are missing in it).

As far as the description goes, it looks like (from hitting `/openapi/v2` directly) our core types wouldn't have a description anyway - near as I can tell both the well-defined provisioning.cattle.io cluster and the less well-defined management.cattle.io cluster objects are missing a description field. I guess this does leave out these fields for the core types, but since you can use k8s for this information I don't think this is a large loss. I agree with you that in an optimal world this would be there for us to consume later, but I'd guess that we can pick that up from the CRD definition (which we call separately), so I'd prefer to look there. 

I can re-add and adjust this, but I would ask that we consider if that type of a change (and writing tests for all of that) is worth it just for the description field.

pkg/schema/converter/crd_test.go
MbolotSuse: I believe that each of these covers a test case for [toResourceField](https://github.com/rancher/steve/blob/1061dd34560d66e31d1d028d7278a6efb6d251c0/pkg/schema/converter/openapiv3.go#L45). I'd like to keep them in here since it asserts on the behavior we expect from all of our types. I could try and restructure these test cases so that there's one that does assertions only on properties, but I think that's a fair amount of changes for test readability. 

So in short, I do think that these are required for full test coverage. I can try and refactor them if you think it's worth it.

KevinJoiner (dismissed):
Looks good with Colleen's punctuation changes.

pkg/schema/converter/k8stonorman.go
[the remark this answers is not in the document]
MbolotSuse: Fix made.

pkg/schema/converter/crd.go
MbolotSuse: Thanks, fix made.

pkg/schema/converter/discovery.go
MbolotSuse: Fix made.

pkg/schema/converter/k8stonorman.go
MbolotSuse: Added this fix and made another, related comment change.

pkg/schema/converter/k8stonorman.go
MbolotSuse: Fix made, thanks.

cmurphy (dismissed):

pkg/schema/converter/k8stonorman.go
MbolotSuse: Took another look at this. While most of this does appear to be the recursive core type definition, there is at least one resource that concerns me - the `apigroups` resource. This resource serves to list all of the apigroups available on the server (like `core` or `apps`) and their associated versions. It looks like most of the logic which generates this resource is in another place as a template - see [here](https://github.com/rancher/steve/blob/a10fe811f58ff7d92ffee51dd3f11010afbaf115/pkg/resources/apigroups/apigroup.go#L14). However, this template won't get applied, and the resource won't be created unless a schema exists, see [here](https://github.com/rancher/steve/blob/a10fe811f58ff7d92ffee51dd3f11010afbaf115/pkg/schema/collection.go#L115-L126). I'm unsure if the UI uses this value, but to me it would seem safer to include it rather than exclude it and hope everything goes ok.

In short, I missed something with one resource, and I'll see if I can fix it by re-introducing the top level resources here as suggested.

KevinJoiner: I looked at this with Micahel, and `apigroups` is not a Kubernetes Resource Type. 
It is just distilled information that can be obtained by parsing the output of `/v1/schemas`

We should not include this extra functionality if the following statements stand true.
1. We desire to keep Steve as lightweight as possible. (AKA only add necessary functionality).
2. The endpoint is not currently needed by the UI.

MbolotSuse:
Closing this in favor of a new PR, since this one is quite out of date.

PR #63, opened 2022-10-27
Pagination, sorting, filtering

cmurphy:
Commit summary:

### Refactor proxy store

Filtering and sorting needs to operate on unstructured data. It also
needs to operate after the parallel partitioner, higher in the store
stack. This means that the proxy Store needs to return raw, unstructured
data up to the partitioner. This change moves all conversions from
unstructured Kubernetes types to apiserver objects up from the proxy
store into the partitioner.

### Add filtering to partition store

Extend the partition store to parse query params as list filters. Filter
keys use dot notation to denote the subfield of an object to filter on.
Example:

GET /v1/secrets?filter=metadata.name=foo

Filters are ANDed together, so an object must match all filters to be
included in the list. Example:

GET /v1/secrets?filter=metadata.name=foo&filter=metadata.namespace=bar

Arrays are searched for matching items. If any item in the array
matches, the item is included in the list. Example:

GET /v1/pods?filter=spec.containers.image=alpine

### Add sorting to partition store

Extend the partition store to parse the "sort" query parameter as a
sorting condition. Dot notation is used to denote the object field.
Preceding the key with "-" denotes descending (reverse) order.

Example sorting by name:

GET /v1/secrets?sort=metadata.name

Reverse sorting by name:

GET /v1/secrets?sort=-metadata.name

All values are converted to strings and sorted lexicographically.

### Add pagination to partition store

Extend the partition store to parse page and pagesize parameters and
return a subset of list results. The total number of pages is included
in the list response.

Request an initial page:

GET /v1/secrets?pagesize=10

or

GET /v1/secrets?pagesize=10&page=1

Request subsequent pages, or arbitrary pages:

GET /v1/secrets?pagesize=10&page=37

If a page number is out of bounds, an empty list is returned.

### Add caching to pagination

Cache filtered, sorted results for fast subsequent page retrieval.

Requests for cached queries need to include the list revision number
along with other queries. If no specific revision is requested, a new
fetch is done in order to get the latest revision. The revision is
included in the list response.

Example first request:

GET /v1/secrets?pagesize=10

Example subsequent page request:

GET /v1/secrets?pagesize=10&page=1&revision=107740

### Add secondary sort parameter

Extend the sorting functionality in the partition store to support
primary and secondary sorting criteria. Sorting parameters are specified
by a single 'sort' query and comma separated. Example:

Sort by name and creation time:

GET /v1/secrets?sort=metadata.name,metadata.creationTimestamp

Reverse sort by name, normal sort by creation time:

GET /v1/secrets?sort=-metadata.name,metadata.creationTimestamp

Normal sort by name, reverse sort by creation time:

GET /v1/secrets?sort=metadata.name,-metadata.creationTimestamp

### Add README
Add an initial README to explain the basics of the API and the new
parameters.

Issue: https://github.com/rancher/rancher/issues/38427

~Depends on https://github.com/rancher/steve/pull/60~
Depends on https://github.com/rancher/steve/pull/63

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,153 @@
+steve
+=====
+
+Steve is a lightweight API proxy for Kubernetes whose aim is to create an
+interface layer suitable for dashboards to efficiently interact with
+Kubernetes.
+
+API Usage
+---------
+
+### Kubernetes proxy
+
+Requests made to `/api`, `/api/*`, `/apis/*`, `/openapi/*` and `/version` will
+be proxied directly to Kubernetes.
+
+### /v1 API
+
+Steve registers all Kubernetes resources as schemas in the /v1 API. Any
+endpoint can support methods GET, POST, PATCH, PUT, or DELETE, depending on
+what the underlying Kubernetes endpoint supports and the user's permissions.
+
+* `/v1/{type}` - all cluster-scoped resources OR all resources in all
+  namespaces of type `{type}` that the user has access to
+* `/v1/{type}/{name}` - cluster-scoped resource of type `{type}` and unique name `{name}`
+* `/v1/{type}/{namespace}` - all resources of type `{type}` under namespace `{namespace}`
+* `/v1/{type}/{namespace}/{name}` - resource of type `{type}` under namespace
+  `{namespace}` with name `{name}` unique within the namespace
+
+### Query parameters
+
+Steve supports query parameters to perform actions or process data on top of
+what Kubernetes supports.
+
+#### `link`
+
+Trigger a link handler, which is registered with the schema. Examples are
+calling the shell for a cluster, or following logs during cluster or catalog
+operations:
+
+```
+GET /v1/management.cattle.io.clusters/local?link=log
+```
+
+#### `action`
+
+Trigger an action handler, which is registered with the schema. Examples are
+generating a kubeconfig for a cluster, or installing an app from a catalog:
+
+```
+POST /v1/catalog.cattle.io.clusterrepos/rancher-partner-charts?action=install
+```
+
+#### `filter`
+
+Only applicable to list requests (`/v1/{type}` and `/v1/{type}/{namespace}`).
+
+Filter results by a designated field. Filter keys use dot notation to denote
+the subfield of an object to filter on. The filter value is matched as a
+substring.
+
+Example, filtering by object name:
+
+```
+/v1/{type}?filter=metadata.name=foo
+```
+
+Filters are ANDed together, so an object must match all filters to be
+included in the list.
+
+```
+/v1/{type}?filter=metadata.name=foo&filter=metadata.namespace=bar
+```
+
+Arrays are searched for matching items. If any item in the array matches, the
+item is included in the list.
+
+```
+/v1/{type}?filter=spec.containers.image=alpine
+```
+
+#### `sort`
+
+Only applicable to list requests (`/v1/{type}` and `/v1/{type}/{namespace}`).
+
+Results can be sorted lexicographically by primary and secondary columns.
+
+Sorting by only a primary column, for example name:
+
+```
+/v1/{type}?sort=metadata.name
+```
+
+Reverse sorting by name:
+
+```
+/v1/{type}?sort=-metadata.name
+```
+
+The secondary sort criteria is comma separated.
+
+Example, sorting by name and creation time in ascending order:
+
+```
+/v1/{type}?sort=metadata.name,metadata.creationTimestamp
+```
+
+Reverse sort by name, normal sort by creation time:
+
+```
+/v1/{type}?sort=-metadata.name,metadata.creationTimestamp
+```
+
+Normal sort by name, reverse sort by creation time:
+
+```
+/v1/{type}?sort=metadata.name,-metadata.creationTimestamp
+```
+
+#### `page`, `pagesize`, and `revision`
+
+Only applicable to list requests (`/v1/{type}` and `/v1/{type}/{namespace}`).
+
+Results can be batched by pages for easier display.
+
+Example initial request returning a page with 10 results:
+
+```
+/v1/{type}?pagesize=10
+```
+
+Pages are one-indexed, so this is equivalent to
+
+```
+/v1/{type}?pagesize=10&page=1
+```
+To retrieve subsequent pages, the page number and the list revision number must
+be included in the request. This ensures the page will be retrieved from the
+cache, rather than making a new request to Kubernetes. If the revision number
+is omitted, a new fetch is performed in order to get the latest revision. The
+revision is included in the list response.
+
+```
+/v1/{type}?pagezie=10&page=2&revision=107440
+```
+
+The total number of pages and individual items are included in the list
+response as `pages` and `count` respectively.
+
+If a page number is out of bounds, an empty list is returned.
+
+`page` and `pagesize` can be used alongside the `limit` and `continue`
+parameters supported by Kubernetes. `limit` and `continue` are typically used
+for server-side chunking and do not guarantee results in any order.
--- a/README.md
+++ b/README.md
@@ -50,6 +50,39 @@ generating a kubeconfig for a cluster, or installing an app from a catalog:
 POST /v1/catalog.cattle.io.clusterrepos/rancher-partner-charts?action=install
 ```
 
+#### `limit`
+
+Only applicable to list requests (`/v1/{type}` and `/v1/{type}/{namespace}`).
+
+Set the maximum number of results to retrieve from Kubernetes. The limit is
+passed on as a parameter to the Kubernetes request. The purpose of setting this
+limit is to prevent a huge response from overwhelming Steve and Rancher. For
+more information about setting limits, review the Kubernetes documentation on
+[retrieving results in
+chunks](https://kubernetes.io/docs/reference/using-api/api-concepts/#retrieving-large-results-sets-in-chunks).
+
+The limit controls the size of the set coming from Kubernetes, and then
+filtering, sorting, and pagination are applied on that set. Because of this, if
+the result set is partial, there is no guarantee that the result returned to
+the client is fully sorted across the entire list, only across the returned
+chunk.
+
+The returned response will include a `continue` token, which indicates that the
+result is partial and must be used in the subsequent request to retrieve the
+next chunk.
+
+The default limit is 100000. To override the default, set `limit=-1`.
+
+#### `continue`
+
+Only applicable to list requests (`/v1/{type}` and `/v1/{type}/{namespace}`).
+
+Continue retrieving the next chunk of a partial list. The continue token is
+included in the response of a limited list and indicates that the result is
+partial. This token can then be used as a query parameter to retrieve the next
+chunk. All chunks have been retrieved when the continue field in the response
+is empty.
+
 #### `filter`
 
 Only applicable to list requests (`/v1/{type}` and `/v1/{type}/{namespace}`).
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/pborman/uuid v1.2.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.12.1
-	github.com/rancher/apiserver v0.0.0-20221205175736-7c507bd5c076
+	github.com/rancher/apiserver v0.0.0-20221220225852-94cba4f28cfd
 	github.com/rancher/dynamiclistener v0.3.5
 	github.com/rancher/kubernetes-provider-detector v0.1.2
 	github.com/rancher/norman v0.0.0-20221205184727-32ef2e185b99
--- a/pkg/stores/partition/listprocessor/processor.go
+++ b/pkg/stores/partition/listprocessor/processor.go
@@ -1,0 +1,150 @@
+// Package listprocessor contains methods for filtering, sorting, and paginating lists of objects.
+package listprocessor
+
+import (
+	"strconv"
+	"strings"
+
+	"github.com/rancher/apiserver/pkg/types"
+	"github.com/rancher/wrangler/pkg/data"
+	"github.com/rancher/wrangler/pkg/data/convert"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+const (
+	defaultLimit  = 100000
+	continueParam = "continue"
+	limitParam    = "limit"
+	filterParam   = "filter"
+)
+
+// ListOptions represents the query parameters that may be included in a list request.
+type ListOptions struct {
+	ChunkSize int
+	Resume    string
+	Filters   []Filter
+}
+
+// Filter represents a field to filter by.
+// A subfield in an object is represented in a request query using . notation, e.g. 'metadata.name'.
+// The subfield is internally represented as a slice, e.g. [metadata, name].
+type Filter struct {
+	field []string
+	match string
+}
+
+// ParseQuery parses the query params of a request and returns a ListOptions.
+func ParseQuery(apiOp *types.APIRequest) *ListOptions {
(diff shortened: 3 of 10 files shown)


go.mod
```diff
@@ -6,6 +6,7 @@ replace (
 	github.com/crewjam/saml => github.com/rancher/saml v0.0.0-20180713225824-ce1532152fde
 	github.com/knative/pkg => github.com/rancher/pkg v0.0.0-20190514055449-b30ab9de040e
 	github.com/matryer/moq => github.com/rancher/moq v0.0.0-20190404221404-ee5226d43009
+	github.com/rancher/apiserver => github.com/cmurphy/apiserver v0.0.0-20221110205644-da6f6ae1675a
```
cmurphy, line 9: DO NOT MERGE without updating this

README.md
```diff
@@ -0,0 +1,144 @@
+steve
```
MbolotSuse, line 1: Thanks for adding docs.

cmurphy:
Latest changes - 
1. updated the pagination cache key to use accesscontrol.AccessFor instead of username and groups
2. changed the default cache size to 1000 and made it configurable

pkg/stores/partition/store.go
```diff
@@ -2,28 +2,85 @@ package partition
...
+	if v := os.Getenv(cacheSizeEnv); v != "" {
+		sizeInt, err := strconv.Atoi(v)
+		if err == nil {
+			cacheSize = sizeInt
+		}
+	}
+	return &Store{
+		Partitioner: partitioner,
+		asl:         asl,
+		listCache:   cache.NewLRUExpireCache(cacheSize),
```
rmweir, line 58: I think ~~the max size should be configurable and~~ that it should be possible to disable the use of the cache entirely. It could be hard to come up with the right number because the size of lists will differ so much depending on object type and user permissions. What if a user's permissions have changed since the last cache? I think the 30 minute expiry should be fine for this but maybe someone from security can approve.

pkg/stores/partition/listprocessor/processor.go
```diff
@@ -0,0 +1,287 @@
...
+		fieldI := strings.Join(filterOpts[i].field, ".")
+		fieldJ := strings.Join(filterOpts[j].field, ".")
+		return fieldI < fieldJ
+	})
+	sortOpts := Sort{}
+	sortKeys := q.Get(sortParam)
+	if sortKeys != "" {
+		sortParts := strings.SplitN(sortKeys, ",", 2)
+		primaryField := sortParts[0]
+		if primaryField[0] == '-' {
```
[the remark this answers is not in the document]
cmurphy: Not quite. The length of the parameter is checked on 120. Then the slice created on 121 must have at least one non-empty string in it. That string is shortened by 1 on 125. At that point, it's possible it could be empty, hence the check on 127.

README.md
```diff
@@ -0,0 +1,144 @@
...
+### /v1 API
+
+steve registers all Kubernetes resources as schemas in the /v1 API. Any
+endpoint can support methods GET, POST, PATCH, PUT, or DELETE depending on what
+the underlying Kubernetes endpoint supports and the user's permissions.
+
+* `/v1/{type}` - all cluster-scoped resources OR all resources in all
+  namespaces of type `{type}` that the user has access to
+* `/v1/{type}/{name}` - cluster-scoped resource of type `{type}` and unique name `{name}`
+* `/v1/{type}/{namespace} - all resources of type `{type}` under namespace `{namespace}`
```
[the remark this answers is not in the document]
cmurphy: :+1:

README.md
```diff
@@ -0,0 +1,144 @@
...
+
+#### `link`
+
+Trigger a link handler, which is registered with the schema. Examples are
+calling the shell for a cluster, or retrieving logs during catalog operations.
+
+#### `action`
+
+Trigger an action handler, which is registered with the schema. Examples are
+generating a kubeconfig for a cluster, or installing an app from a catalog.
```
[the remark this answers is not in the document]
cmurphy: :+1:

pkg/stores/partition/listprocessor/processor.go
```diff
@@ -0,0 +1,287 @@
...
+		if !ok {
+			subField = append(subField, filter.field[len(filter.field)-1])
+			filter.field = filter.field[:len(filter.field)-1]
+		}
+	}
+	if !ok {
+		return false
+	}
+	switch typedVal := objValue.(type) {
+	case string, int, bool:
```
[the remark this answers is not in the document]
cmurphy: Fixed, added a test to cover this

pkg/stores/partition/store.go
cmurphy: > it should be possible to disable the use of the cache entirely

done, added a test for this

> What if a user's permissions have changed since the last cache?

This is done now with the access set check, the access set ID is part of the cache key and is recalculated on every request.

README.md
```diff
@@ -0,0 +1,153 @@
+steve
+=====
+
+steve is a lightweight API proxy for kubernetes whose aim is to create an
```
KevinJoiner, line 4:
```suggestion
Steve is a lightweight API proxy for Kubernetes whose aim is to create an
```

README.md
```diff
@@ -0,0 +1,153 @@
...
+interface layer suitable for dashboards to efficiently interact with
+kubernetes.
```
KevinJoiner, line 6:
```suggestion
Kubernetes.
```

README.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Kubernetes proxy
+
+Requests made to `/api`, `/api/*`, `/apis/*`, `/openapi/*` and `/version` will
+be proxied directly to Kubernetes.
+
+### /v1 API
+
+steve registers all Kubernetes resources as schemas in the /v1 API. Any
+endpoint can support methods GET, POST, PATCH, PUT, or DELETE depending on what
```
KevinJoiner, line 19:
```suggestion
endpoint can support methods GET, POST, PATCH, PUT, or DELETE, depending on what
```

README.md
```diff
@@ -0,0 +1,153 @@
...
+* `/v1/{type}` - all cluster-scoped resources OR all resources in all
+  namespaces of type `{type}` that the user has access to
+* `/v1/{type}/{name}` - cluster-scoped resource of type `{type}` and unique name `{name}`
+* `/v1/{type}/{namespace}` - all resources of type `{type}` under namespace `{namespace}`
+* `/v1/{type}/{namespace}/{name}` - resource of type `{type}` under namespace
+  `{namespace}` with name `{name}` unique within the namespace
+
+### Query parameters
+
+steve supports query parameters to perform actions or process data on top of
```
KevinJoiner, line 31:
```suggestion
Steve supports query parameters to perform actions or process data on top of
```

README.md
```diff
@@ -0,0 +1,153 @@
...
+what Kubernetes supports.
+
+#### `link`
+
+Trigger a link handler, which is registered with the schema. Examples are
+calling the shell for a cluster, or following logs during cluster or catalog
```
KevinJoiner, line 37:
```suggestion
calling the shell for a cluster, or following logs during a cluster or catalog
```

README.md
```diff
@@ -0,0 +1,153 @@
...
+#### `action`
+
+Trigger an action handler, which is registered with the schema. Examples are
+generating a kubeconfig for a cluster, or installing an app from a catalog:
+
+```
+POST /v1/catalog.cattle.io.clusterrepos/rancher-partner-charts?action=install
+```
+
+#### `filter` (new for Rancher v2.7.x)
```
KevinJoiner, line 53: We could leave this off since Steve is used outside of Rancher.
```suggestion
#### `filter` 
```

README.md
```diff
@@ -0,0 +1,153 @@
...
+
+```
+/v1/{type}?filter=spec.containers.image=alpine
+```
+
+#### `sort` (new for Rancher v2.7.x)
+
+Only applicable to list requests (`/v1/{type}` and `/v1/{type}/{namespace}`)
+
+Results can be sorted lexicographically by primary and secondary colums.
```
KevinJoiner, line 85:
```suggestion
Results can be sorted lexicographically by primary and secondary columns.
```

README.md
```diff
@@ -0,0 +1,153 @@
...
+/v1/{type}?sort=metadata.name
+```
+
+Reverse sorting by name:
+
+```
+/v1/{type}?sort=-metadata.name
+```
+
+Secondary sort criteria is comma separated.
```
KevinJoiner, line 99:
```suggestion
The secondary sort criteria is comma separated.
```

README.md
```diff
@@ -0,0 +1,153 @@
...
+/v1/{type}?sort=metadata.name,-metadata.creationTimestamp
+```
+
+#### `page` and `pagesize` (new for Rancher v2.7.x)
+
+Only applicable to list requests (`/v1/{type}` and `/v1/{type}/{namespace}`)
+
+Results can be batched by pages for easier display.
+
+Example initial request, returns a page with 10 results:
```
KevinJoiner, line 125:
```suggestion
Example initial request returns a page with 10 results:
```

pkg/stores/partition/listprocessor/processor.go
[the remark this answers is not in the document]
KevinJoiner: You will still need the check for both primaryField and secondaryField because of the `strings.SplitN(...)`
https://go.dev/play/p/ELlnx4eMRwS

pkg/stores/partition/store_test.go
```diff
@@ -0,0 +1,1832 @@
+package partition
```
KevinJoiner, line 1: Nothing to Change here; just a suggestion for future tests. If you leverage `testify.Suites` It is easier to break up tests into separate functions.

KevinJoiner (changes requested):
Nice job on the implementation. The separate commits made it easy to follow.

pkg/stores/partition/store.go
KevinJoiner: I see you added an ENV check for disabling the cache, but would it also be beneficial to use a query parameter to bypass the cache per request? This would allow the UI to use the cache based on the resource type.

KevinJoiner (changes requested):
Also, can you add package comments?

pkg/stores/partition/store.go
```diff
@@ -175,17 +300,79 @@ func (s *Store) Watch(apiOp *types.APIRequest, schema *types.APISchema, wr types
...
 }
 
-// getLimit extracts the limit parameter from the request or sets a default of 100000.
-// Since a default is always set, this implies that clients must always be
-// aware that the list may be incomplete.
-func getLimit(req *http.Request) int {
-	limitString := req.URL.Query().Get("limit")
-	limit, err := strconv.Atoi(limitString)
+func toAPI(schema *types.APISchema, obj runtime.Object) types.APIObject {
+	if obj == nil || reflect.ValueOf(obj).IsNil() {
```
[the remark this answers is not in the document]
cmurphy: > I think between these two cases, it is better to assume that the runtime.Object's underlying value is a non-nil struct, than it is to assume that the runtime.Object's underlying value is a struct with a pointer receiver mainly because the latter is not idiomatic behavior in go.

runtime.Object should basically always contain a struct with methods defined for pointer receivers because the types that implement [runtime.Object](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#Object) include [metav1.TypeMeta](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#TypeMeta.GetObjectKind) as an [embedded field](https://github.com/rancher/rancher/blob/38979f8ebe5c1ecb0121494d185758741fcb0d5b/pkg/apis/management.cattle.io/v3/public_authn_types.go#L12) and have autogenerated [DeepCopyObject pointer methods](https://github.com/rancher/rancher/blob/38979f8ebe5c1ecb0121494d185758741fcb0d5b/pkg/apis/management.cattle.io/v3/zz_generated_deepcopy.go#L539). So I think we still need reflect no matter what. We could add safety by checking `.Kind` before checking `.IsNil()`.

cmurphy:
> Also, can you add package comments?

giphy

pkg/stores/partition/store.go
```diff
@@ -88,38 +158,96 @@ func (s *Store) List(apiOp *types.APIRequest, schema *types.APISchema) (types.AP
...
 
-	list, err := lister.List(apiOp.Context(), limit, resume)
+	key, err := s.getCacheKey(apiOp, opts)
 	if err != nil {
 		return result, err
 	}
 
-	for items := range list {
-		result.Objects = append(result.Objects, items...)
+	var list []unstructured.Unstructured
```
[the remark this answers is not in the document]
KevinJoiner: I can agree with that sentiment. I can be swayed either way so let's just leave it as is for now.

README.md
KevinJoiner: You are right, I misread "operations".

README.md
KevinJoiner: We can use a semicolon without a conjunction if they are independent clauses.
```suggestion
Example initial request; returns a page with 10 results.
```

cmurphy: Changed to "returning" to make it one clause and remove the comma

pkg/stores/partition/listprocessor/processor.go
```diff
@@ -0,0 +1,300 @@
...
+		Sort:       sortOpts,
+		Pagination: pagination,
+		Revision:   revision,
+	}
+}
+
+// getLimit extracts the limit parameter from the request or sets a default of 100000.
+// The default limit can be explicitly disabled by setting it to a negative value.
+// If the default is accepted, clients must be aware that the list may be incomplete, and use the "continue" token to get the next chunk of results.
+func getLimit(apiOp *types.APIRequest) int {
```
KevinJoiner, line 166: We should either update our behavior to state that if the limit is 0 or negative we unset the limit parameter or leave the check that overrides `0`to be the `defaultLimit`. I couldn't find any official documentation on `limit=0` in the kubeapi references, but through some simple curl testing, if I add  `limit=0` I get back all the results. This could be unexpected for users since our previous implementation would override the limit if it was set to '0'

pkg/stores/partition/parallel.go
```diff
@@ -87,6 +87,9 @@ func (p *ParallelPartitionLister) List(ctx context.Context, limit int, resume st
 			limit = state.Limit
 		}
 	}
+	if revision != "" {
+		state.Revision = revision
+	}
```
KevinJoiner, line 92: Do we need this check since `state.Revision` is already an empty string?

pkg/stores/partition/store.go
```diff
@@ -184,11 +187,14 @@ func (s *Store) List(apiOp *types.APIRequest, schema *types.APISchema) (types.AP
 		}
 	}
 	if list == nil { // did not look in cache or was not found in cache
-		stream, err := lister.List(apiOp.Context(), opts.ChunkSize, opts.Resume)
+		stream, err := lister.List(apiOp.Context(), opts.ChunkSize, opts.Resume, opts.Revision)
 		if err != nil {
 			return result, err
 		}
 		list = listprocessor.FilterList(stream, opts.Filters)
+		if lister.Err() != nil {
```
KevinJoiner, line 195: Maybe a comment here stating that we are checking for any errors returned during the listing operation. And that FilterList guarantees the stream has been consumed. 
I just have the fear of things getting moved around since there is no clear dependency on `FilterList(...)`

KevinJoiner (changes requested):

pkg/stores/partition/store.go
cmurphy: yeah good point, it wasn't even obvious to me at first why I couldn't check the error first

rmweir (approved):
LGTM

cmurphy:
Updated to include merged apiserver change

PR #70, opened 2023-01-02
Expose WarningBuffer type for external use

crobby:
Part of issue: https://github.com/rancher/rancher/issues/39413

### DIFF
--- a/pkg/stores/proxy/proxy_store.go
+++ b/pkg/stores/proxy/proxy_store.go
@@ -61,15 +61,12 @@ type ClientGetter interface {
 	TableAdminClientForWatch(ctx *types.APIRequest, schema *types.APISchema, namespace string, warningHandler rest.WarningHandler) (dynamic.ResourceInterface, error)
 }
 
-type warningBuffer struct {
-	Warnings []types.Warning
-}
+// WarningBuffer holds warnings that may be returned from the kubernetes api
+type WarningBuffer []types.Warning
 
-func (w *warningBuffer) HandleWarningHeader(code int, agent string, text string) {
-	if w.Warnings == nil {
-		w.Warnings = []types.Warning{}
-	}
-	w.Warnings = append(w.Warnings, types.Warning{
+// HandleWarningHeader takes the components of a kubernetes warning header and stores them
+func (w *WarningBuffer) HandleWarningHeader(code int, agent string, text string) {
+	*w = append(*w, types.Warning{
 		Code:  code,
 		Agent: agent,
 		Text:  text,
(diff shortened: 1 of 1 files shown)


pkg/stores/proxy/proxy_store.go
```diff
@@ -61,11 +61,11 @@ type ClientGetter interface {
 	TableAdminClientForWatch(ctx *types.APIRequest, schema *types.APISchema, namespace string, warningHandler rest.WarningHandler) (dynamic.ResourceInterface, error)
 }
 
-type warningBuffer struct {
+type WarningBuffer struct {
 	Warnings []types.Warning
 }
```
KevinJoiner, line 66: nit: because I don't know the timeline on this, but if you don't make this a struct it makes it easier for callers to use since we would remove the need for a pointer receiver to use a Warning Buffer which is better inline with ['Make the zero value useful.' ](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=385s&ab_channel=TheGoProgrammingLanguage)go proverb
```suggestion
type WarningBuffers []types.Warning
```

pkg/stores/proxy/proxy_store.go
```diff
@@ -61,11 +61,11 @@ type ClientGetter interface {
 	TableAdminClientForWatch(ctx *types.APIRequest, schema *types.APISchema, namespace string, warningHandler rest.WarningHandler) (dynamic.ResourceInterface, error)
 }
 
-type warningBuffer struct {
+type WarningBuffer struct {
```
KevinJoiner, line 64: comments

pkg/stores/proxy/proxy_store.go
```diff
@@ -61,11 +61,11 @@ type ClientGetter interface {
...
 	Warnings []types.Warning
 }
 
-func (w *warningBuffer) HandleWarningHeader(code int, agent string, text string) {
+func (w *WarningBuffer) HandleWarningHeader(code int, agent string, text string) {
```
KevinJoiner, line 68: comments

pkg/stores/proxy/proxy_store.go
```diff
@@ -61,11 +61,11 @@ type ClientGetter interface {
...
 	if w.Warnings == nil {
 		w.Warnings = []types.Warning{}
 	}
```
KevinJoiner, line 71: This can be deleted since it is safe to append to a nil list.
```suggestion
```

KevinJoiner (changes requested):

pkg/stores/proxy/proxy_store.go
crobby: not sure I follow this just yet.  What does HandleWarningHeader look like for this scenario?

KevinJoiner (approved):

maxsokolovsky (approved):

PR #72, opened 2023-01-11
Fixes a bug with resource.error watch events

MbolotSuse:
Watch events need to have a type defined by
k8s.io/apimachinery/pkg/watch to be properly interpreted. Fixes a bug where this was not the case

Related to rancher/rancher#40131

### DIFF
--- a/pkg/stores/proxy/proxy_store.go
+++ b/pkg/stores/proxy/proxy_store.go
@@ -257,7 +257,7 @@ func (s *Store) list(apiOp *types.APIRequest, schema *types.APISchema, client dy
 
 func returnErr(err error, c chan watch.Event) {
 	c <- watch.Event{
-		Type: "resource.error",
+		Type: watch.Error,
 		Object: &metav1.Status{
 			Message: err.Error(),
 		},


MbolotSuse:
## Background

When a k8s watch returns an error, we create a `watch.Event` that has an error type. This creates an api object that has a name of `resource.error`, which signals to users that the watch function has failed to process an item from the watch stream.

When converting these watch events to the apiObject defined by rancher's apiserver package, there's a switch which determines the name of the event - this is done in the [partition store](https://github.com/rancher/steve/blob/f4338dd8396f7630456c5ededb008da407987df5/pkg/stores/partition/store.go#L369). 

## Bug Details and Fix
- The `watch.Event` needs to have the type field set to `watch.Error` in cases of errors, so that the partition store can convert the name to `"resource.error"`. In the current format, since the type was a manual string not defined in the watch package, it used the default `"resource.change"` event, which resulted in the issue reported by the UI

## Testing 
- Manual test run to validate that this fixes the issue. No Unit test yet. Issue is reproducible on v2 cluster create/delete which appears to create a couple of these watch errors

KevinJoiner (approved):

cmurphy (approved):

PR #75, opened 2023-02-16
Vai - SQLite-based caching

moio:
**WIP** this requires introduction/discussion before an actual review, but it is here for people to take a look if interested.

## Summary

[Vai](https://hackweek.opensuse.org/22/projects/vai-a-kubernetes-api-caching-layer) provides an alternative implementation of `client-go`'s Informer object that is backed by a SQLite database (instead of plain Go objects in RAM). Its intention is:

- to provide a more RAM-efficient cache where a large amount of resources is used, with the same consistency guarantees Kubernetes is built upon
- to provide a flexible way to query the cache contents via SQL with rich filtering, sorting, pagination and join semantics - much richer than the `Indexer` interface provides

**This PR is an experimental version of Steve that uses Vai to implement the /v1 API**.

## Main potential benefits

- lower RAM usage, lower CPU usage, faster average response times, less variance in response times, lower load on `kube-apiserver` on "list" operations. [One first set of load tests](https://github.com/moio/scalability-tests/blob/20230306_steve_vai_tests/docs/20230306%20-%20steve%20vai%20tests.md) suggests all those are achievable
- obtain similar benefits for WebSockets ("watch" operations)
- cache preloading/warming
- sharing of caches with Wrangler/Lasso, thereby reducing CPU/RAM footprint further
- more flexible querying, making it easier to implement UI features
- potentially less maintenance burden, as good part of the complexity is moved either inside `client-go`'s proven constucts such as `Reflector` or inside SQLite itself
- allow in future to change Steve's architecture so that a live and fast connection to read data downstream cluster data may not be present (eg. read from the SQL cache instead)

## Changes in this PR

Commits are divided in three portions:

- Drop portion: remove mechanisms to be implemented in vai, leaving everything else as working as possible, and keeping tests green
- Refactoring portion: work around dropped code to allow the plugging of vai code
- `ListOptionIndexer` patches: basically import vai code. History has been kept for future maintenance, but could be squashed into one commit as well
- One final glue commit to put all together: `proxy.Store: make use of sqlcache.Informer`

PR is best reviewed commit-by-commit.

## Open work items

If the general direction of this project is accepted, more work will be needed that was not tackled during the present experimentation phase:

- either specify in code indexable properties for each resource type or add automatic indexing on request
- implement `LabelSelector`/`FieldSelector` support (if deemed useful)
- evaluate if handling warning buffers differently is necessary
- evaluate transparent gzip compression of resources in SQLite blobs
- add garbage collection of stale rows
- evaluate delegating WebSocket "watch" functionality to Vai's Informer
- support TableClient
- potentially add more tests
- merge the [accompanying rancher/rancher changes](https://github.com/rancher/rancher/compare/release/v2.7...moio:rancher:steve_vai): two trivial, one temporary, one to be discussed (allowing CGO, [using a pure-Go SQLite port](https://gitlab.com/cznic/sqlite) or using [nocgo](https://github.com/notti/nocgo))
- evaluate to reintroduce [a hack for special-case handling of star namespaces](https://github.com/rancher/steve/blob/3d3cc7767928d9de7999cc88081a839c59b6417a/pkg/stores/proxy/proxy_store.go#L204-L205) (currently it is implemented properly)
- evaluate cache warm-up strategies


moio:
closing as this will be worked on other PRs

PR #105, opened 2023-05-10
Add projectsornamespaces query parameter

cmurphy:
Add a new query parameter to filter resources by their namespace or their namespace's project. This parameter is separate from the existing `filter` parameter.

Filter by a comma-separated list of projects and/or namespaces with:

?projectsornamespaces=p1,n1,n2

The result can be negated with the ! operator:

?projectsornamespaces!=p1,n1,n2

https://github.com/rancher/rancher/issues/40140

### DIFF
--- a/pkg/stores/partition/listprocessor/processor.go
+++ b/pkg/stores/partition/listprocessor/processor.go
@@ -10,19 +10,24 @@ import (
 	"github.com/rancher/apiserver/pkg/types"
 	"github.com/rancher/wrangler/pkg/data"
 	"github.com/rancher/wrangler/pkg/data/convert"
+	corecontrollers "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
 const (
-	defaultLimit  = 100000
-	continueParam = "continue"
-	limitParam    = "limit"
-	filterParam   = "filter"
-	sortParam     = "sort"
-	pageSizeParam = "pagesize"
-	pageParam     = "page"
-	revisionParam = "revision"
-	orOp          = ","
+	defaultLimit            = 100000
+	continueParam           = "continue"
+	limitParam              = "limit"
+	filterParam             = "filter"
+	sortParam               = "sort"
+	pageSizeParam           = "pagesize"
+	pageParam               = "page"
+	revisionParam           = "revision"
+	projectsOrNamespacesVar = "projectsornamespaces"
+	projectIDFieldLabel     = "field.cattle.io/projectId"
+
+	orOp  = ","
+	notOp = "!"
 )
 
 var opReg = regexp.MustCompile(`[!]?=`)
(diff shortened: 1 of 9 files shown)


pkg/stores/partition/listprocessor/processor.go
```diff
@@ -360,3 +394,28 @@ func PaginateList(list []unstructured.Unstructured, p Pagination) ([]unstructure
...
+
+func FilterByProjectsAndNamespaces(list []unstructured.Unstructured, projectsOrNamespaces ProjectsOrNamespacesFilter, namespaceCache corecontrollers.NamespaceCache) []unstructured.Unstructured {
+	if len(projectsOrNamespaces.filter) == 0 {
+		return list
+	}
+	result := []unstructured.Unstructured{}
+	for _, obj := range list {
+		namespaceName := obj.GetNamespace()
+		namespace, _ := namespaceCache.Get(namespaceName)
+		if namespace == nil {
```
MbolotSuse, line 406: IIRC the go-ism here is that in the case of an `obj, err` return, the presence of an error indicates that the first value is invalid, but not necessarily nil. So in this case you could be getting a corrupt (i.e. non-nil but otherwise invalid object) namespace, but since you are ignoring the error we will keep trying to process it.

TL;DR: Maybe change this (and the preceeding line) to :

```golang
namespace, err := namespaceCache.Get(namespaceName)
if namespace == nil || err != nil {
```

pkg/stores/partition/listprocessor/processor.go
```diff
@@ -360,3 +394,28 @@ func PaginateList(list []unstructured.Unstructured, p Pagination) ([]unstructure
...
 	return list[offset : offset+p.pageSize], pages
 }
+
+func FilterByProjectsAndNamespaces(list []unstructured.Unstructured, projectsOrNamespaces ProjectsOrNamespacesFilter, namespaceCache corecontrollers.NamespaceCache) []unstructured.Unstructured {
+	if len(projectsOrNamespaces.filter) == 0 {
+		return list
+	}
+	result := []unstructured.Unstructured{}
+	for _, obj := range list {
+		namespaceName := obj.GetNamespace()
```
MbolotSuse, line 404: Nit:
Might be good after this line to do a:

```golang
if namespaceName == ""{
    continue
}
```

I'm slightly uncomfortable on relying on `namespaceCache.Get("")` returning nil.

pkg/stores/partition/listprocessor/processor.go
```diff
@@ -360,3 +394,28 @@ func PaginateList(list []unstructured.Unstructured, p Pagination) ([]unstructure
...
+		namespace, _ := namespaceCache.Get(namespaceName)
+		if namespace == nil {
+			continue
+		}
+		projectLabel, _ := namespace.GetLabels()[projectIDFieldLabel]
```
MbolotSuse, line 409: IIRC the annotation is the canonical source of project/namespace membership, and we should use that rather than the label where possible (and since we are just checking the string here and not using a selector, this case seems suitable more for the annotation).

MbolotSuse (changes requested):
Couple of mostly small items, overall the approach is very readable and well put-together.

pkg/stores/partition/listprocessor/processor.go
cmurphy: The annotation may be the source of truth but it doesn't make the label wrong. The label is kept in sync with the annotation. Using the label avoids having to do the ugly string split.

MbolotSuse: That's fair - sounds good.

MbolotSuse (approved):

PR #109, opened 2023-06-07
Fix drone validate build

moio:
go1.19 provides go=1.19.9, the right symbol for the major version is golang(API)

See https://build.opensuse.org/package/view_file/openSUSE:Factory/go1.19/go1.19.spec?expand=1

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -35,7 +35,7 @@ steps:
   - name: validate
     image: registry.suse.com/bci/bci-base:15.4
     commands:
-      - zypper in -y go=1.19 git tar gzip make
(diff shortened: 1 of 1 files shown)


brandond (approved):

moio:
> go1.19 also seems to work, but I'll default to you on the best way to get the matching version of go that we want.

Indeed the difference is tiny - the only advantage in relying on the symbol instead of the package name means that the package name could be renamed without us being affected.

I will not be against just using `go1.19` if that's easier to understand for non-SLE-centric developers. ☮️

PR #110, opened 2023-06-07
Proxy store improve error handling

moio:
Master port of #107

### DIFF
--- a/pkg/stores/proxy/proxy_store.go
+++ b/pkg/stores/proxy/proxy_store.go
@@ -311,7 +311,6 @@ func (s *Store) listAndWatch(apiOp *types.APIRequest, client dynamic.ResourceInt
 					rowToObject(obj)
 					result <- watch.Event{Type: watch.Modified, Object: obj}
 				} else {
-					logrus.Debugf("notifier watch error: %v", err)
 					returnErr(errors.Wrapf(err, "notifier watch error: %v", err), result)
 				}
 			}
@@ -323,7 +322,6 @@ func (s *Store) listAndWatch(apiOp *types.APIRequest, client dynamic.ResourceInt
 		for event := range watcher.ResultChan() {
 			if event.Type == watch.Error {
 				if status, ok := event.Object.(*metav1.Status); ok {
-					logrus.Debugf("event watch error: %s", status.Message)
 					returnErr(fmt.Errorf("event watch error: %s", status.Message), result)
 				} else {
 					logrus.Debugf("event watch error: could not decode event object %T", event.Object)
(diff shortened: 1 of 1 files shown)


brandond (approved):

pkg/stores/proxy/proxy_store.go
```diff
@@ -363,12 +363,22 @@ func (s *Store) WatchNames(apiOp *types.APIRequest, schema *types.APISchema, w t
 	go func() {
 		defer close(result)
 		for item := range c {
+			if item.Type == watch.Error {
+				if status, ok := item.Object.(*metav1.Status); ok {
+					logrus.Debugf("WatchNames received error: %s", status.Message)
+				} else {
+					logrus.Debugf("WatchNames received error: %v", item)
+				}
```
rmweir, line 371: I think these logs may be too chatty for the the final version. Errors come through here frequently. It happens every time a ui session expires, etc.

rmweir (changes requested):
I think for the final vers

pkg/stores/proxy/proxy_store.go
rmweir: After reviewing customer logs with this patch, I can confirm that they are also redundant as the errors are already logged in debug.

moio: Thanks for the observations. Today I had a fresh look and observed the following:

* there only are 5 places that send events down this channel, all of them in `listAndWatch`
  * 2 of them are guaranteed to _never_ send down errors (`Type == Error` events): lines [312](https://github.com/rancher/steve/blob/3cbeea22d38a6250d4c1211a2eaeb3c764c4c22c/pkg/stores/proxy/proxy_store.go#L312) and [336](https://github.com/rancher/steve/blob/3cbeea22d38a6250d4c1211a2eaeb3c764c4c22c/pkg/stores/proxy/proxy_store.go#L336)
  * 3 of them are guaranteed to _always_ send down errors (using `returnErr`): lines [293](https://github.com/rancher/steve/blob/3cbeea22d38a6250d4c1211a2eaeb3c764c4c22c/pkg/stores/proxy/proxy_store.go#L293), [315](https://github.com/rancher/steve/blob/3cbeea22d38a6250d4c1211a2eaeb3c764c4c22c/pkg/stores/proxy/proxy_store.go#L315), and [327](https://github.com/rancher/steve/blob/3cbeea22d38a6250d4c1211a2eaeb3c764c4c22c/pkg/stores/proxy/proxy_store.go#L327)
    * those `returnErr` calls are preceded by `logrus.Debugf()` in 2 out of 3 cases


My ignorance and the fact that `returnErr` calls were not always preceded by `logrus.Debugf()` are the reasons why I purposely erred on the side of excess and logged all cases, as the main goal was diagnosis. Now that the problem and the solution are understood better, I agree such redundancy should be removed.

With that said, I suggest to keep logging all cases we used to log before (including, as you mentioned, UI expiration cases, refreshes of watches and so on). I believe this will be especially useful to help frontend developers to troubleshoot a possible reopening of issue https://github.com/rancher/rancher/issues/37627 without intervening in Steve source code if possible. Chattiness I am less worried about, as they are all under the debug flag, so they will only be visible to engineers explicitly looking for problems. And as you mentioned elsewhere, it's important to retain the ability of getting some measure of "time density" - too many might signal problems.

I will be keeping logging lines here (as opposed to moving them to `returnErr` or `listAndWatch`) as we have to type cast/check in this place anyway.

I hope you will agree and remain available to discuss and rethink this if you have any further observations.

moio: Addressed in  bce6c5b

rmweir: I think that makes a lot more sense, LGTM.

rmweir (approved):
LGTM

cmurphy (approved):

PR #112, opened 2023-06-16
Drop unrecognized fields before update

cmurphy:
Add a nested store to the proxy store to strip non-Kubernetes fields from the object being updated.

The steve formatter and proxy store adds fields to objects when it outputs them to the client, for usability by the UI. It adds the object's fields[1], relationships to other objects[2], a summary of the object's state[3], and additional information in the conditions[4]. These fields are not native to Kubernetes, so when a client submits the object back as an update, Kubernetes reports a warning that they are unrecognized. This change ensures the extra fields are removed before submitting the update.

The store drops the entire status field before updating. Steve stores only support client-go's Update and not UpdateStatus[5], so updating the status has no effect anyway.
[1] https://github.com/rancher/steve/blob/bf2e9655f5dde8f55b23f67e64f0186fc68789d7/pkg/stores/proxy/proxy_store.go#L189
[2] https://github.com/rancher/steve/blob/bf2e9655f5dde8f55b23f67e64f0186fc68789d7/pkg/resources/common/formatter.go#L106
[3] https://github.com/rancher/steve/blob/bf2e9655f5dde8f55b23f67e64f0186fc68789d7/pkg/resources/common/formatter.go#L100
[4] https://github.com/rancher/steve/blob/bf2e9655f5dde8f55b23f67e64f0186fc68789d7/pkg/resources/common/formatter.go#L108
[5] https://pkg.go.dev/k8s.io/client-go/dynamic#ResourceInterface

https://github.com/rancher/rancher/issues/41772

### DIFF
--- a/pkg/stores/proxy/unformatter.go
+++ b/pkg/stores/proxy/unformatter.go
@@ -1,0 +1,66 @@
+package proxy
+
+import (
+	"github.com/rancher/apiserver/pkg/types"
+	"github.com/rancher/wrangler/pkg/data"
+	"github.com/rancher/wrangler/pkg/data/convert"
+)
+
+// unformatterStore removes fields added by the formatter that kubernetes cannot recognize.
+type unformatterStore struct {
+	types.Store
+}
+
+// ByID looks up a single object by its ID.
+func (u *unformatterStore) ByID(apiOp *types.APIRequest, schema *types.APISchema, id string) (types.APIObject, error) {
+	return u.Store.ByID(apiOp, schema, id)
+}
+
+// List returns a list of resources.
+func (u *unformatterStore) List(apiOp *types.APIRequest, schema *types.APISchema) (types.APIObjectList, error) {
+	return u.Store.List(apiOp, schema)
+}
+
+// Create creates a single object in the store.
+func (u *unformatterStore) Create(apiOp *types.APIRequest, schema *types.APISchema, data types.APIObject) (types.APIObject, error) {
+	return u.Store.Create(apiOp, schema, data)
+}
+
+// Update updates a single object in the store.
+func (u *unformatterStore) Update(apiOp *types.APIRequest, schema *types.APISchema, data types.APIObject, id string) (types.APIObject, error) {
+	data = unformat(data)
+	return u.Store.Update(apiOp, schema, data, id)
+}
+
+// Delete deletes an object from a store.
+func (u *unformatterStore) Delete(apiOp *types.APIRequest, schema *types.APISchema, id string) (types.APIObject, error) {
+	return u.Store.Delete(apiOp, schema, id)
+
+}
+
+// Watch returns a channel of events for a list or resource.
+func (u *unformatterStore) Watch(apiOp *types.APIRequest, schema *types.APISchema, wr types.WatchRequest) (chan types.APIEvent, error) {
+	return u.Store.Watch(apiOp, schema, wr)
+}
+
+func unformat(obj types.APIObject) types.APIObject {
+	unst, ok := obj.Object.(map[string]interface{})
+	if !ok {
+		return obj
+	}
+	data.RemoveValue(unst, "metadata", "fields")
+	data.RemoveValue(unst, "metadata", "relationships")
+	data.RemoveValue(unst, "metadata", "state")
+	conditions, ok := data.GetValue(unst, "status", "conditions")
+	if ok {
+		conditionsSlice := convert.ToMapSlice(conditions)
+		for i := range conditionsSlice {
+			data.RemoveValue(conditionsSlice[i], "error")
+			data.RemoveValue(conditionsSlice[i], "transitioning")
+			data.RemoveValue(conditionsSlice[i], "lastUpdateTime")
+		}
+		data.PutValue(unst, conditionsSlice, "status", "conditions")
+	}
+	obj.Object = unst
+	return obj
+}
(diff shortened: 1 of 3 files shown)


pkg/stores/proxy/unformatter.go
```diff
@@ -0,0 +1,56 @@
...
+
+func unformat(obj types.APIObject) types.APIObject {
+	unst, ok := obj.Object.(map[string]interface{})
+	if !ok {
+		return obj
+	}
+	data.RemoveValue(unst, "metadata", "fields")
+	data.RemoveValue(unst, "metadata", "relationships")
+	data.RemoveValue(unst, "metadata", "state")
+	data.RemoveValue(unst, "status") // client-go UpdateStatus is not supported
```
KevinJoiner, line 53: I was under the impression that the Update call from [ResourceInterface](https://pkg.go.dev/k8s.io/client-go/dynamic#ResourceInterface) includes the Status since it is a part of the object. And the UpdateStatus call is used to update the status only. 

If this is not true and Update calls to K8s will not update the status even if it differs from the stored object then this should be dropped.

KevinJoiner (changes requested):

pkg/stores/proxy/unformatter.go
cmurphy: I don't think that's the case. Here's what the [docs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource) say:

> PUT/POST/PATCH requests to the custom resource ignore changes to the status stanza.

Here's a brief reproducer showing this: https://gist.github.com/cmurphy/1512fa9482f8eba5cfed352f5d809894

Even though the status isn't updated, the API server still validates its fields, which is why we see the warnings about .status.conditions in the response in steve. So the status object still needs to be valid even though it's not being used.

KevinJoiner: Thanks for the example! It was easy to run and worked exactly as you said. I didn't know about that behavior before this. Though I did notice that the behavior depends on `status` being defined as a subresource in the CRD definition. If I removed the `status` subresource line in `crds.yaml`, then changes to that status field persisted which concerns me because some of our resources such as `v3.User` do not have `status` defined as a subresource`
``` yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  creationTimestamp: "2023-06-16T14:15:37Z"
  generation: 1
  name: users.management.cattle.io
  resourceVersion: "743"
  uid: 06aa895c-bfbf-4f71-8a4f-43e31cd2bfcd
spec:
  conversion:
    strategy: None
  group: management.cattle.io
  names:
    kind: User
    listKind: UserList
    plural: users
    singular: user
  scope: Cluster
  versions:
  - name: v3
    schema:
      openAPIV3Schema:
        type: object
        x-kubernetes-preserve-unknown-fields: true
    served: true
    storage: true
   ```

Comment (cmurphy) [reply pkg/stores/proxy/unformatter.go:53]: Thanks, that's unfortunate that our CRDs don't conform to k8s standards. I'll rework this to drop the individual condition fields in the condition slice.

Comment (KevinJoiner) [reply pkg/stores/proxy/unformatter.go]: We may be able to fix most of these with our crd generation changes.

File: pkg/stores/proxy/unformatter_test.go [hunk 2]
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	"testing"
+
+	"github.com/rancher/apiserver/pkg/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_unformat(t *testing.T) {
+	tests := []struct {
```
maxsokolovsky, line 11: It's always good to start a test with `t.Parallel()` and also run subtests in parallel, unless there is global state.

```
for _, test := range tests {
	test := test
	t.Run(test.name, func(t *testing.T) {
		t.Parallel()
```

pkg/stores/proxy/unformatter_test.go
```diff
@@ -0,0 +1,110 @@
+package proxy
+
+import (
+	"testing"
+
+	"github.com/rancher/apiserver/pkg/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_unformat(t *testing.T) {
```
maxsokolovsky, line 10: Would it be better if this tested the exported method `Update` on the type instead?
`unformat` is just an implementation detail and might change, while we want to be certain that the public interface works as expected.

cmurphy: In this case I don't find a lot of value in it. I would have to implement a mock store to hold the underlying `types.Store`, so it would be adding a lot of boilerplate to test the entire `.Update()` flow, when all that really matters is this one function.

pkg/stores/proxy/unformatter_test.go
cmurphy: done

maxsokolovsky: Nit: the test itself can also be run in parallel:
```suggestion
	func TestUnformat(t *testing.T) {
	    t.Parallel()
```

pkg/stores/proxy/unformatter_test.go
maxsokolovsky: Makes sense. I just thought there might already be a simple mock store for testing...

maxsokolovsky (approved):

PR #125, opened 2023-09-12
Run container as a non-root user

maxsokolovsky:


macedogm (changes requested):
Thanks for making this PR!

It's only missing the line `USER 1000` right before the last instruction in the Dockefile `ENTRYPOINT ["/usr/bin/steve"]` to effectively make the container run as a non-root user.

For example, see how it's done in shell: https://github.com/rancher/shell/blob/fbde71b17730534ceb5eaed10724fb7445e2550c/package/Dockerfile#L40-L42

macedogm:
Additionally, if possible, please, I would recommend to switch `FROM registry.suse.com/bci/bci-micro:15.5.11.2` to be `FROM registry.suse.com/bci/bci-micro:15.5`. As we don't merge bump PRs very often for the moment, using a less specific tag for the image will make it pull the last version of the base image for `bci/bci-micro:15.5`, otherwise we might end up stuck with vulnerable packages in the base OS layer due to the exact/restrict tag version in use.

macedogm (approved):

Dockerfile
```diff
@@ -7,8 +7,18 @@ RUN \
...
+RUN echo "$user:x:1000:1000::/home/$user:/bin/bash" >> /etc/passwd && \
+    echo "$user:x:1000:" >> /etc/group && \
+    mkdir /home/$user && \
+    chown -R $user:$user /home/$user
+
 COPY --from=build /steve /usr/bin/steve
 # Hack to make golang do files,dns search order
 ENV LOCALDOMAIN=""
+USER $user
+WORKDIR /home/$user
```
cmurphy, line 23: (nit) I don't think WORKDIR is needed, since there are no subsequent commands that use it as a relative path.

cmurphy (approved):

Dockerfile
maxsokolovsky: Fixed!

PR #128, opened 2023-09-20
Add section on pagination design to README 

cmurphy:


KevinJoiner (approved):

cmurphy:
Fixed a typo and added some missing information to the API section

README.md
```diff
@@ -646,8 +653,99 @@ the user can define endpoints via a
...
+arbitrary fields. However, it also needs to be run after the parallelized
+partitioner has accumulated all the results, because each concurrent fetcher
+will only contain partial results. Therefore, the data remains in an
+unstructured format until after the listprocessor has been run, then the data
+is converted to a structured type. The below diagram illustrates the conversion
+sequence.
+
+![](./docs/store-flow.svg)
+
+#### Unit tests
```
MbolotSuse, line 708: Nit: While I think that the section on integration tests is useful (since those are in a separate repo in a hard-to-identify location), I'm not sure that I feel the same way on the unit tests. As long as you know what packages the code is in, it's easy to identify the test files. I would recommend just identifying the packages that are relevant to the list processor instead of including the section on unit tests.

MbolotSuse (approved):
One nit, but otherwise LGTM, thanks for adding docs.

README.md
cmurphy: It may be obvious where the unit tests are, but I think it's less obvious what to put in those unit tests. In some cases making changes to `pkg/stores/partition/listprocessor` could have an effect on the output of `pkg/stores/partition/store.go`, and it's worthwhile to add tests in both places even if the code in store.go didn't change (even if that's stretching the definition of "unit"), so I think it's valuable to keep that recommendation here.

KevinJoiner (approved):

PR #132, opened 2023-10-13
Updating ByNames to not return nil, nil

MbolotSuse:
ByNames could previously return a nil value and a nil error. This caused issues when other parts of the application
(pkg/stores/partition/parallel.go) tried to use the result. Now this will return an empty list on the error condition, instead of nil.

Related to rancher/rancher#43030.

## Overview

Steve's [partion lister](https://github.com/rancher/steve/blob/811e0b3572fd1f5008ec3b151658c97d578deca6/pkg/stores/partition/parallel.go#L179) makes in internal call to a `Lister`, which is (in turn) a proxy/rbac store. After making this call, it attempts to set several values from the result of that call, including the ResourceVersion of the list (which is useful for future calls to these resources, potentially including paginated use cases). In the case where the user has a permission granting permissions on specific resourceNames in the entire cluster (see below for a production use-case) this will result in a panic - since the proxy store returns `nil` for both the err and the list in this case.

This results in a panic in steve and the binary using steve (in this case, the cluster agent), which causes the linked issue above.

This case is triggered in (at least) the monitoring v2 use case - the [monitoring-ui-view](https://github.com/rancher/charts/blob/706052fd6f93d216c55d6528bc5de3a5cf06d440/charts/rancher-monitoring/103.0.0%2Bup45.31.1/templates/rancher-monitoring/clusterrole.yaml#L123) cluster role grants permissions on specific resource names, and can be assigned to users at the cluster level using a cluster role binding (see rancher/dashboard#9792 for an example).

## Solution

Change the proxy store to return a valid value in the identified use case (user has permissions on all resources of a specific name at the cluster scope). Long term, we may want this to return valid values, but given the explicit warnings in the code I think this needs more investigation before we go that route.

### DIFF
--- a/pkg/stores/proxy/proxy_store.go
+++ b/pkg/stores/proxy/proxy_store.go
@@ -205,9 +205,10 @@ func tableToObjects(obj map[string]interface{}) []unstructured.Unstructured {
 // be returned in the list.
 func (s *Store) ByNames(apiOp *types.APIRequest, schema *types.APISchema, names sets.String) (*unstructured.UnstructuredList, []types.Warning, error) {
 	if apiOp.Namespace == "*" {
-		// This happens when you grant namespaced objects with "get" by name in a clusterrolebinding. We will treat
-		// this as an invalid situation instead of listing all objects in the cluster and filtering by name.
-		return nil, nil, nil
+		// This happens when you grant namespaced objects with "get" or "list "by name in a clusterrolebinding.
+		// We will treat this as an invalid situation instead of listing all objects in the cluster
+		// and filtering by name.
+		return &unstructured.UnstructuredList{}, nil, nil
 	}
 	buffer := WarningBuffer{}
 	adminClient, err := s.clientGetter.TableAdminClient(apiOp, schema, apiOp.Namespace, &buffer)
--- a/pkg/stores/proxy/proxy_store_test.go
+++ b/pkg/stores/proxy/proxy_store_test.go
@@ -68,6 +68,18 @@ func TestWatchNamesErrReceive(t *testing.T) {
 	assert.Equal(t, 0, len(c.ResultChan()), "Expected all secrets to have been received")
 }
 
+func TestByNames(t *testing.T) {
+	s := Store{}
+	apiSchema := &types.APISchema{Schema: &schemas.Schema{}}
+	apiOp := &types.APIRequest{Namespace: "*", Schema: apiSchema, Request: &http.Request{}}
+	names := sets.NewString("some-resource", "some-other-resource")
+	result, warn, err := s.ByNames(apiOp, apiSchema, names)
+	assert.NotNil(t, result)
+	assert.Len(t, result.Items, 0)
+	assert.Nil(t, err)
+	assert.Nil(t, warn)
+}
+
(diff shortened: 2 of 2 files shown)


pkg/stores/proxy/proxy_store.go
```diff
@@ -219,6 +214,12 @@ func (s *Store) ByNames(apiOp *types.APIRequest, schema *types.APISchema, names
 	if err != nil {
 		return nil, nil, err
 	}
+	if apiOp.Namespace == "*" {
+		// This happens when you grant namespaced objects with "get" by name in a clusterrolebinding. We will treat
+		// this as an invalid situation instead of listing all objects in the cluster and filtering by name.
+		objs.Items = []unstructured.Unstructured{}
+		return objs, nil, nil
+	}
```
KevinJoiner, line 222: Does moving this after `s.list(...)` cause us to now list all objects in the cluster?

pkg/stores/proxy/proxy_store.go
```diff
@@ -219,6 +214,12 @@ func (s *Store) ByNames(apiOp *types.APIRequest, schema *types.APISchema, names
 	if err != nil {
 		return nil, nil, err
 	}
+	if apiOp.Namespace == "*" {
+		// This happens when you grant namespaced objects with "get" by name in a clusterrolebinding. We will treat
+		// this as an invalid situation instead of listing all objects in the cluster and filtering by name.
+		objs.Items = []unstructured.Unstructured{}
+		return objs, nil, nil
```
KevinJoiner, line 221: It's not a great design pattern to return a nil object and a nil error. However I think the calling function should be able to handle `obj.Items` being nil.
```suggestion
		objs = &unstructured.UnstructuredList{}
		return objs, nil, nil
```

KevinJoiner (changes requested):

pkg/stores/proxy/proxy_store.go
MbolotSuse: Yes, it will.

pkg/stores/proxy/proxy_store.go
MbolotSuse: I purposefully modified the return of a real list call. In this case in particular, I'm trying to preserve a valid `ResourceVersion`, which the [consuming function](https://github.com/rancher/steve/blob/811e0b3572fd1f5008ec3b151658c97d578deca6/pkg/stores/partition/parallel.go#L179) uses.

In short, I'd rather produce a valid (but empty) list call, rather than attempting to construct a minimum 0 value for the unstructuredList that other calls can use.

pkg/stores/proxy/proxy_store.go
MbolotSuse: Resolving, discussed offline, will move above.

pkg/stores/proxy/proxy_store.go
MbolotSuse: Discussed offline - it seems that it's not necessary to use a real call and other areas (including the UI) can handle that fine. To save on performance, this will be moved to the start of the function and return an empty unstructured list like @KevinJoiner suggested.

rmweir (approved):
LGTM

KevinJoiner (approved):

PR #140, opened 2023-12-21
Update dynamic listener

chiukapoor:
### Update
* ~~Bump k8s dependencies to 0.28.6~~
* ~~Update wrangler to v2.1.3~~
* Update dynamic listener to v0.5.0-rc2

### Related
* https://github.com/rancher/rancher/issues/43109


kinarashah (commented):
WIP. Discussed offline, we need to ensure master branches of dependency repos use the corresponding commits/tags from master branch only. Will re-review once updated.

MbolotSuse (approved):

PR #141, opened 2023-12-22
WatchNames: return errors via WebSocket

moio:
Align behavior with plain `Watch`, which also does the same. UI code actually expects the same behavior in both cases.

Fixes https://github.com/rancher/rancher/issues/41809

### DIFF
--- a/pkg/stores/proxy/proxy_store.go
+++ b/pkg/stores/proxy/proxy_store.go
@@ -370,6 +370,7 @@ func (s *Store) WatchNames(apiOp *types.APIRequest, schema *types.APISchema, w t
 				} else {
 					logrus.Debugf("WatchNames received error: %v", item)
 				}
+				result <- item
 				continue
 			}
 


rmweir (approved):
LGTM

pkg/stores/proxy/proxy_store.go
```diff
@@ -370,6 +370,7 @@ func (s *Store) WatchNames(apiOp *types.APIRequest, schema *types.APISchema, w t
 				} else {
 					logrus.Debugf("WatchNames received error: %v", item)
 				}
+				result <- item
```
MbolotSuse, line 373: I'm a bit concerned on the security of this option. From the [go docs](https://pkg.go.dev/k8s.io/apimachinery@v0.29.0/pkg/watch#Event) for the `Object` field of the `watch.Event` type:

```
//  * If Type is Error: *api.Status is recommended; other types may make sense
//    depending on context.
```

This to me indicates that this could contain a real object. 

AFAIK, The function that this is using (`WatchNames`) is used when the user doesn't have access to all objects, but still needs to watch for the resources that they do have access to. To achieve this, this function [filters](https://github.com/rancher/steve/blob/9935404019063126b6ac856db1ad56e0bc457ff8/pkg/stores/proxy/proxy_store.go#L376) the objects by names to ensure that the user has access to the object in the channel. The user only gets the object back if it has access to that object. However, since this filtering happens after the error handling logic, this new code could theoretically return an object that the user doesn't have access to.

MbolotSuse (changes requested):
One comment on the safety of returning the error event without doing any processing of the object that it contains.

I also have an overall question on the solution: my first glance would say that the ideal solution would be to immediately return a `resource.error` event when we are given an invalid resource version, rather than a `resource.start`, then a `resource.error`, then a `resource.stop`. I think that the idea here is just to align this with standard watch rather than re-evaluate the API behavior, but wanted to ask just to be sure - would you mind elaborating on why you choose the start -> error -> stop flow here?

pkg/stores/proxy/proxy_store.go
moio: You are right that sending arbitrary objects down the `result` channel in the `Object` field of `watch.Event` is not prevented by types alone.

Nevertheless I believe the use is safe here, because:

- `item` is only received from the `c` channel, which is created and returned by [watch](https://github.com/rancher/steve/blob/2365164708b72f5c10dd5a38799543a4c28aa793/pkg/stores/proxy/proxy_store.go#L358)
- in turn, `watch` only passes `c` to [listAndWatch](https://github.com/rancher/steve/blob/2365164708b72f5c10dd5a38799543a4c28aa793/pkg/stores/proxy/proxy_store.go#L405) after its creation and before its closing. Therefore, `listAndWatch` is the only starting place where values can be sent to `c`
- `listAndWatch` only sends values in 5 places:
  - [three](https://github.com/rancher/steve/blob/2365164708b72f5c10dd5a38799543a4c28aa793/pkg/stores/proxy/proxy_store.go#L296) [of](https://github.com/rancher/steve/blob/2365164708b72f5c10dd5a38799543a4c28aa793/pkg/stores/proxy/proxy_store.go#L317) [them](https://github.com/rancher/steve/blob/2365164708b72f5c10dd5a38799543a4c28aa793/pkg/stores/proxy/proxy_store.go#L328) via the `returnErr` function, which [crafts a new watch.Event](https://github.com/rancher/steve/blob/2365164708b72f5c10dd5a38799543a4c28aa793/pkg/stores/proxy/proxy_store.go#L264C12-L269) with `Type == watch.Error` with an `api.Status` as its `Object`. Therefore these objects are safe to return
  - [two](https://github.com/rancher/steve/blob/2365164708b72f5c10dd5a38799543a4c28aa793/pkg/stores/proxy/proxy_store.go#L315) of [them](https://github.com/rancher/steve/blob/2365164708b72f5c10dd5a38799543a4c28aa793/pkg/stores/proxy/proxy_store.go#L337) via direct send. In both cases code checks immediately before that is not an error. Therefore in this case the code added by this PR is not executed


So that's my proof by exhaustion. I understand something other than a proof by exhaustion might be more desirable, but I'd argue such a bigger refactoring should be separate from this PR, which is meant to include the bugfix exclusively.

moio:
> I also have an overall question on the solution: my first glance would say that the ideal solution would be to immediately return a `resource.error` event when we are given an invalid resource version, rather than a `resource.start`, then a `resource.error`, then a `resource.stop`. I think that the idea here is just to align this with standard watch rather than re-evaluate the API behavior, but wanted to ask just to be sure - would you mind elaborating on why you choose the start -> error -> stop flow here?

As you mentioned, here I am proposing to align behavior with the `Watch` function. UI code does not differentiate the two cases, as from the UI perspective a watch call is a watch call - only Steve makes a distinction between `Watch` and `WatchNames` internally depending on permissions.

I think you are right in pointing out this particular flow could be optimized to be less chatty, but that would need to be done for `Watch` and `WatchNames`, aligning client UI code in the process. To me, that's a valid optimization/refactoring but it is a separate effort/PR from just fixing the bug. It should be done later, separately, and subject to an evaluation of importance - as this is could be a pretty rare case.

pkg/stores/proxy/proxy_store.go
MbolotSuse: Thanks for giving that a look. I think that I agree that we don't need function-level safety here, but I'm a bit concerned about one additional case that's not covered here. [listAndWatch](https://github.com/rancher/steve/blob/9935404019063126b6ac856db1ad56e0bc457ff8/pkg/stores/proxy/proxy_store.go#L329-L333) currently has a case that looks like it functions much the same as the case that you fixed does (i.e. debug the error and don't send on the channel). Would you mind seeing if you need to fix that as well? If so, how you handle that case will determine if this is safe since this will be another place where a value is sent on the channel,.

moio: It is indeed a similar. I'll open a follow-up PR to keep that case separate, as the user case I am working for isn't experiencing that particular error (I can tell from the debug logs not containing the string logged there).

moio:
@MbolotSuse following our discussions can you approve and merge this?

Relatedly: user reported success with the patch in the meantime.

pkg/stores/proxy/proxy_store_test.go
```diff
@@ -96,7 +96,13 @@ func receiveUntil(wc chan watch.Event, d time.Duration) error {
 			}
 
 			if event.Type == watch.Error {
-				return errors.New(fmt.Sprintf("watch chan should not have sent events of type [%s]", watch.Error))
+				status, ok := event.Object.(*metav1.Status)
```
MbolotSuse, line 99: I don't think this test is working quite right for a few reasons:
- The goal of the `recieveUntil` function seems to be to receive events from the channel, and then return once it gets all of the expected events.
  - Right now, the return part of the function [returns](https://github.com/rancher/steve/blob/2365164708b72f5c10dd5a38799543a4c28aa793/pkg/stores/proxy/proxy_store_test.go#L114) once it has received all expected secrets, but doesn't check for all received errors.
  - This causes the function to return even if it hasn't received the error yet.
  -  You can fix this by changing the linked line to check if `errMsgs` is empty as well.
- Right now you are checking if you got the exact error message. But this won't work because [listAndWatch](https://github.com/rancher/steve/blob/2365164708b72f5c10dd5a38799543a4c28aa793/pkg/stores/proxy/proxy_store.go#L328) is wrapping the error message. 
  - I think if you change L103 to check for containment/the end of the error message, that should do the trick.

When combined, these issues seem to mean that the test isn't checking if it got an error.

pkg/stores/proxy/proxy_store.go
MbolotSuse: I'll resolve this thread for now - if you can look at the other part relatively soon I would appreciate it, so that we don't forget about the issue here.

moio:
@MbolotSuse please review once again and, if all looks OK,merge this directly. Thanks in advance

moio:
Partially related, about your question which was left pending:

> [listAndWatch](https://github.com/rancher/steve/blob/9935404019063126b6ac856db1ad56e0bc457ff8/pkg/stores/proxy/proxy_store.go#L329-L333) currently has a case that looks like it functions much the same as the case that you fixed does (i.e. debug the error and don't send on the channel). Would you mind seeing if you need to fix that as well?

I took a deeper look. To recap, `listAndWatch` is the function producing events, including errors, consumed by `WatchNames`, the the subject of this PR. `WatchNames`, in turn, forwards events to the UI and this PR is about forwarding errors to the UI too, so that the UI can act on them. `Watch` does the same.

Therefore when `listAndWatch` receives an error (from the Kubernetes watch operation) it has two options:
 - log and ignore or
 - send it to the UI, by forwarding it `WatchNames`/`Watch`

At this point `listAndWatch` is forwarding all errors that can be type-asserted to `*v1.Status` - which is recommended but not guaranteed by the API:

https://github.com/kubernetes/apimachinery/blob/f14778da5523847e4c07346e3161a4b4f6c9186e/pkg/watch/watch.go#L67-L69

I think that approach makes sense - if Steve is at least able to extract a string or a code from the error then there is a chance the UI can do something about it, therefore forward (eg. the "resource too old" situation that kicked off this PR).

For an error whose type is completely unknown, there is no way to tell if it makes sense to forward to the UI or not and even if it does, there is no way to extract a meaningful string or code that the UI could check and react on a priori. It will need intervention from a Steve programmer anyway to add code for a specific type when such a case is found, and the best way to serve that future programmer is to have a reliable log.

What could be argued, IMO, is to raise the log level from debug to info or even error to increase the visibility of those cases, if they are important at all. ATM I do not have reasons to change that, but others might.

Please help me double-check the reasoning above. If it makes sense, I will not open a follow-up PR.

MbolotSuse (approved):

PR #146, opened 2024-01-22
Namespace schema does not correctly show permitted collectionMethods

votdev:
Fixes: https://github.com/rancher/rancher/issues/36799


pkg/schema/factory.go
```diff
@@ -162,6 +180,8 @@ func (c *Collection) schemasForSubject(access *accesscontrol.AccessSet) (*types.
 			continue
 		}
 
+		s.CollectionMethods = uniq(s.CollectionMethods)
```
votdev, line 183: Maybe this call can be moved to https://github.com/votdev/steve/blob/7a749b127d4fa57af225a7c209ee0d0bb7beed73/pkg/schema/factory.go#L166, but i'm not sure how it is possible to get 3 `GET` entries (see [here](https://github.com/rancher/rancher/issues/36799)) in the slice in that case (i assume only 2).

PR #158, opened 2024-02-09
Adding Access Set logic to the formatter

MbolotSuse:
Related to rancher/rancher#39129. 

## Overview

This PR changes the formatter commonly used by many resources to use an accessSet lookup to determine if the user has update/remove on a given object. If they don't, the corresponding link is removed.

### DIFF
--- a/pkg/resources/common/formatter.go
+++ b/pkg/resources/common/formatter.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"net/http"
 	"strings"
 
 	"github.com/rancher/apiserver/pkg/types"
@@ -12,7 +13,6 @@ import (
 	"github.com/rancher/steve/pkg/summarycache"
 	"github.com/rancher/wrangler/v2/pkg/data"
 	corecontrollers "github.com/rancher/wrangler/v2/pkg/generated/controllers/core/v1"
-	"github.com/rancher/wrangler/v2/pkg/slice"
 	"github.com/rancher/wrangler/v2/pkg/summary"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/resources/common/formatter.go
```diff
@@ -78,21 +78,35 @@ func formatter(summarycache *summarycache.SummaryCache) types.Formatter {
...
+		hasUpdate := accessSet.Grants("update", gvr.GroupResource(), resource.APIObject.Namespace(), resource.APIObject.Name())
+		hasDelete := accessSet.Grants("delete", gvr.GroupResource(), resource.APIObject.Namespace(), resource.APIObject.Name())
+
 		selfLink := selfLink(gvr, meta)
 
 		u := request.URLBuilder.RelativeToRoot(selfLink)
 		resource.Links["view"] = u
 
-		if _, ok := resource.Links["update"]; !ok && slice.ContainsString(resource.Schema.CollectionMethods, "PUT") {
-			resource.Links["update"] = u
```
JonCrowther, line 87: Previously we would set `resource.Links["update"] = u` when it `resource.Links["update"]` didn't exist. We're now assuming it should be explicitly set. While I do also prefer that style of only giving what is explicitly stated, is it possible that this may cause a regression? If a caller of the `DefaultTemplate` assumes the update portion will get filled in, after this change it won't. So in the UI example, maybe the "Edit" button is greyed out when it should be enabled.

MbolotSuse: I think that the underlying access control system would have set this value to something, so I'm not too concerned about the value not being there when it should be. 

That being said, I think you are right that this change could cause a regressions since `u` may (and I think is) different than what is previously set. I'll look at adding a new statement that overrides the update link in the case where we are allowed and not blocked. Good Catch.

MbolotSuse: I've made a change here to get this a bit closer to the old side. I think this still differs a bit from the old side (i.e. now if we know there's an update we add the link even if the underlying code never set it), but I think that this should do what we want. Please feel free to re-open if this doesn't fit with your thinking.

ericpromislow (changes requested):
After rebase, another call to `formatter()` will need to have the `AccessSet` arg, and this cascades up a couple of function calls.

ericpromislow:
Closing this ... please see https://github.com/rancher/steve/pull/308

PR #161, opened 2024-02-28
Resource schema improved cache

MbolotSuse:
## Overview

Followup to #139 and related to rancher/rancher#41301. This PR improves the caching mechanism used by the schema definitions endpoint to refresh once a CRD has been changed, rather than on request after a certain time period.

This also fixes a small bug in preferred versions - before, we selected the version that was preferred for a resource from the APIResourceList. We now source this information from the groups, like we do in other [areas of steve](https://github.com/rancher/steve/blob/641178e7cbc7edf18ff2c3dbc1f09ec863153d1a/pkg/schema/converter/discovery.go#L56).

## Solution Detail
- Adds a new `DebouncableRefresher` object and `Refreshable` interface. This allows a caller to queue multiple refreshes by repeated calls to "RefreshAfter" and only have the refresher run after the latest duration has elapsed.
- Adds a new "refreshHandler" which controls how often the schema definitions are refreshed. This triggers a refresh in 3 cases:
  - When a new CRD is added, mirroring what is currently done for [schemas](https://github.com/rancher/steve/blob/641178e7cbc7edf18ff2c3dbc1f09ec863153d1a/pkg/controllers/schema/schemas.go#L71).
  - When a new apiService is added, mirroring what is currently done for [schemas](https://github.com/rancher/steve/blob/641178e7cbc7edf18ff2c3dbc1f09ec863153d1a/pkg/controllers/schema/schemas.go#L70).
  - After a specified duration has passed (e.x. every 10 minutes).
- Wraps the schema definition handler in a `DebouncableRefresher`, and makes Refresh of the definitions exclusively handled by the refreshHandler. 
- Adds two new environment variables to allow for performance tuning on the refresh:
  - `CATTLE_CRD_REFRESH_DELAY_SECONDS` - the amount of time (in seconds) steve should wait after a CRD or API Service is created before attempting to refresh the schema definitions. Defaults to 2.
  - `CATTLE_BACKGROUND_REFRESH_MINUTES` - the amount of time (in minutes) steve should wait between force refreshes of the definitions (note that a force refresh is not delayed by the creation of a CRD/API Service, though it still does use the same debounce refresher when requesting a refresh). Defaults to 10.
- Changes the `indexSchemaNames` logic to derive preferredVersion from the group a resource belongs to. 
  - This no longer checks for a `ErrGroupDiscoveryFailed` - from what I saw in the source, this error seemed to be in cases where client-go could find information for a group but wasn't able to get information on resources in a group. Since we only need group information, this isn't necessary anymore. 
  - Since this doesn't return a partially-filled-in map anymore, some of the calling functions were changed to just bubble the error up.
- Tests were also added for the above.

### DIFF
--- a/pkg/debounce/refresher.go
+++ b/pkg/debounce/refresher.go
@@ -1,0 +1,51 @@
+package debounce
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+// Refreshable represents an object which can be refreshed. This should be protected by a mutex for concurrent operation.
+type Refreshable interface {
+	Refresh() error
+}
+
+// DebounceableRefresher is used to debounce multiple attempts to refresh a refreshable type.
+type DebounceableRefresher struct {
+	sync.Mutex
+	// Refreshable is any type that can be refreshed. The refresh method should by protected by a mutex internally.
+	Refreshable Refreshable
+	current     context.CancelFunc
+}
+
+// RefreshAfter requests a refresh after a certain time has passed. Subsequent calls to this method will
+// delay the requested refresh by the new duration. Note that this is a total override of the previous calls - calling
+// RefreshAfter(time.Second * 2) and then immediately calling RefreshAfter(time.Microsecond * 1) will run a refresh
+// in one microsecond
+func (d *DebounceableRefresher) RefreshAfter(duration time.Duration) {
+	d.Lock()
+	defer d.Unlock()
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	if d.current != nil {
+		d.current()
+	}
+	d.current = cancel
+	go func() {
+		timer := time.NewTimer(duration)
+		defer timer.Stop()
+		select {
+		case <-ctx.Done():
+			// this indicates that the context was cancelled. Do nothing.
+		case <-timer.C:
+			// note this can cause multiple refreshes to happen concurrently
+			err := d.Refreshable.Refresh()
+			if err != nil {
+				logrus.Errorf("failed to refresh with error: %v", err)
+			}
+		}
+	}()
+}
--- a/pkg/debounce/refresher_test.go
+++ b/pkg/debounce/refresher_test.go
@@ -1,0 +1,47 @@
+package debounce
+
+import (
+	"fmt"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+type refreshable struct {
+	wasRefreshed atomic.Bool
+	retErr       error
+}
+
+func (r *refreshable) Refresh() error {
+	r.wasRefreshed.Store(true)
+	return r.retErr
+}
+
+func TestRefreshAfter(t *testing.T) {
+	ref := refreshable{}
+	debounce := DebounceableRefresher{
+		Refreshable: &ref,
+	}
+	debounce.RefreshAfter(time.Millisecond * 2)
+	debounce.RefreshAfter(time.Microsecond * 2)
+	time.Sleep(time.Millisecond * 1)
+	// test that the second refresh call overrode the first - Micro < Milli so this should have ran
+	require.True(t, ref.wasRefreshed.Load())
+	ref.wasRefreshed.Store(false)
+	time.Sleep(time.Millisecond * 2)
+	// test that the call was debounced - though we called this twice only one refresh should be called
+	require.False(t, ref.wasRefreshed.Load())
+
+	ref = refreshable{
+		retErr: fmt.Errorf("Some error"),
+	}
+	debounce = DebounceableRefresher{
+		Refreshable: &ref,
+	}
+	debounce.RefreshAfter(time.Microsecond * 2)
+	// test the error case
+	time.Sleep(time.Millisecond * 1)
+	require.True(t, ref.wasRefreshed.Load())
+}
--- a/pkg/schema/definitions/handler.go
+++ b/pkg/schema/definitions/handler.go
@@ -1,7 +1,6 @@
 package definitions
 
 import (
-	"errors"
 	"fmt"
 	"net/http"
 	"sync"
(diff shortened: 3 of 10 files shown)


pkg/schema/definitions/handler.go
```diff
@@ -46,24 +38,40 @@ type schemaDefinitionHandler struct {
 	schemaToModel map[string]string
 }
 
+// Refresh writeLocks and updates the cache with new schemaDefinitions. Will result in a call to kubernetes to retrieve
+// the openAPI schemas.
+func (s *SchemaDefinitionHandler) Refresh() error {
+	s.Lock()
+	defer s.Unlock()
```
moio, line 45: IIUC this locks readers for the whole duration of the OpenAPI schema refresh.

Could we lock later, ideally only when all OpenAPI calls and network I/O are finished, so that most of the refresh won't block?

pkg/debounce/refresher.go
```diff
@@ -0,0 +1,51 @@
...
+	d.current = cancel
+	go func() {
+		timer := time.NewTimer(duration)
+		defer timer.Stop()
+		select {
+		case <-ctx.Done():
+			// this indicates that the context was cancelled. Do nothing.
+		case <-timer.C:
+			// note this can cause multiple refreshes to happen concurrently
+			err := d.Refreshable.Refresh()
```
moio, line 45: Would it help to pass `ctx` into `Refresh`, so that it's notified if it's being canceled?

Maybe this is something for a future PR, not sure if it will be needed.

pkg/schema/definitions/refresh.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/rancher/steve/pkg/debounce"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiregv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
+)
+
+// refreshHandler triggers refreshes for a DebounceAble refresher after a CRD/APIService has been changed
```
moio, line 12:
```suggestion
// refreshHandler triggers refreshes for a Debounceable refresher after a CRD/APIService has been changed
```

moio (commented):
A couple of questions and a couple of nitpicks. Overall great work that goes in the right direction, thanks!

pkg/schema/definitions/handler.go
MbolotSuse: I think that you are correct. In the previous implementation I think this had value since the refresh was initiated by the refresh, so the change you are talking about could have resulted in multiple calls to the openapi endpoints on the k8s side. 

However, since the refresh is no longer user-initiated I think the approach that you are talking about will work much better. I'll swap this around.

pkg/debounce/refresher.go
MbolotSuse: I think in this case there's limited value here since the Refresh function that we are actually calling doesn't consume/use a context. I think this would be an enhancement for a future PR if/when we come up with a refresh function that uses a context. 

WDYT?

pkg/debounce/refresher_test.go
```diff
@@ -0,0 +1,44 @@
...
+	wasRefreshed atomic.Bool
+	retErr       error
+}
+
+func (r *refreshable) Refresh() error {
+	r.wasRefreshed.Store(true)
+	return r.retErr
+}
+
+func TestRefreshAfter(t *testing.T) {
```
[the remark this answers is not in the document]
MbolotSuse: Ordinarily I would be opposed to this, but agree that it would be valuable here since there's no table.

pkg/debounce/refresher.go
moio: I am completely OK treating this as a future enhancement as well

pkg/debounce/refresher.go
```diff
@@ -0,0 +1,51 @@
...
+// delay the requested refresh by the new duration. Note that this is a total override of the previous calls - calling
+// RefreshAfter(time.Second * 2) and then immediately calling RefreshAfter(time.Microsecond * 1) will run a refresh
+// in one microsecond
+func (d *DebouncableRefresher) RefreshAfter(duration time.Duration) {
+	d.Lock()
+	defer d.Unlock()
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	if d.current != nil {
+		d.current()
```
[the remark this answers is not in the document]
MbolotSuse: I thought about this a bit. My thinking is:
- The cancel doesn't affect anything else, but it is closing the previously opened channel, which is not closed when the ticker calls the refresh.
- Trying to nil out the previous value runs the risk of improperly closing resources (causing scaling issues) or accidentally causing a nil pointer.

Due to the above, I think that the current functionality is working as intended. If you disagree, let me know, I can look at it further.

pkg/debounce/refresher_test.go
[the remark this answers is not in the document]
MbolotSuse: I've added some comments, please re-open if this isn't enough.

pkg/debounce/refresher.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// Refreshable represents an object which can be refreshed. This should be protected by a mutex for concurrent operation.
+type Refreshable interface {
+	Refresh() error
+}
+
+// DebouncableRefresher is used to debounce multiple attempts to refresh a refreshable type.
+type DebouncableRefresher struct {
```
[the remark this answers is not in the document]
MbolotSuse: I've made this change. I think I got all of the references and comments, LMK if you see one that I missed.

pkg/schema/definitions/handler.go
MbolotSuse: Change made. This also allowed me to clean up some areas around the function calls/interfaces - please let me know if you see anything else here.

JonCrowther (approved):

moio (approved):
LGTM

PR #166, opened 2024-03-04
changing permission for kubeconfig file inside the shell pod

diogoasouza:
This PR changes the permission of the kubeconfig file used in the `shell` container used by the impersonator. It was using 0644 which was causing warnings when using helm. This changes the permission to 0600.

**Explanation:**
Kubernetes always [mounts](https://github.com/kubernetes/kubernetes/blob/46a2137c1ba017970c316c0ec10c074cb6450732/pkg/kubelet/kubelet_pods.go#L344) [configMaps](https://github.com/kubernetes/kubernetes/blob/46a2137c1ba017970c316c0ec10c074cb6450732/pkg/volume/configmap/configmap.go#L164) and `Secret` volumes as `readOnly`, regardless of the flag passed when creating the mount. This is [not reflected](https://github.com/kubernetes/api/blob/dd2e75089b339ee9b1b2aed2c275bf905079dde3/core/v1/types.go#L2155) in the documentation of the `readOnly` flag. To work around that, I'm using an `init-container` to copy the content to an `emptyDir` volume, changing the permission and using that volume in the `shell` container

### DIFF
--- a/pkg/podimpersonation/podimpersonation.go
+++ b/pkg/podimpersonation/podimpersonation.go
@@ -512,10 +512,12 @@ func (s *PodImpersonation) adminKubeConfig(user user.Info, role *rbacv1.ClusterR
 
 func (s *PodImpersonation) augmentPod(pod *v1.Pod, sa *v1.ServiceAccount, secret *v1.Secret, imageOverride string) *v1.Pod {
 	var (
-		zero = int64(0)
-		t    = true
-		f    = false
-		m    = int32(420)
+		zero      = int64(0)
+		t         = true
+		f         = false
+		m         = int32(0o644)
+		m2        = int32(0o600)
+		shellUser = 1000
 	)
 
 	pod = pod.DeepCopy()
@@ -536,10 +538,17 @@ func (s *PodImpersonation) augmentPod(pod *v1.Pod, sa *v1.ServiceAccount, secret
 		v1.Volume{
 			Name: "user-kubeconfig",
 			VolumeSource: v1.VolumeSource{
+				EmptyDir: &v1.EmptyDirVolumeSource{},
+			},
+		},
+		v1.Volume{
+			Name: "user-kube-configmap",
+			VolumeSource: v1.VolumeSource{
 				ConfigMap: &v1.ConfigMapVolumeSource{
 					LocalObjectReference: v1.LocalObjectReference{
 						Name: s.userConfigName(),
 					},
+					DefaultMode: &m2,
 				},
 			},
 		},
@@ -553,25 +562,50 @@ func (s *PodImpersonation) augmentPod(pod *v1.Pod, sa *v1.ServiceAccount, secret
 			},
 		})
 
+	image := imageOverride
+	if image == "" {
+		image = s.imageName()
+	}
+
 	for i, container := range pod.Spec.Containers {
 		for _, envvar := range container.Env {
 			if envvar.Name != "KUBECONFIG" {
 				continue
 			}
 
+			vmount := v1.VolumeMount{
+				Name:      "user-kubeconfig",
+				MountPath: "/tmp/.kube",
+			}
+			cfgVMount := v1.VolumeMount{
+				Name:      "user-kube-configmap",
+				MountPath: "/home/.kube/config",
+				SubPath:   "config",
+			}
+
+			pod.Spec.InitContainers = []v1.Container{
+				{
+					Name:            "init-kubeconfig-volume",
+					Image:           image,
+					Command:         []string{"sh", "-c", fmt.Sprintf("cp %s %s && chown %d %s/config", cfgVMount.MountPath, vmount.MountPath, shellUser, vmount.MountPath)},
+					Resources:       v1.ResourceRequirements{},
+					ResizePolicy:    nil,
+					ImagePullPolicy: v1.PullIfNotPresent,
+					SecurityContext: &v1.SecurityContext{
+						RunAsUser:  &zero,
+						RunAsGroup: &zero,
+					},
+					VolumeMounts: []v1.VolumeMount{cfgVMount, vmount},
+				},
+			}
+
 			pod.Spec.Containers[i].VolumeMounts = append(container.VolumeMounts, v1.VolumeMount{
 				Name:      "user-kubeconfig",
-				ReadOnly:  true,
 				MountPath: envvar.Value,
 				SubPath:   "config",
 			})
 			break
 		}
-	}
-
-	image := imageOverride
-	if image == "" {
-		image = s.imageName()
 	}
 
 	pod.Spec.Containers = append(pod.Spec.Containers, v1.Container{
--- a/pkg/podimpersonation/podimpersonation.go
+++ b/pkg/podimpersonation/podimpersonation.go
@@ -573,6 +573,9 @@ func (s *PodImpersonation) augmentPod(pod *v1.Pod, sa *v1.ServiceAccount, secret
 				continue
 			}
 
+			//This mounts two volumes, one configMap and one emptyDir.
+			//The reason for this is that we need to change the permissions on the kubeconfig file
+			//and, since a configMap volume is always read-only, we need an emptyDir volume as well.
 			vmount := v1.VolumeMount{
 				Name:      "user-kubeconfig",
 				MountPath: "/tmp/.kube",
@@ -583,21 +586,18 @@ func (s *PodImpersonation) augmentPod(pod *v1.Pod, sa *v1.ServiceAccount, secret
 				SubPath:   "config",
 			}
 
-			pod.Spec.InitContainers = []v1.Container{
-				{
-					Name:            "init-kubeconfig-volume",
-					Image:           image,
-					Command:         []string{"sh", "-c", fmt.Sprintf("cp %s %s && chown %d %s/config", cfgVMount.MountPath, vmount.MountPath, shellUser, vmount.MountPath)},
-					Resources:       v1.ResourceRequirements{},
-					ResizePolicy:    nil,
-					ImagePullPolicy: v1.PullIfNotPresent,
-					SecurityContext: &v1.SecurityContext{
-						RunAsUser:  &zero,
-						RunAsGroup: &zero,
-					},
-					VolumeMounts: []v1.VolumeMount{cfgVMount, vmount},
+			pod.Spec.InitContainers = append(pod.Spec.InitContainers, v1.Container{
+				Name:            "init-kubeconfig-volume",
+				Image:           image,
+				Command:         []string{"sh", "-c", fmt.Sprintf("cp %s %s && chown %d %s/config", cfgVMount.MountPath, vmount.MountPath, shellUser, vmount.MountPath)},
+				ImagePullPolicy: v1.PullIfNotPresent,
+				SecurityContext: &v1.SecurityContext{
+					RunAsUser:  &zero,
+					RunAsGroup: &zero,
 				},
-			}
+				VolumeMounts: []v1.VolumeMount{cfgVMount, vmount},
+			},
+			)
 
 			pod.Spec.Containers[i].VolumeMounts = append(container.VolumeMounts, v1.VolumeMount{
 				Name:      "user-kubeconfig",
--- a/pkg/podimpersonation/podimpersonation_test.go
+++ b/pkg/podimpersonation/podimpersonation_test.go
@@ -1,0 +1,91 @@
+package podimpersonation
+
+import (
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"testing"
+	"time"
+)
+
+func TestAugmentPod(t *testing.T) {
+	var (
+		zero = int64(0)
+	)
+	testCases := []struct {
+		name          string
+		imageOverride string
+		envVars       []v1.EnvVar
+	}{
+		{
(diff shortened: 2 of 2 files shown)


pkg/podimpersonation/podimpersonation.go
```diff
@@ -559,9 +568,33 @@ func (s *PodImpersonation) augmentPod(pod *v1.Pod, sa *v1.ServiceAccount, secret
...
+			cfgVMount := v1.VolumeMount{
+				Name:      "user-kubeconfig-map",
+				MountPath: "/home/.kube/config",
+				SubPath:   "config",
+			}
+
+			pod.Spec.InitContainers = []v1.Container{
+				{
+					Name:         "init-container",
+					Image:        "busybox",
```
rohitsakala, line 584: Should we use our own busybox ? rancher/mirrored-bci-busybox or rancher/library-busybox. I don't know which is the latest one. Maybe ask @pjbgf ?

pkg/podimpersonation/podimpersonation.go
```diff
@@ -535,11 +537,18 @@ func (s *PodImpersonation) augmentPod(pod *v1.Pod, sa *v1.ServiceAccount, secret
 		},
 		v1.Volume{
 			Name: "user-kubeconfig",
+			VolumeSource: v1.VolumeSource{
+				EmptyDir: &v1.EmptyDirVolumeSource{},
+			},
+		},
+		v1.Volume{
+			Name: "user-kubeconfig-map",
```
rohitsakala, line 545: nit: user-kube-configmap to

pkg/podimpersonation/podimpersonation.go
```diff
@@ -559,9 +568,33 @@ func (s *PodImpersonation) augmentPod(pod *v1.Pod, sa *v1.ServiceAccount, secret
...
+					},
+					VolumeMounts: []v1.VolumeMount{cfgVMount, vmount},
+				},
+			}
+
 			pod.Spec.Containers[i].VolumeMounts = append(container.VolumeMounts, v1.VolumeMount{
 				Name:      "user-kubeconfig",
-				ReadOnly:  true,
 				MountPath: envvar.Value,
 				SubPath:   "config",
```
rohitsakala, line 599: Do we need this subpath?

pkg/podimpersonation/podimpersonation.go
pjbgf: `registry.suse.com/bci/bci-busybox`.

pkg/podimpersonation/podimpersonation.go
diogoasouza: We only use what's inside the `config` key of the ConfigMap. We don't _need_ it, but I think it's good to specify a `SubPath` if we know exactly what we need

pkg/podimpersonation/podimpersonation.go
diogoasouza: Good call, done

pkg/podimpersonation/podimpersonation.go
diogoasouza: Done

pkg/podimpersonation/podimpersonation.go
```diff
@@ -512,10 +512,12 @@ func (s *PodImpersonation) adminKubeConfig(user user.Info, role *rbacv1.ClusterR
...
 	var (
-		zero = int64(0)
-		t    = true
-		f    = false
-		m    = int32(420)
+		zero      = int64(0)
+		t         = true
+		f         = false
+		m         = int32(420)
+		m2        = int32(384)
```
pjbgf, line 519: Using Go's octal syntax makes it slightly more readable:
```suggestion
		m         = 0o644
		m2        = 0o600
```

diogoasouza: Done

diogoasouza:
Can I get a review from someone on the frameworks team?
@rancher/rancher-squad-frameworks

pkg/podimpersonation/podimpersonation.go
```diff
@@ -559,9 +568,33 @@ func (s *PodImpersonation) augmentPod(pod *v1.Pod, sa *v1.ServiceAccount, secret
...
+			}
+			cfgVMount := v1.VolumeMount{
+				Name:      "user-kube-configmap",
+				MountPath: "/home/.kube/config",
+				SubPath:   "config",
+			}
+
+			pod.Spec.InitContainers = []v1.Container{
+				{
+					Name:         "init-container",
```
rohitsakala, line 583: Can we rename to something descriptive ? instead of `init-container`

pkg/podimpersonation/podimpersonation.go
```diff
@@ -559,9 +568,33 @@ func (s *PodImpersonation) augmentPod(pod *v1.Pod, sa *v1.ServiceAccount, secret
...
+				MountPath: "/tmp/.kube",
+			}
+			cfgVMount := v1.VolumeMount{
+				Name:      "user-kube-configmap",
+				MountPath: "/home/.kube/config",
+				SubPath:   "config",
+			}
+
+			pod.Spec.InitContainers = []v1.Container{
+				{
```
rohitsakala, line 582: Due to the new `init-container` when I click on the `Kubectl shell` button on Rancher Manager, it is taking 18 seconds to load the shell.

Before the `init-container` it takes only 4 seconds to load the shell.

This is a serious degradation of user experience.

Maybe due to this, I find the `sed` solution better, if we can't find a way to mitigate this user experience issue.

tomleb:
Can you link a GH issue associated with this change?

I'm also not very familiar with this area but I'll do my best at reviewing this.

 From what I understand, we want to mount some kubeconfig that's stored on a `ConfigMap` and we want it to have `0600` privileges to avoid a warning from the `helm` binary.

Is it not possible to mount this file with the `defaultMode` set to 0600? I see you're using it for the one in the init container, but why not in the container itself directly?

Here's an example (taken and modified from the upstream k8s docs):

```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: game-config
  namespace: default
data:
  game.properties: |
    enemies=aliens
    lives=3
    enemies.cheat=true
    enemies.cheat.level=noGoodRotten
    secret.code.passphrase=UUDDLRLRBABAS
    secret.code.allowed=true
    secret.code.lives=30    
  ui.properties: |
    color.good=purple
    color.bad=yellow
    allow.textmode=true
    how.nice.to.look=fairlyNice    
---
apiVersion: v1
kind: Pod
metadata:
  name: dapi-test-pod
spec:
  containers:
    - name: test-container
      image: registry.k8s.io/busybox
      command: [ "/bin/sh", "-c", "ls -l /etc/config/" ]
      volumeMounts:
      - name: game-config
        subPath: game.properties
        mountPath: /etc/config/game.properties
  volumes:
    - name: game-config
      configMap:
        name: game-config
        defaultMode: 0600
  restartPolicy: Never
```

Looking at the logs of the pod:

```sh
$ kubectl logs dapi-test-pod
total 4
-rw-------    1 root     root           162 Mar  8 13:12 game.properties
```

The mode bits are as expected.

diogoasouza:
@tomleb Hi Tom, thanks for the review. 
Regarding that, it's true that we can change the `defaultMode`, the problem is that the user that [runs](https://github.com/rancher/shell/blob/b88ba63c4a0da560fa99f9ca9a357023440c0b42/package/Dockerfile#L71) inside the `shell` container is not the `root` user. Setting the `runAsUser` in the container does not change the ownership of the files inside the mounted volume as well

pkg/podimpersonation/podimpersonation.go
diogoasouza: It was the `imagePullPolicy` that was causing this. The default value is `Always` and I changed it to `PullIfNotPresent`. Should be fine now

pkg/podimpersonation/podimpersonation.go
diogoasouza: Done

pkg/podimpersonation/podimpersonation.go
```diff
@@ -559,9 +568,34 @@ func (s *PodImpersonation) augmentPod(pod *v1.Pod, sa *v1.ServiceAccount, secret
...
+					Name:            "init-kubeconfig-volume",
+					Image:           "registry.suse.com/bci/bci-busybox",
```
tomleb, line 584: I'm not a fan of using `:latest` tags for images. Something that used to work might stop working randomly if that `:latest` image breaks.

diogoasouza: Makes sense. Changed it to use `registry.suse.com/bci/bci-busybox:15.5.16.1` which is the latest one available as of today

pjbgf: +1 on not using `latest`. Overall we tend to recommend folks tag to the minor: `registry.suse.com/bci/bci-busybox:15.5`. So you latest patches, without the risk of breaking changes.

tomleb: Michael reminded me about airgap. It seems like hardcoding `registry.suse.com/bci/bci-busybox:whatever` would not work for airgap use cases where the customer does not have access to the SUSE registry.

@diogoasouza one idea here would be to re-use the shell image but change its entrypoint via `command:`. That would avoid making the busybox available for airgap use case (no idea if it's already done btw).

diogoasouza: @tomleb I checked and the busybox image is mirrored, but only the `15.4` version, and that was more than one year ago. I changed the init container to use the shell image, it shouldn't affect the latency since the image is going to have to be pulled anyway for the `proxy` and `shell` containers.

pkg/podimpersonation/podimpersonation.go
diogoasouza: @rohitsakala Yes, sorry if I wasn't clear. I meant that we only copy the `config` from the `configMap` to this volume and that's why we only need to mount this path to the container

rohitsakala (approved):

pkg/podimpersonation/podimpersonation.go
```diff
@@ -559,9 +568,34 @@ func (s *PodImpersonation) augmentPod(pod *v1.Pod, sa *v1.ServiceAccount, secret
...
+			cfgVMount := v1.VolumeMount{
+				Name:      "user-kube-configmap",
+				MountPath: "/home/.kube/config",
+				SubPath:   "config",
+			}
+
+			pod.Spec.InitContainers = []v1.Container{
+				{
+					Name:            "init-kubeconfig-volume",
+					Image:           s.imageName(),
```
tomleb, line 584: This should probably respect the[ image override that is done further down](https://github.com/rancher/steve/blob/dcea050d923061f427bb3f7e0a6a8f2a2d2992b9/pkg/podimpersonation/podimpersonation.go#L606-L609) in this function.

diogoasouza: Done

rohitsakala (approved):

pkg/podimpersonation/podimpersonation.go
```diff
@@ -553,27 +562,49 @@ func (s *PodImpersonation) augmentPod(pod *v1.Pod, sa *v1.ServiceAccount, secret
...
+		image = s.imageName()
+	}
+
 	for i, container := range pod.Spec.Containers {
 		for _, envvar := range container.Env {
 			if envvar.Name != "KUBECONFIG" {
 				continue
 			}
 
+			vmount := v1.VolumeMount{
```
tomleb, line 576: nit: Could add a comment to explain why we're doing this since it's not obvious at first

tomleb (approved):
One nit but otherwise LGTM.

Thanks for adding some tests.

tomleb:
Needs one more approval from frameworks folks.

Also this PR will need someone from your team to QA it :+1:

diogoasouza:
> Needs one more approval from frameworks folks.
> 
> Also this PR will need someone from your team to QA it 👍

Sure! I'll put it in our QA backlog after it's merged

pkg/podimpersonation/podimpersonation_test.go
```diff
@@ -0,0 +1,92 @@
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			p := newPod(tc.envVars)
+			impersonator := New("", nil, time.Minute, func() string { return "rancher/shell:v0.1.22" })
+
+			pod := impersonator.augmentPod(p, nil, &v1.Secret{ObjectMeta: metav1.ObjectMeta{Name: "s"}}, tc.imageOverride)
+
+			assert.Len(t, pod.Spec.Volumes, len(p.Spec.Volumes)+4, "expected four new volumes")
+			if len(tc.envVars) != 0 {
+				assert.Len(t, pod.Spec.Containers[0].VolumeMounts, len(p.Spec.Containers[0].VolumeMounts)+1, "expeted kubeconfig volume to be mounted")
```
JonCrowther, line 46:
```suggestion
				assert.Len(t, pod.Spec.Containers[0].VolumeMounts, len(p.Spec.Containers[0].VolumeMounts)+1, "expected kubeconfig volume to be mounted")
```

JonCrowther (commented):
Everything looks good. Just found a minor typo

pkg/podimpersonation/podimpersonation_test.go
diogoasouza: Fixed it

JonCrowther (approved):

diogoasouza:
I think it has the required approvals, can you merge it @tomleb ?

PR #182, opened 2024-04-09
First patch towards showing the apps in the UI for non-admin users:

ericpromislow:
It doesn't make sense to return an empty list when all namespaces are selected. With this change
`curl https://DOMAIN/k8s/clusters/CLUSTER/v1/catalog.cattle.io.apps` shows all the apps the user has permission to see, but they don't show up in the UI.

Partial fix for issue [#43036](https://github.com/rancher/rancher/issues/43036)


ericpromislow:
backport

ericpromislow:
And this change isn't needed for both 2.8 and 2.9 -- see dashboard issue https://github.com/rancher/dashboard/issues/10866 for 2.8 only

PR #184, opened 2024-04-10
Fix the missing-apps problem for rancher 2.8

ericpromislow:
It doesn't make sense to return an empty list when all namespaces are selected. With this change
`curl https://DOMAIN/k8s/clusters/CLUSTER/v1/catalog.cattle.io.apps` shows all the apps the user has permission to see, but they don't show up in the UI.

Partial fix for issue https://github.com/rancher/rancher/issues/43036

There are build issues with rancher 2.9, and then I can't create a digital ocean cluster, so copying this PR from 182 for (`rancher release/v2.9` , `steve master`).

### DIFF
--- a/pkg/stores/proxy/proxy_store.go
+++ b/pkg/stores/proxy/proxy_store.go
@@ -204,12 +204,6 @@ func tableToObjects(obj map[string]interface{}) []unstructured.Unstructured {
 // With this filter, the request can be performed successfully, and only the allowed resources will
 // be returned in the list.
 func (s *Store) ByNames(apiOp *types.APIRequest, schema *types.APISchema, names sets.String) (*unstructured.UnstructuredList, []types.Warning, error) {
-	if apiOp.Namespace == "*" {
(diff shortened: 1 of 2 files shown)


ericpromislow:
This change is not needed, and possibly incorrect. Using `curl` from the command-line shows that this code returns the correct apps for both 2.8 and 2.9. The apps appear correctly in the UI in 2.9, but not in 2.8.  

I've written up the issue at https://github.com/rancher/dashboard/issues/10866

Code-related details (referring to `steve/pkg/stores/proxy/{rbac_store.go,proxy_store.go}:

For both 2.8 and 2.9 (and later versions of 2.7 with the null-pointer fix):

I haven't worked out the calling sequence, but when logged in as a standard user there are three calls to Store.List (sorry, don't know which instance of Store) -- the call to Store.B8yNames returns 0 items, but there are two other calls to s.list with different schema parameters, I assume for each of the user namespaces,  and they return the correct values. They don't go through `Store.ByNames -- they go through Store.list => k8sClient.List. So I would mark this as closed in 2.9 and needing a dashboard backport for 2.8, as curl is behaving the same

PR #194, opened 2024-04-12
Add multiple base branches and a rancher package group

JonCrowther:


MbolotSuse (approved):

tomleb (commented):
We may want to increase the `prHourlyLimit` because with more branches, it might take a while before we get PRs for the `master` branch if renovate is trying to update the other base branches.

Since steve runs renovate twice during the morning, that's maximum 4 PRs per day, which isn't too bad with 1 branch but with 3 branch might not be enough

PR #202, opened 2024-04-26
Fixing schema definitions bugs

MbolotSuse:
## Overview

This is related to rancher/rancher#45158, and resolves two of the issues identified in the [issue summary](https://github.com/rancher/rancher/issues/45158#issuecomment-2069689829). Specifically:
- There are some resources (e.x. `generateKubeconfigOutput`) which are added as custom schemas - these don't represent real k8s resources. However, since they are still top-level schemas, they need definitions.
- Some resources (e.x. `monitoring.coreos.com.alertmanagerconfig` from the rancher-monitoring chart) are not in the preferred version of the group, but are in some version of the group. `monitoring.coreos.com.alertmanagerconfig`, for example, appears in the `v1alpha1` version, but not in the `v1` version which is the preferred version of the group.

This does not address the third issue - this should be addressed by the work on rancher/rancher#45157, since these resources represent CRDs that have no typing information in kubernetes.

## Solution Detail

- We now add the `baseSchema` configured for the server to the SchemaDefinitionHandler. This contains the configured, built-in schemas (such as [generateKubeconfigOutput](https://github.com/rancher/rancher/blob/381540fbb7ded6961f72465e8d11fa1b56742a6a/pkg/api/steve/clusters/clusters.go#L56)).
- When processing a request, if the schema is a baseSchema (as determined by trying to lookup the schema from the baseSchemas), then we extract the definition from the schema's resourceFields rather than from the openapi models.
  - In these cases, we still extract the definition from the schema on the request rather than the `baseSchema` (which is just used to determine if this resource is a base schema). This is done in case the schema has been transformed for this user - baseSchemas are setup-wide, where the schemas on the request are specific to the user.
  - This approach requires that the schema still has the resource fields set. For this reason, these resources still have `resourceFileds` in `/v1/schemas`. Since the impact of this is relatively low, this wasn't resolved in this PR.
  - This transformation is done at the time the schema is called, rather than at refresh time, so there's some overhead in processing these requests. However, that overhead is comparable (and probably lower) to what we do in recursively gathering the schemas on a proper k8s resource.
  - This approach requires that we parse the `type` field on the resource fields for cases of complex types like `map[string]`. This is done using an import from wrangler, which is used for similar purposes in [rancher](https://github.com/rancher/rancher/blob/0b9e86cf33297eb458b3a9ac7e8a53972c0062e9/pkg/schemas/mapper/namespace_reference.go#L63)
- The logic for preferred group/version was tweaked slightly. 
  - In the current/`master` state, resources are only added when they are in the preferred version for a group.  
  - In the new state, we add a resource if one of the two is true:
    - The resource is in the preferred version for a group
    - We don't have a model for that schema yet
  - This ensures that if we have a preferred version of a resource, that version is used (since it is always added, it will override any other non-preferred version). However, if none exists, we still have a version of the resource, so we can still get a legitimate definition.

## Notes
Tests were added for the above cases to maintain coverage.

### DIFF
--- a/pkg/schema/definitions/handler.go
+++ b/pkg/schema/definitions/handler.go
@@ -8,6 +8,7 @@ import (
 	"github.com/rancher/apiserver/pkg/apierror"
 	"github.com/rancher/apiserver/pkg/types"
 	"github.com/rancher/steve/pkg/schema/converter"
+	wranglerDefinition "github.com/rancher/wrangler/v2/pkg/schemas/definition"
 	"github.com/rancher/wrangler/v2/pkg/schemas/validation"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/discovery"
@@ -30,6 +31,8 @@ var (
 type SchemaDefinitionHandler struct {
 	sync.RWMutex
 
+	// baseSchema are the schemas (which may not represent a real CRD) added to the server
+	baseSchema *types.APISchemas
 	// client is the discovery client used to get the groups/resources/fields from kubernetes.
 	client discovery.DiscoveryInterface
 	// models are the cached models from the last response from kubernetes.
@@ -76,6 +79,20 @@ func (s *SchemaDefinitionHandler) byIDHandler(request *types.APIRequest) (types.
 	// would make this endpoint only usable by one caller at a time
 	s.RLock()
 	defer s.RUnlock()
+
+	if baseSchema := s.baseSchema.LookupSchema(requestSchema.ID); baseSchema != nil {
+		// if this schema is a base schema it won't be in the model cache. In this case, and only this case, we process
+		// the fields independently
+		definitions := baseSchemaToDefinition(*requestSchema)
+		return types.APIObject{
+			ID:   request.Name,
+			Type: "schemaDefinition",
+			Object: schemaDefinition{
+				DefinitionType: requestSchema.ID,
+				Definitions:    definitions,
+			},
+		}, nil
+	}
 
 	if s.models == nil {
 		return types.APIObject{}, apierror.NewAPIError(notRefreshedErrorCode, "schema definitions not yet refreshed")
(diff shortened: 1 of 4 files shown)


pkg/schema/definitions/handler.go
```diff
@@ -130,13 +147,52 @@ func (s *SchemaDefinitionHandler) indexSchemaNames(models proto.Models, groups *
...
+
+// parseFieldType parses a schemas.Field's type to a type (first return) and subType (second return)
+func parseFieldType(fieldType string) (string, string) {
+	splitted := strings.Split(fieldType, "[")
+	if len(splitted) == 2 && len(splitted[1]) > 0 {
+		// example: map[someVal] - split removes the leading "[" trim removes the trailing "]"
+		return splitted[0], strings.TrimSuffix(splitted[1], "]")
+	}
+	return fieldType, ""
+}
```
tomleb, line 198: This somewhat duplicates the functionality here: https://github.com/rancher/wrangler/blob/v2.2.0-rc6/pkg/schemas/definition/definition.go#L25. Wrangler doesn't have a function to get both types and subtypes though, a bit unfortunate.

Though a mix of `wrangler.Subtype` and `wrangler.Is{Map,Array,Reference}Type` should do the trick.

WDYT?

pkg/schema/definitions/handler.go
```diff
@@ -130,13 +147,52 @@ func (s *SchemaDefinitionHandler) indexSchemaNames(models proto.Models, groups *
...
+				schemaToModel[schemaID] = modelName
+			}
 		}
 	}
 	return schemaToModel
 }
+
+// baseSchemaToDefinition converts a given schema to the definition map. This should only be used with baseSchemas, whose definitions
+// are expected to be set by another application and may not be k8s resources.
+func baseSchemaToDefinition(schema types.APISchema) map[string]definition {
```
tomleb, line 170: Is my understanding correct that this doesn't need to be recursive because a field has either a primitive type or:
- array whose subtype is primitive
- map whose subtype is primitive
- reference references another type
?

tomleb (commented):
Overall looks good to me, 1 question and 1 suggestion

pkg/schema/definitions/handler.go
MbolotSuse: No, from what I have seen these types can be complex (see the count schema for an example). However, our current schema resource doesn't handle this case well (see the count schema and the item count resource), so we don't need to handle it here.

pkg/schema/definitions/handler.go
MbolotSuse: I'm a bit split on it. I don't want to re-implement logic, but I also don't want to make a longer, worse-performing function in order to avoid re-implementing ~5 lines. Let me try implementing a replacement function for this and see how that looks.

MbolotSuse: I've update the function to use this approach. It might have slightly less performance but I think it reads a bit cleaner and avoid us doing a string split here. LMK what you think.

tomleb (approved):

nflynt (approved):
:+1:

PR #206, opened 2024-05-02
Bumping version 0.28.6 to 0.29.3 for k8s dependencies.

ericpromislow:
Related to issue: 45350


go.mod
```diff
@@ -30,13 +32,13 @@ require (
...
 	golang.org/x/sync v0.7.0
 	gopkg.in/yaml.v3 v3.0.1
-	k8s.io/api v0.28.6
-	k8s.io/apiextensions-apiserver v0.28.6
-	k8s.io/apimachinery v0.28.6
-	k8s.io/apiserver v0.28.6
+	k8s.io/api v0.30.0
+	k8s.io/apiextensions-apiserver v0.29.4
+	k8s.io/apimachinery v0.30.0
+	k8s.io/apiserver v0.29.4
```
tomleb, line 38:
```suggestion
	k8s.io/api v0.29.3
	k8s.io/apiextensions-apiserver v0.29.3
	k8s.io/apimachinery v0.29.3
	k8s.io/apiserver v0.29.3
```

ericpromislow:
+1 on getting rid of the `0.30` parts -- `go mod tidy` made that happen, and I should have fixed it. But the latest tag now is `v0.29.4`. Is there a reason we should go with 0.29.3 instead?

MbolotSuse:
@ericpromislow rancher's [client-go fork](https://github.com/rancher/client-go/releases) doesn't produce updates for every k8s version. We try to stick to what our fork uses to make sure we don't introduce odd compatibility issues.

tomleb (commented):
Note that we're going to have to update norman, apiserver and dynamiclistener here to their version that supports k8s 1.29. So it might be best to start with these first.

Here's a graph of dependencies:
image

PR #218, opened 2024-05-24
Migrate from Drone to GHA

nflynt:
This is a lightweight port of the CI logic from Drone into Github Actions, addressing https://github.com/rancher/rancher/issues/45230

Two things to call out here: 

This is using the `golangci-lint` action, which performs linting validation as its own step. Later, the `validate.sh` script does its own call to`golangci-lint run`. For the lightest possible touch I've left this logic alone, in case anyone is relying on the local Makefile for linting during development. For CI purposes this feels redundant; should we perhaps remove the extra call from `validate.sh`?

Secondly, this removes FOSSA scanning, as we are currently stalled on API key provisioning upstream. This will be re-added at a future date, pending https://github.com/rancher/rancher/issues/45495


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,33 @@
...
+    - name : Checkout repository
+      # https://github.com/actions/checkout/releases/tag/v4.1.1
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+    - name: Install Go
+      # https://github.com/actions/setup-go/releases/tag/v5.0.0
+      uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
+      with:
+        go-version-file: 'go.mod'
+    - name: Install mockgen
+      run: go install github.com/golang/mock/mockgen@v1.6.0
```
JonCrowther, line 23: Mockgen is archived now. We should use `go install go.uber.org/mock/mockgen@latest` if possible since it's actively supported

JonCrowther (changes requested):

.github/workflows/ci.yaml
tomleb: Is this the right place to fix this issue? mockgen has been archived a year ago (June 2023) and many other repos are still using it. (eg: rancher/rancher, webhook, lasso, etc)

Given the deadline for GHA I'd rather spend time on doing a simple port from drone to GHA, and we can open an tech-debt issue for mockgen. What do you think?

nflynt: I was planning to open a separate issue for mockgen, let's not block the GHA migration on that. Several other repos need the same treatment

JonCrowther: Ok, as long as there's some sort of tracking related to fixing this I'm fine. Probably not worth addressing in this PR.

JonCrowther (approved):

nflynt:
mockgen dependency moved to this issue for tracking https://github.com/rancher/rancher/issues/45611

PR #220, opened 2024-05-28
[Backport] v2.7: Migrate from Drone to GHA

nflynt:
This is a backport of https://github.com/rancher/steve/pull/218

In the interest of the lightest possible touch, I'm matching the versions of all dependencies used in base v2.7. We can always update these later. As with the upstream PR, the FOSSA task is removed pending API Key provisioning, and will be re-added at a later date.


tomleb (approved):
LGTM

PR #224, opened 2024-06-04
k8s 1.30 dependencies update

vardhaman22:
issue: https://github.com/rancher/rancher/issues/45332
- updated wrangler from v2 to v3
- bumped norman and apiserver 
- bumped dynamiclistener to v0.6.0-rc1
- bumped k8s dependencies to v0.30.1
- bumped golangci lint to v1.59.0 in the workflow.
- removed client-go rancher fork replace statement since all the fixes are included in upstream for client-go https://github.com/rancher/client-go/pull/11#issuecomment-2036238634 [upstreamPR](https://github.com/kubernetes/kubernetes/pull/117046)


go.mod
```diff
@@ -1,12 +1,13 @@
...
-go 1.22
+go 1.22.0
+
+toolchain go1.22.1
 
 replace (
 	github.com/crewjam/saml => github.com/rancher/saml v0.2.0
 	github.com/knative/pkg => github.com/rancher/pkg v0.0.0-20181214184433-b04c0947ad2f
 	github.com/matryer/moq => github.com/rancher/moq v0.0.0-20190404221404-ee5226d43009
-	k8s.io/client-go => github.com/rancher/client-go v1.29.3-rancher1
```
MbolotSuse, line 9: Are you sure that we don't need the changes from the client-go fork in steve?

vardhaman22: was about to update the PR comment...it was discussed in PR comment : https://github.com/rancher/client-go/pull/11#issuecomment-2036238634
so with k8s 0.30 client go all fixes have gone into upstream.

MbolotSuse: Thanks, I'll leave this thread open so that other reviewers can see it easily, but if all the fixes are upstream then that's a good change.

nflynt (approved):

go.mod
kinarashah: https://github.com/rancher/security-team/issues/349#issue-1820307873 has more context on why our custom changes are no longer required.

kinarashah (approved):

PR #230, opened 2024-07-02
Adding APIGroups back to the schema

MbolotSuse:
## Overview 

Related to rancher/rancher#45955. Previous changes changed the order that schemas were evaluated in (used to be `models`, then `ServerGroupsAndResources`, is now `ServerGroupsAndResources`, then `models`). The APIGroups resource, which is not itself a server group/resource, but is a model, was omitted because of this. This re-adds this custom schema first (so that the description later on can fill in more detail), making the endpoint visible again in steve.

## Solution Detail
- Adds a `BaseSchema` to the `apigroup` package.
  - This package defines the template (which is applied to customize the schema later) for the apigroup schema.
- Changes schema generation functionality to add the `BaseSchema` for `apigroup`.
  - This is done first so that `addDescription` will add a description for this schema.

### DIFF
--- a/pkg/resources/apigroups/apigroup.go
+++ b/pkg/resources/apigroups/apigroup.go
@@ -7,9 +7,21 @@ import (
 
 	"github.com/rancher/apiserver/pkg/store/empty"
 	"github.com/rancher/apiserver/pkg/types"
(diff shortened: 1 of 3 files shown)


ericpromislow (approved):
Looks good, tests pass

tomleb (approved):

PR #245, opened 2024-07-18
refactor(accesscontrol): add unit tests for CacheKey

aruiz14:
This is a follow up to https://github.com/rancher/steve/pull/244 with a small refactor to make the function easier to test, and adding unit tests for `CacheKey`.

### DIFF
--- a/pkg/accesscontrol/access_store.go
+++ b/pkg/accesscontrol/access_store.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/hex"
+	"hash"
 	"sort"
 	"time"
 
 	v1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/rbac/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/util/cache"
 	"k8s.io/apiserver/pkg/authentication/user"
 )
@@ -19,9 +21,20 @@ type AccessSetLookup interface {
 	PurgeUserData(id string)
 }
 
+type policyRules interface {
+	get(string) *AccessSet
+	getRoleBindings(string) []*rbacv1.RoleBinding
+	getClusterRoleBindings(string) []*rbacv1.ClusterRoleBinding
+}
+
+type roleRevisions interface {
+	roleRevision(string, string) string
+}
+
 type AccessStore struct {
-	users  *policyRuleIndex
-	groups *policyRuleIndex
+	users  policyRules
+	groups policyRules
+	roles  roleRevisions
 	cache  *cache.LRUExpireCache
 }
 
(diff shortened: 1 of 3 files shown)


pkg/accesscontrol/policy_rule_index.go
```diff
@@ -20,13 +19,12 @@ type policyRuleIndex struct {
 	rCache              v1.RoleCache
 	crbCache            v1.ClusterRoleBindingCache
 	rbCache             v1.RoleBindingCache
-	revisions           *roleRevisionIndex
```
aruiz14, line 23: I moved this to AccessStore, since it was only used by `addRolesToHash` which also makes more sense there. `policyRuleIndex` does not need to know about any hashing.

bigkevmcd (dismissed):

pkg/accesscontrol/access_store.go
```diff
@@ -73,16 +86,34 @@ func (l *AccessStore) PurgeUserData(id string) {
...
+
+	l.addRolesToHash(d, user.GetName(), l.usersPolicyRules)
+	for _, group := range groups {
+		l.addRolesToHash(d, group, l.groupsPolicyRules)
 	}
 
 	return hex.EncodeToString(d.Sum(nil))
 }
+
+var null = []byte{'\x00'}
```
bigkevmcd, line 102: This could be simplified to `[]byte{0}`

bigkevmcd (changes requested):
We could do with tests for `AccessFor` and `PurgeUserData` ?

Also, `NewAccessStore` doesn't get tested, I'm guessing because currently there's no way to modify `usersPolicyRules` and `groupsPolicyRules` ?

pkg/accesscontrol/access_store.go
aruiz14: Good point, thanks! In fact, writing this null does not seem to be even needed, so I'll just remove it.

aruiz14:
> We could do with tests for `AccessFor` and `PurgeUserData` ?
Done, thanks for pointing it out.

> Also, `NewAccessStore` doesn't get tested, I'm guessing because currently there's no way to modify `usersPolicyRules` and `groupsPolicyRules` ?
That would be hard to do without changing the public API or mocking `rbac.Interface` 😔 
I will consider it for a follow up PR.

bigkevmcd (dismissed):

pkg/accesscontrol/access_store.go
```diff
@@ -31,10 +44,10 @@ type roleKey struct {
 }
 
 func NewAccessStore(ctx context.Context, cacheResults bool, rbac v1.Interface) *AccessStore {
```
ericpromislow, line 46: I know this part has already been removed, but it looks like `cacheResults` is always true in non-test calls to `NewAccessStore` , and there are no external non-test invocations of `AccessStore{...}`.   So why don't we assume caching is always on?

aruiz14: I made myself the same question 😄
However, my intention was to keep the signature of exported methods unchanged.
Please consider opening a new PR with that proposal.

ericpromislow (approved):
lgtm

bigkevmcd (approved):
Looks good

aruiz14:
@ericpromislow I lack permissions to merge this, please feel free to do it when you have the chance

PR #249, opened 2024-07-29
Handle schema-less CRDs

ericpromislow:
Related to [#46213](https://github.com/rancher/rancher/issues/46213)

### DIFF
--- a/pkg/schema/definitions/fixtures_test.go
+++ b/pkg/schema/definitions/fixtures_test.go
@@ -4,8 +4,10 @@ import (
 	"bytes"
 	"fmt"
 
+	openapi_v2 "github.com/google/gnostic-models/openapiv2"
 	"github.com/rancher/wrangler/v3/pkg/yaml"
 	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/kube-openapi/pkg/util/proto"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


pkg/schema/definitions/handler.go
```diff
@@ -223,7 +223,11 @@ func listGVKModels(models proto.Models, groups *metav1.APIGroupList, crdCache wa
 				Version: version.Name,
 				Kind:    crd.Spec.Names.Kind,
 			}
-			gvkToCRD[gvk] = version.Schema.OpenAPIV3Schema
+			if version.Schema != nil {
+				gvkToCRD[gvk] = version.Schema.OpenAPIV3Schema
+			} else {
+				gvkToCRD[gvk] = nil
```
brandond, line 229: nit: this is a `map[schema.GroupVersionKind]*apiextv1.JSONSchemaProps`.  We don't iterate over its keys, and as far as I can tell it is only used below as `CRD:       gvkToCRD[*gvk]`. Since we're not iterating over the map, explicitly adding an entry with a nil pointer doesn't seem to make sense - the default value for a pointer is already nil.
```suggestion
```

brandond (changes requested):

pkg/schema/definitions/fixtures_test.go
```diff
@@ -398,3 +400,114 @@ definitions:
...
+        properties:
+          spec:
+            type: object
+            properties:
+              rkeConfig:
+                type: object
+                nullable: true
+        x-kubernetes-preserve-unknown-fields: true
+    served: true
+    storage: true
```
MbolotSuse, line 434: From the issue, I don't think that this was the use case that was causing our problems. I think we need something with an empty openapiV3 schema. So like the below:
```suggestion
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: schemaless.management.cattle.io
spec:
  conversion:
    strategy: None
  group: management.cattle.io
  names:
    kind: Schemaless
    listKind: SchemalessList
    plural: schemalese
    singular: schemaless
  scope: Cluster
  preserveUnkownFields: true
  versions:
  - name: v2
    served: true
    storage: true
```
Would you mind updating this and the models to be a bit more like what we could see in production?

MbolotSuse (changes requested):
One comment on the tests, in addition to what @brandond brought up.

ericpromislow:
Done in #253

PR #253, opened 2024-08-05
Handle Schema-less CRDs

MbolotSuse:
Continuation of #249. Related to rancher/rancher#46213.

## Overview

Some CRDs created on older versions of Kubernetes may not have any openAPI schema created for them. Steve would not handle the nil schema in these cases, produving a panic.

### DIFF
--- a/pkg/schema/definitions/fixtures_test.go
+++ b/pkg/schema/definitions/fixtures_test.go
@@ -4,10 +4,8 @@ import (
 	"bytes"
 	"fmt"
 
-	openapi_v2 "github.com/google/gnostic-models/openapiv2"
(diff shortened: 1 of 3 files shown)


tomleb (approved):
LGTM.

Might be useful to add a comment in the fixtures to specify _why_ we're testing a schemaless CRD even though one cannot create such CRDs (other than by installing a very very old k8s version and using a v1beta1 CRD).

nflynt (approved):
:+1:

hansbogert:
it says **related** to https://github.com/rancher/rancher/issues/46213, does it actually **fix** it?

PR #261, opened 2024-08-13
Pass the debug status to the dynamiclistener server

ericpromislow:
This PR works with https://github.com/rancher/dynamiclistener/pull/118 -- the logger is global, so if steve is setting it to debug-mode, the dynamiclistener server needs to know what log-level the logger is at.


ericpromislow:
No need for this PR -- the dynamiclistener can look at the current logger to figure out whether we're debugging or not.

PR #263, opened 2024-08-14
sql: pass a discovery to lasso to do watchability checks

moio:
Fixes https://github.com/rancher/rancher/issues/46030

Requires https://github.com/rancher/lasso/pull/93 to be merged (and lasso to be bumped here) before merging.

### DIFF
--- a/pkg/stores/sqlproxy/proxy_store.go
+++ b/pkg/stores/sqlproxy/proxy_store.go
@@ -39,6 +39,7 @@ import (
 	apitypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -126,6 +127,7 @@ type Store struct {
 	namespaceCache Cache
 	lock           sync.Mutex
 	columnSetter   SchemaColumnSetter
+	discovery      discovery.DiscoveryInterface
 }
 
 type CacheFactoryInitializer func() (CacheFactory, error)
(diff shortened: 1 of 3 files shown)


pkg/stores/sqlproxy/proxy_store.go
MbolotSuse, line 1: I think we need to handle cases (like on L631) where the factor can't get a cache more gracefully now. My thinking is that if someone tries to list for this type and it's not watchable, either we need to return an error indicating that this endpoint can't be used (probably a [501](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501) error) or we should try to do an un-cached list and go from there. Either way I think we probably need a more detailed error handling mechanism there.

MbolotSuse (changes requested):
One comment on error handling.

pkg/stores/sqlproxy/proxy_store.go
moio: With this patch (and the accompanying steve one) we return error 500

I am in conversations with UI folks to see how this can be worked around in the few cases where we really need non-watchable but listable resources

MbolotSuse: > we return error 500

I think that we should be a bit more specific than this - [500](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) is a bit too generic for this condition, which we know isn't resolvable. 

> I am in conversations with UI folks to see how this can be worked around in the few cases where we really need non-watchable but listable resources

Sounds good, I'll leave this thread open until we have an idea of how our API contract should be formed. That being said, even in this original PR I think that we want to be more specific on the error code.

moio: UI is OK with error 500.

In any case, this is expected to be temporary as we agreed to support listing unwatchable resources before the feature is turned on by default (https://github.com/rancher/rancher/issues/46768), so the error code will not matter for long - and at the moment error 500 already produces the right UI behavior (table reports "n/a" instead of the expected values).

moio:
See https://github.com/rancher/lasso/pull/93#issuecomment-2312401589

PR #270, opened 2024-08-30
Adding Group, Version, Kind in Update proxy_store.go

gehrkefc:
This is related to rancher/rancher issue: [42823](https://github.com/rancher/rancher/issues/42823)

### DIFF
--- a/pkg/stores/partition/store.go
+++ b/pkg/stores/partition/store.go
@@ -389,7 +389,7 @@ func ToAPIEvent(apiOp *types.APIRequest, schema *types.APISchema, event watch.Ev
 
 	if event.Type == watch.Error {
 		status, _ := event.Object.(*metav1.Status)
-		apiEvent.Error = fmt.Errorf(status.Message)
+		apiEvent.Error = fmt.Errorf("%s", status.Message)
 		return apiEvent
 	}
 
--- a/pkg/stores/partition/store.go
+++ b/pkg/stores/partition/store.go
@@ -4,6 +4,7 @@ package partition
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"reflect"
@@ -389,7 +390,7 @@ func ToAPIEvent(apiOp *types.APIRequest, schema *types.APISchema, event watch.Ev
 
 	if event.Type == watch.Error {
 		status, _ := event.Object.(*metav1.Status)
-		apiEvent.Error = fmt.Errorf("%s", status.Message)
+		apiEvent.Error = errors.New(status.Message)
 		return apiEvent
 	}
 
--- a/pkg/stores/proxy/proxy_store.go
+++ b/pkg/stores/proxy/proxy_store.go
@@ -432,7 +432,15 @@ func (s *Store) Create(apiOp *types.APIRequest, schema *types.APISchema, params 
 	}
 
 	gvk := attributes.GVK(schema)
-	input["apiVersion"], input["kind"] = gvk.ToAPIVersionAndKind()
+	apiVersion, kind := gvk.ToAPIVersionAndKind()
+
+	if value, found := input["apiVersion"]; !found || value == "" {
+		input["apiVersion"] = apiVersion
+	}
+
+	if value, found := input["kind"]; !found || value == "" {
+		input["kind"] = kind
+	}
 
 	buffer := WarningBuffer{}
 	k8sClient, err := metricsStore.Wrap(s.clientGetter.TableClient(apiOp, schema, ns, &buffer))
@@ -506,7 +514,15 @@ func (s *Store) Update(apiOp *types.APIRequest, schema *types.APISchema, params 
 	}
 
 	gvk := attributes.GVK(schema)
-	input["apiVersion"], input["kind"] = gvk.ToAPIVersionAndKind()
+	apiVersion, kind := gvk.ToAPIVersionAndKind()
+
+	if value, found := input["apiVersion"]; !found || value == "" {
+		input["apiVersion"] = apiVersion
+	}
+
+	if value, found := input["kind"]; !found || value == "" {
+		input["kind"] = kind
+	}
 
 	opts := metav1.UpdateOptions{}
 	if err := decodeParams(apiOp, &opts); err != nil {
--- a/pkg/stores/proxy/proxy_store.go
+++ b/pkg/stores/proxy/proxy_store.go
@@ -516,7 +516,7 @@ func (s *Store) Update(apiOp *types.APIRequest, schema *types.APISchema, params 
 
 	resourceVersion := input.String("metadata", "resourceVersion")
 	if resourceVersion == "" {
-		return nil, nil, fmt.Errorf(errResourceVersionRequired)
+		return nil, nil, errors.New(errResourceVersionRequired)
 	}
 
 	gvk := attributes.GVK(schema)
--- a/pkg/stores/proxy/proxy_store.go
+++ b/pkg/stores/proxy/proxy_store.go
@@ -446,15 +446,7 @@ func (s *Store) Create(apiOp *types.APIRequest, schema *types.APISchema, params 
 	}
 
 	gvk := attributes.GVK(schema)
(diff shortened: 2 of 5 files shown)


pkg/stores/proxy/proxy_store.go
```diff
@@ -505,6 +505,9 @@ func (s *Store) Update(apiOp *types.APIRequest, schema *types.APISchema, params
 		return nil, nil, fmt.Errorf("metadata.resourceVersion is required for update")
 	}
 
+	gvk := attributes.GVK(schema)
+	input["apiVersion"], input["kind"] = gvk.ToAPIVersionAndKind()
```
ericpromislow, line 509: What if both or one of these fields are specified in the body?  Maybe it should be something like:
```
   apiVersion, kind := gvk.ToAPIVersionAndKind()
   v, ok := input["apiVersion"]
   if !ok || v == "" { input["apiVersion"] = apiVersion }
   v, ok = input["kind"]
   if !ok || v == "" { input["kind"] = kind }
```

ericpromislow (changes requested):
Probably don't overwrite supplied values?

gehrkefc:
> Probably don't overwrite supplied values?

Added a check based in our conversation. Should we consider a situation where the user provides only one of the values and we complete only the other one ? Can this generate any kind of inconsistency ?

ericpromislow (dismissed):
Looks ok to me. I don't see an inconsistency if only one value is specified -- defaults and specified values work the same way elsewhere, like command-line options

pkg/stores/proxy/proxy_store.go
```diff
@@ -445,7 +446,15 @@ func (s *Store) Create(apiOp *types.APIRequest, schema *types.APISchema, params
 	}
 
 	gvk := attributes.GVK(schema)
-	input["apiVersion"], input["kind"] = gvk.ToAPIVersionAndKind()
+	apiVersion, kind := gvk.ToAPIVersionAndKind()
```
tomleb, line 449: This changes the behavior where before the change only the GVK was considered, and after the change the input is considered first.

Imo this _could_ be breaking the UI if they were sending the wrong apiVersion / kind before but never noticed it. That seems unlikely though.

Ideally we would have the same behavior between Update and Create. @MbolotSuse what's your thought on this?

MbolotSuse: I agree that create and update should have the same behavior. IIUC that is part of this issue, so making them behave the same way would be in scope for this issue.

As far as the "breaking" part goes, I'm not sure about that part. I don't think that this is something that the UI should be relying on, but I also don't want to potentially break the UI if we don't need to. In this case, I think the best thing to do is address the core of the issue - update should do what create does, and that means overriding for now.

gehrkefc: Ok, I'll change both of the methods to override the data from input. Thank you!

pkg/stores/proxy/proxy_store_test.go
```diff
@@ -488,3 +489,242 @@ func TestCreate(t *testing.T) {
...
+		value   *unstructured.Unstructured
+		warning []types.Warning
+		err     error
+	}
+
+	sampleCreateInput := input{
+		apiOp: &types.APIRequest{
+			Request: &http.Request{
+				URL:    &url.URL{},
+				Method: http.MethodPut,
```
tomleb, line 511: Does this need to be `http.MethodPost`?

pkg/stores/proxy/proxy_store_test.go
```diff
@@ -488,3 +489,242 @@ func TestCreate(t *testing.T) {
...
+			},
+			Schema: &types.APISchema{
+				Schema: &schemas.Schema{
+					ID: "testing",
+				},
+			},
+			Method: http.MethodPut,
```
tomleb, line 518: Does this need to be `http.MethodPost`?

pkg/stores/sqlproxy/proxy_store_test.go
```diff
@@ -1104,3 +1104,242 @@ func TestCreate(t *testing.T) {
...
+		value   *unstructured.Unstructured
+		warning []types.Warning
+		err     error
+	}
+
+	sampleCreateInput := input{
+		apiOp: &types.APIRequest{
+			Request: &http.Request{
+				URL:    &url.URL{},
+				Method: http.MethodPut,
```
tomleb, line 1126: Does this need to be `http.MethodPost`?

pkg/stores/sqlproxy/proxy_store_test.go
```diff
@@ -1104,3 +1104,242 @@ func TestCreate(t *testing.T) {
...
+			},
+			Schema: &types.APISchema{
+				Schema: &schemas.Schema{
+					ID: "testing",
+				},
+			},
+			Method: http.MethodPut,
```
tomleb, line 1133: Does this need to be `http.MethodPost`?

pkg/stores/proxy/proxy_store_test.go
```diff
@@ -488,3 +489,242 @@ func TestCreate(t *testing.T) {
...
+
+	testCases := []struct {
+		name               string
+		updateCallbackFunc clientgotesting.ReactionFunc
+		createInput        *input
+		updateInput        input
+		expected           expected
+	}{
+		{
+			name: "update - usual request",
```
tomleb, line 551: Would be great to have a test case for the case where apiVersion and kind is missing to cover the change in this PR.

pkg/stores/sqlproxy/proxy_store_test.go
```diff
@@ -1104,3 +1104,242 @@ func TestCreate(t *testing.T) {
...
+
+	testCases := []struct {
+		name               string
+		updateCallbackFunc clientgotesting.ReactionFunc
+		createInput        *input
+		updateInput        input
+		expected           expected
+	}{
+		{
+			name: "update - usual request",
```
tomleb, line 1166: Would be great to have a test case for the case where apiVersion and kind is missing to cover the change in this PR.

gehrkefc: You're right. Just added the tests.

tomleb (dismissed):

pkg/stores/partition/store.go
```diff
@@ -389,7 +389,7 @@ func ToAPIEvent(apiOp *types.APIRequest, schema *types.APISchema, event watch.Ev
 
 	if event.Type == watch.Error {
 		status, _ := event.Object.(*metav1.Status)
-		apiEvent.Error = fmt.Errorf(status.Message)
+		apiEvent.Error = fmt.Errorf("%s", status.Message)
```
ericpromislow, line 392: Sorry, the `fmt.Errorf("%s", status.Message)` should be `errors.New(status.Message)` because no formatting is necessary.

pkg/stores/proxy/proxy_store.go
```diff
@@ -515,9 +516,12 @@ func (s *Store) Update(apiOp *types.APIRequest, schema *types.APISchema, params
 
 	resourceVersion := input.String("metadata", "resourceVersion")
 	if resourceVersion == "" {
-		return nil, nil, fmt.Errorf("metadata.resourceVersion is required for update")
+		return nil, nil, fmt.Errorf(errResourceVersionRequired)
```
ericpromislow, line 519: `fmt.Errorf` should be `errors.New`

ericpromislow (changes requested):
I see we have lots of `fmt.Errorf(STRING)` but we shouldn't add more.

Everything else looks fine.

pkg/stores/partition/store.go
gehrkefc: Ok, just fixed it. Thanks!

pkg/stores/proxy/proxy_store.go
gehrkefc: Just fixed it. Thanks!

ericpromislow (approved):
Looks good

tomleb (approved):

PR #271, opened 2024-08-30
Index more sqlite cache fields

ericpromislow:
Related to [#46642](https://github.com/rancher/rancher/issues/46642)

I don't see a straightforward way to test this

### DIFF
--- a/pkg/resources/virtual/common/common.go
+++ b/pkg/resources/virtual/common/common.go
@@ -23,11 +23,6 @@ type DefaultFields struct {
 	Cache SummaryCache
 }
 
-// GetTransform produces the default transformation func
-func (d *DefaultFields) GetTransform() func(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
-	return d.TransformCommon
-}
-
 func (d *DefaultFields) GetUnstructuredObjectWrapper(obj interface{}) (interface{}, bool, error) {
 	raw, isSignal, err := GetUnstructured(obj)
 	if isSignal {
--- a/pkg/resources/virtual/common/common.go
+++ b/pkg/resources/virtual/common/common.go
@@ -23,17 +23,6 @@ type DefaultFields struct {
 	Cache SummaryCache
 }
 
-func (d *DefaultFields) GetUnstructuredObjectWrapper(obj interface{}) (interface{}, bool, error) {
-	raw, isSignal, err := GetUnstructured(obj)
-	if isSignal {
-		return obj, true, nil
-	}
-	if err != nil {
-		return nil, false, err
-	}
-	return raw, false, nil
-}
-
 // TransformCommon implements virtual.VirtualTransformFunc, and adds reserved fields/summary
 func (d *DefaultFields) TransformCommon(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
 	obj = addIDField(obj)
--- a/pkg/resources/virtual/common/common.go
+++ b/pkg/resources/virtual/common/common.go
@@ -10,7 +10,6 @@ import (
 	wranglerSummary "github.com/rancher/wrangler/v3/pkg/summary"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/tools/cache"
 )
 
 // SummaryCache provides an interface to get a summary/relationships for an object. Implemented by the summaryCache
@@ -25,25 +24,29 @@ type DefaultFields struct {
 }
 
 // GetTransform produces the default transformation func
-func (d *DefaultFields) GetTransform() cache.TransformFunc {
-	return d.transform
+func (d *DefaultFields) GetTransform() func(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	return d.TransformCommon
 }
 
-// transform implements virtual.VirtualTransformFunc, and adds reserved fields/summary
-func (d *DefaultFields) transform(obj any) (any, error) {
-	raw, isSignal, err := getUnstructured(obj)
+func (d *DefaultFields) GetUnstructuredObjectWrapper(obj interface{}) (interface{}, bool, error) {
+	raw, isSignal, err := GetUnstructured(obj)
 	if isSignal {
-		return obj, nil
+		return obj, true, nil
 	}
 	if err != nil {
-		return nil, err
+		return nil, false, err
 	}
-	raw = addIDField(raw)
-	raw, err = addSummaryFields(raw, d.Cache)
+	return raw, false, nil
+}
+
+// TransformCommon implements virtual.VirtualTransformFunc, and adds reserved fields/summary
+func (d *DefaultFields) TransformCommon(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	obj = addIDField(obj)
+	obj, err := addSummaryFields(obj, d.Cache)
 	if err != nil {
 		return nil, fmt.Errorf("unable to add summary fields: %w", err)
 	}
-	return raw, nil
+	return obj, nil
 }
 
 // addSummaryFields adds the virtual fields for object state.
--- a/pkg/resources/virtual/common/common_test.go
+++ b/pkg/resources/virtual/common/common_test.go
@@ -175,7 +175,7 @@ func TestTransformCommonObjects(t *testing.T) {
 				require.Equal(t, test.input, test.wantOutput)
 				return
 			}
-			output, err := df.GetTransform()(raw)
+			output, err := df.TransformCommon(raw)
 			if test.wantError {
 				require.Error(t, err)
 			} else {
--- a/pkg/resources/virtual/common/testutil.go
+++ b/pkg/resources/virtual/common/testutil.go
@@ -1,0 +1,16 @@
+package common
+
+import (
+	"github.com/rancher/steve/pkg/summarycache"
+	"github.com/rancher/wrangler/v3/pkg/summary"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type FakeSummaryCache struct {
+	SummarizedObject *summary.SummarizedObject
+	Relationships    []summarycache.Relationship
+}
+
+func (f *FakeSummaryCache) SummaryAndRelationship(runtime.Object) (*summary.SummarizedObject, []summarycache.Relationship) {
+	return f.SummarizedObject, f.Relationships
+}
--- a/pkg/resources/virtual/events/events_test.go
+++ b/pkg/resources/virtual/events/events_test.go
@@ -1,0 +1,93 @@
+package events_test
+
+import (
+	"testing"
+
+	"github.com/rancher/steve/pkg/resources/virtual/events"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func TestTransformEvents(t *testing.T) {
+	tests := []struct {
+		name       string
+		input      any
+		wantOutput any
+		wantError  bool
+	}{
+		{
+			name: "fix event fields",
+			input: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"apiVersion": "/v1",
+					"kind":       "Event",
+					"metadata": map[string]interface{}{
+						"name":      "gregsFarm",
+						"namespace": "gregsNamespace",
+					},
+					"id":   "eventTest1id",
+					"type": "Gorniplatz",
+				},
+			},
+			wantOutput: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"apiVersion": "/v1",
+					"kind":       "Event",
+					"metadata": map[string]interface{}{
+						"name":      "gregsFarm",
+						"namespace": "gregsNamespace",
+					},
+					"id":    "eventTest1id",
+					"type":  "Gorniplatz",
+					"_type": "Gorniplatz",
+				},
+			},
+		},
+		{
+			name: "don't fix non-default-group event fields",
+			input: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"apiVersion": "palau.io/v1",
+					"kind":       "Event",
+					"metadata": map[string]interface{}{
+						"name":      "gregsFarm",
+						"namespace": "gregsNamespace",
+					},
+					"id":   "eventTest1id",
+					"type": "Gorniplatz",
+				},
+			},
+			wantOutput: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"apiVersion": "palau.io/v1",
+					"kind":       "Event",
+					"metadata": map[string]interface{}{
+						"name":      "gregsFarm",
+						"namespace": "gregsNamespace",
+					},
+					"id":   "eventTest1id",
+					"type": "Gorniplatz",
+				},
+			},
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			var output interface{}
+			var err error
+			raw, ok := test.input.(*unstructured.Unstructured)
+			if ok && raw.GetKind() == "Event" && raw.GetAPIVersion() == "/v1" {
+				output, err = events.TransformEventObject(raw)
+			} else {
+				output = raw
+				err = nil
+			}
+			require.Equal(t, test.wantOutput, output)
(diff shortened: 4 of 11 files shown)


pkg/stores/sqlproxy/proxy_store.go
```diff
@@ -58,7 +58,16 @@ var (
 		"_v1_Pod":       {{`spec`, `containers`, `image`}, {`spec`, `nodeName`}},
 		"_v1_ConfigMap": {{`metadata`, `labels[harvesterhci.io/cloud-init-template]`}},
 
-		"management.cattle.io_v3_Node": {{`status`, `nodeName`}},
+		"cluster.x-k8s.io_v1beta1_Machine":     {{`spec`, `clusterName`}},
+		"management.cattle.io_v3_Node":         {{`status`, `nodeName`}},
+		"management.cattle.io_v3_NodePool":     {{`spec`, `clusterName`}},
+		"management.cattle.io_v3_NodeTemplate": {{`spec`, `clusterName`}},
+		"provisioning.cattle.io_Cluster": {
```
MbolotSuse, line 65: AFAIK this needs a group as well
```suggestion
		"provisioning.cattle.io_v1_Cluster": {
```

MbolotSuse (changes requested):

pkg/stores/sqlproxy/proxy_store.go
```diff
@@ -53,12 +53,28 @@ var (
...
+			{"status", "nodeName"}},
+		gvkKey("management.cattle.io", "v3", "NodePool"): {
+			{"spec", "clusterName"}},
+		gvkKey("management.cattle.io", "v3", "NodeTemplate"): {
+			{"spec", "clusterName"}},
+		gvkKey("provisioning.cattle.io", "v1", "Cluster"): {
+			{"metadata", "labels[provider.cattle.io]"},
+			{"status", "provider"},
+			{"status", "allocatable", "cpu"},
+			{"status", "allocatable", "memory"}},
```
MbolotSuse, line 77: I think this is missing "status.allocatable.pods" which was also asked for on the provisioning cluster.

In addition, there's quite a few other types/fields which were asked for in comments to the original issue. This includes:
- https://github.com/rancher/rancher/issues/46642#issuecomment-2298884495 (event fields)
- https://github.com/rancher/rancher/issues/46642#issuecomment-2302085672 (clusterrepo/operation fields).

MbolotSuse (changes requested):

ericpromislow:
I figured out the other fields, except I'm not sure if the events should go in the empty-string group or in group `"events.k8s.io"`

pkg/stores/sqlproxy/proxy_store.go
```diff
@@ -53,12 +53,59 @@ var (
...
+		},
+		gvkKey("catalog.cattle.io", "v1", "Operation"): {
+			{"status", "action"},
+			{"status", "namespace"},
+			{"status", "releaseName"},
+		},
+		gvkKey("cluster.x-k8s.io", "v1beta1", "Machine"): {
+			{"spec", "clusterName"}},
+		gvkKey("events.k8s.io", "v1", "Event"): {
+			{"spec", "clusterName"}},
```
MbolotSuse, line 91: I think this is a duplicate of L62.

pkg/stores/sqlproxy/proxy_store.go
```diff
@@ -53,12 +53,59 @@ var (
...
 	paramCodec                = runtime.NewParameterCodec(paramScheme)
 	typeSpecificIndexedFields = map[string][][]string{
-		"_v1_Namespace": {{`metadata`, `labels[field.cattle.io/projectId]`}},
-		"_v1_Node":      {{`status`, `nodeInfo`, `kubeletVersion`}, {`status`, `nodeInfo`, `operatingSystem`}},
-		"_v1_Pod":       {{`spec`, `containers`, `image`}, {`spec`, `nodeName`}},
-		"_v1_ConfigMap": {{`metadata`, `labels[harvesterhci.io/cloud-init-template]`}},
-
-		"management.cattle.io_v3_Node": {{`status`, `nodeName`}},
+		gvkKey("", "v1", "Event"): {
+			{"_type"},
```
MbolotSuse, line 57: The syntax of this type make me think that this is a reserved field. Basically, Steve has a few reserved fields which it adds to an object (id is another). When an object already has this field, Steve moves the original field to "_fieldName".

Vai doesn't store the modified versions of these objects - instead, it stores the original. You'll need to use the new virtual fields logic (take a look [here](https://github.com/rancher/steve/blob/main/pkg/resources/virtual/virtual.go)) to start to make sure that this is filterable/sortable.

You can also try testing this to see if I am wrong.

MbolotSuse (changes requested):

pkg/stores/sqlproxy/proxy_store.go
ericpromislow: Yes.

pkg/stores/sqlproxy/proxy_store.go
ericpromislow: This converts to key `"_v1_Event"`, I don't see that looks like a reserved name.

Also this block is a duplicate of `gvkKey("events.k8s.io", "v1", "Event")` and it's possible that we don't need this block.

MbolotSuse: Sorry for not clarifying - it's not the resource name that is the problem, it's the `_type` field.

ericpromislow: Testing shows that for `/v1/events`, specifying `?filter=type=STRING` converts `type` into `_type`:
```
$ curl -skL ... /v1/events' | jq . | egrep '"_?type"'
  "type": "collection",
      "type": "event",
      "_type": "Normal",
      "type": "event",
      "_type": "Warning",
…
# Searching for 'type=event' gives the empty-list returned JSON payload:
$ curl -skL ... /v1/events'?filter=type=event' | wc
       1       1     197
# Searching for `?filter=type=....` converts `type` to `_type`
$ curl -skL ... /v1/events'?filter=type=Normal' | wc
       1       2981  602473

Comment (ericpromislow) [reply pkg/stores/sqlproxy/proxy_store.go:62]: As for `events`, when I look at the dashboard PR it looks like they're matching on `_v1_Event` and not `events.k8s.io_v1_Event`

Comment (MbolotSuse) [reply pkg/stores/sqlproxy/proxy_store.go:62]: The filter needs to apply on _type - since that's the field that they see they need to be able to filter on the final field rather than on the original field before steve's transformation. Please give this another look.

File: pkg/resources/virtual/common/common.go [hunk 5]
```diff
@@ -29,6 +29,11 @@ func (d *DefaultFields) GetTransform() cache.TransformFunc {
 	return d.transform
 }
 
+// GetEventTransform produces the default transformation func
+func (d *DefaultFields) GetEventTransform() cache.TransformFunc {
```
MbolotSuse, line 33: I don't think that this fits the purpose of "DefaultFields". Default fields are the fields that are the same for all objects, but this function is new code specific to a single type. I think this needs to go in it's own package, along with the logic to change the type field.

pkg/resources/virtual/virtual.go
```diff
@@ -23,6 +23,9 @@ func NewTransformBuilder(cache common.SummaryCache) *TransformBuilder {
 }
 
 // GetTransformFunc retrieves a TransformFunc for a given GVK. Currently only returns a transformFunc for defaultFields
-func (t *TransformBuilder) GetTransformFunc(_ schema.GroupVersionKind) cache.TransformFunc {
+func (t *TransformBuilder) GetTransformFunc(gvk schema.GroupVersionKind) cache.TransformFunc {
+	if gvk.Kind == "Event" {
```
MbolotSuse, line 27: I would recommend checking full GVK here - events can also be implemented by other API groups and this would add transform functions for those types as well.

MbolotSuse (changes requested):

pkg/resources/virtual/common/common.go
ericpromislow: I've worked out something that still uses the `common` package, pulling the event-specific transformer into another file. It looks a bit hokey but it can be used to scale up to cover more types in the future.

pkg/resources/virtual/virtual.go
ericpromislow: Done, with an extra test.

pkg/resources/virtual/common/common_test.go
MbolotSuse, line 1: I don't think that these tests should go in this file. The tests for this relate to another transform function, namely the one that we define as part of `event_fields.go`, so I think the tests for this should go in a corresponding file.

pkg/resources/virtual/common/event_fields.go
MbolotSuse, line 1: Ideally, this should go in a separate package. Common is for things that are common to all transform functions. Since this is specific to a single type, it shouldn't be in the common package.

pkg/resources/virtual/common/common.go
```diff
@@ -38,8 +38,12 @@ func (d *DefaultFields) transform(obj any) (any, error) {
 	if err != nil {
 		return nil, err
 	}
+	return d.transformCommon(raw)
+}
+
+func (d *DefaultFields) transformCommon(raw *unstructured.Unstructured) (any, error) {
```
MbolotSuse, line 44: If we intend this function to be re-used by other transform functions, then it should be exported (rather then re-used by other functions in the same package).

I would recommend looking more into the interface here - since an object could go through multiple transform functions, I do like that we are avoiding the marshal to unstructured and the signal error checking, but I don't like that we are doing so through embedded objects and use of private methods.

MbolotSuse (changes requested):

pkg/stores/sqlproxy/proxy_store.go
```diff
@@ -247,7 +282,11 @@ func getFieldForGVK(gvk schema.GroupVersionKind) [][]string {
 }
 
 func keyFromGVK(gvk schema.GroupVersionKind) string {
```
moio, line 284: Nitpick: as this function is only called once it could be inlined.

ericpromislow: I missed this first time through. Done now.

pkg/resources/virtual/virtual.go
```diff
@@ -16,13 +20,37 @@ type TransformBuilder struct {
...
+// GetTransformFunc returns the func to transform a raw object into a fixed object, if needed
+func (t *TransformBuilder) GetTransformFunc(gvk schema.GroupVersionKind) cache.TransformFunc {
+	converters := make([]func(*unstructured.Unstructured) (*unstructured.Unstructured, error), 0)
+	if gvk.Kind == "Event" && gvk.Group == "" && gvk.Version == "v1" {
+		converters = append(converters, events.TransformEventObject)
+	}
+	converters = append(converters, t.defaultFields.TransformCommon)
+
+	return func(any interface{}) (interface{}, error) {
+		var raw interface{}
```
moio, line 38:
```suggestion
```

pkg/resources/virtual/virtual.go
```diff
@@ -16,13 +20,37 @@ type TransformBuilder struct {
...
+		raw, isSignal, err := t.defaultFields.GetUnstructuredObjectWrapper(any)
+		if isSignal {
+			return raw, err
+		}
```
moio, line 42:
```suggestion
		}
		if err != nil {
			return nil, err
		}
```

pkg/resources/virtual/common/common.go
```diff
@@ -25,25 +24,29 @@ type DefaultFields struct {
 }
 
 // GetTransform produces the default transformation func
-func (d *DefaultFields) GetTransform() cache.TransformFunc {
-	return d.transform
+func (d *DefaultFields) GetTransform() func(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	return d.TransformCommon
 }
```
moio, line 29: Nitpick: now that `TransformCommon` is public, there is effectively no difference between `GetTransform` and `TransformCommon`. I suggest removing one of the two.

pkg/resources/virtual/common/common.go
```diff
@@ -25,25 +24,29 @@ type DefaultFields struct {
...
 
-// transform implements virtual.VirtualTransformFunc, and adds reserved fields/summary
-func (d *DefaultFields) transform(obj any) (any, error) {
-	raw, isSignal, err := getUnstructured(obj)
+func (d *DefaultFields) GetUnstructuredObjectWrapper(obj interface{}) (interface{}, bool, error) {
```
moio, line 31: Nitpick: this is only ever used inside of GetTransformFunc. I suggest inlining (using `GetUnstructured` directly) there.

moio (commented):
Looks solid to me, the only real small problem being a missed error handling `if`.

Other than that it's totally optional nitpicks.

Thanks for all the work! Looking forward to see this in.

pkg/resources/virtual/virtual.go
ericpromislow: Right

pkg/resources/virtual/common/common.go
ericpromislow: Yes, no need for 'GetTransform'

pkg/resources/virtual/common/common.go
ericpromislow: Yup, true

pkg/resources/virtual/common/common.go
moio: lgtm

moio (approved):
LGTM

pkg/resources/virtual/virtual_test.go
```diff
@@ -0,0 +1,202 @@
...
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			fakeCache := common.FakeSummaryCache{
+				SummarizedObject: test.hasSummary,
+				Relationships:    test.hasRelationships,
+			}
+			tb := virtual.NewTransformBuilder(&fakeCache)
+			raw, isSignal, err := common.GetUnstructured(test.input)
+			require.False(t, isSignal)
+			require.Nil(t, err)
```
MbolotSuse, line 189: Nit: I'm not sure that this should be part of this set of tests. The GetUnstructured call is part of the implementation detail of the Transform function - not something that is directly part of this package.

MbolotSuse (approved):
I've got a few concerns on implementation, but the functionality looks good. If we can't address the issues here due to timeline, let's get a separate issue up to track that as tech-debt.

pkg/resources/virtual/common/testutil.go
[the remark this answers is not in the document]
ericpromislow: Done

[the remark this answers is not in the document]
ericpromislow: Done

pkg/resources/virtual/events/events_test.go
```diff
@@ -0,0 +1,93 @@
...
+						"namespace": "gregsNamespace",
+					},
+					"id":    "eventTest1id",
+					"type":  "Gorniplatz",
+					"_type": "Gorniplatz",
+				},
+			},
+		},
+		{
+			name: "don't fix non-default-group event fields",
```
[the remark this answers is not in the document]
ericpromislow: Yes, I thought I could test core code that did it but it's done in a package that calls the `event` package, so it doesn't make sense here.

pkg/resources/virtual/virtual.go
```diff
@@ -16,13 +20,36 @@ type TransformBuilder struct {
...
-// GetTransformFunc retrieves a TransformFunc for a given GVK. Currently only returns a transformFunc for defaultFields
-func (t *TransformBuilder) GetTransformFunc(_ schema.GroupVersionKind) cache.TransformFunc {
-	return t.defaultFields.GetTransform()
+// GetTransformFunc returns the func to transform a raw object into a fixed object, if needed
+func (t *TransformBuilder) GetTransformFunc(gvk schema.GroupVersionKind) cache.TransformFunc {
+	converters := make([]func(*unstructured.Unstructured) (*unstructured.Unstructured, error), 0)
+	if gvk.Kind == "Event" && gvk.Group == "" && gvk.Version == "v1" {
+		converters = append(converters, events.TransformEventObject)
+	}
+	converters = append(converters, t.defaultFields.TransformCommon)
```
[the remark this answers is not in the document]
ericpromislow: I'd rather leave the transformers running from specific to general, until we find they can be run in any order.

PR #272, opened 2024-09-03
Added Namespace check to proxy_store.go Create

gehrkefc:
This is related to issue: [42788](https://github.com/rancher/rancher/issues/42788).

This adds a namespace check to ensure namespace when it's needed. Also it adds tests to Create function.

### DIFF
--- a/pkg/stores/proxy/proxy_store.go
+++ b/pkg/stores/proxy/proxy_store.go
@@ -13,15 +13,6 @@ import (
 	"strconv"
 
 	"github.com/pkg/errors"
-	"github.com/rancher/apiserver/pkg/types"
-	"github.com/rancher/steve/pkg/accesscontrol"
-	"github.com/rancher/steve/pkg/attributes"
-	metricsStore "github.com/rancher/steve/pkg/stores/metrics"
-	"github.com/rancher/steve/pkg/stores/partition"
-	"github.com/rancher/wrangler/v3/pkg/data"
-	corecontrollers "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
-	"github.com/rancher/wrangler/v3/pkg/schemas/validation"
-	"github.com/rancher/wrangler/v3/pkg/summary"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -34,9 +25,23 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+
+	"github.com/rancher/apiserver/pkg/types"
+	"github.com/rancher/wrangler/v3/pkg/data"
+	corecontrollers "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
+	"github.com/rancher/wrangler/v3/pkg/schemas/validation"
+	"github.com/rancher/wrangler/v3/pkg/summary"
+
+	"github.com/rancher/steve/pkg/accesscontrol"
+	"github.com/rancher/steve/pkg/attributes"
+	metricsStore "github.com/rancher/steve/pkg/stores/metrics"
+	"github.com/rancher/steve/pkg/stores/partition"
 )
 
-const watchTimeoutEnv = "CATTLE_WATCH_TIMEOUT_SECONDS"
+const (
+	watchTimeoutEnv      = "CATTLE_WATCH_TIMEOUT_SECONDS"
+	errNamespaceRequired = "metadata.namespace or apiOp.namespace are required"
+)
 
 var (
 	lowerChars  = regexp.MustCompile("[a-z]+")
@@ -429,15 +434,19 @@ func (s *Store) Create(apiOp *types.APIRequest, schema *types.APISchema, params 
 		input.SetNested(schema.ID[0:1]+"-", "metadata", "generateName")
 	}
 
-	if namespace == "" {
-		if apiOp.Namespace == "" {
-			return nil, nil, validation.ErrorCode{
-				Status: http.StatusUnprocessableEntity,
-				Code:   "metadata.namespace or apiOp.namespace are required",
+	if attributes.Namespaced(schema) {
+		if namespace == "" {
+			if apiOp.Namespace == "" {
+				return nil, nil, validation.ErrorCode{
+					Status: http.StatusUnprocessableEntity,
+					Code:   errNamespaceRequired,
+				}
 			}
+
+			namespace = apiOp.Namespace
 		}
 
-		input.SetNested(apiOp.Namespace, "metadata", "namespace")
+		input.SetNested(namespace, "metadata", "namespace")
 	}
 
 	gvk := attributes.GVK(schema)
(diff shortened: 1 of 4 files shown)


MbolotSuse (dismissed):
A couple of other changes that would be good:
- When a certain feature (ui-sql-cache) is enabled, there's a mirror to this file which is used. AFAICT, [that file](https://github.com/rancher/steve/blob/ea6c83b279c03ba2858639edd462985f1016c443/pkg/stores/sqlproxy/proxy_store.go#L510) will need this change as well - can you take a look at that?
- Typically we want tests to be added for each PR. Can you look into adding tests for this case (both for this change and the change in the sqlproxy/proxy_store)?

pkg/stores/proxy/proxy_store.go
```diff
@@ -422,20 +422,29 @@ func (s *Store) Create(apiOp *types.APIRequest, schema *types.APISchema, params
...
+
+	if name == "" && generateName == "" {
+		input.SetNested(schema.ID[0:1]+"-", "metadata", "generateName")
 	}
-	if ns == "" && apiOp.Namespace != "" {
-		ns = apiOp.Namespace
-		input.SetNested(ns, "metadata", "namespace")
+
+	if namespace == "" {
+		if apiOp.Namespace == "" {
```
MbolotSuse, line 433: @tomleb pointed out that this needs to consider cluster-scoped resources as well (so things like cluster roles). In those cases, you won't have a namespace or an apiOp.Namespace, but that is fine since those resources don't go in a namespace. I would take a look at `attributes.Namespaced(schema)` to see if that can give you sufficient information here.

gehrkefc: @MbolotSuse Added attributes.Namespaced check before this. Also added tests for Create function.

pkg/stores/proxy/proxy_store_test.go
```diff
@@ -121,3 +129,310 @@ func receiveUntil(wc chan watch.Event, d time.Duration) error {
 		}
 	}
 }
+
+func TestCreate(t *testing.T) {
```
gehrkefc, line 133: @ericpromislow Added those tests we've had chatted about.

pkg/stores/proxy/proxy_store.go
```diff
@@ -422,20 +427,33 @@ func (s *Store) Create(apiOp *types.APIRequest, schema *types.APISchema, params
...
+				return nil, nil, validation.ErrorCode{
+					Status: http.StatusUnprocessableEntity,
+					Code:   errNamespaceRequired,
+				}
+			}
+
+			namespace = apiOp.Namespace
+		}
+
+		input.SetNested(namespace, "metadata", "namespace")
```
tomleb, line 449: nit: Can be simplified a bit to remove 3 nested ifs. (Basically, no need to set the namespace if it's already set).


```suggestion
	if attributes.Namespaced(schema) && namespace == "" {
		if apiOp.Namespace == "" {
			return nil, nil, validation.ErrorCode{
				Status: http.StatusUnprocessableEntity,
				Code:   errNamespaceRequired,
			}
		}

		namespace = apiOp.Namespace
		input.SetNested(namespace, "metadata", "namespace")
```

pkg/stores/proxy/proxy_store_test.go
```diff
@@ -121,3 +129,310 @@ func receiveUntil(wc chan watch.Event, d time.Duration) error {
...
+					"metadata": map[string]interface{}{
+						"name": "testing-secret",
+					},
+				}},
+				warning: []types.Warning{},
+				err:     nil,
+			},
+		},
+		{
+			name: "missing name / generateName",
```
tomleb, line 248: Would be great to test the case where name is missing but generateName is there.

You can probably use the `createReactorFunc` field to set name to something, haven't tried this though so if you can't that's fine.

pkg/stores/proxy/proxy_store.go
```diff
@@ -34,9 +25,23 @@ import (
...
+	"github.com/rancher/steve/pkg/accesscontrol"
+	"github.com/rancher/steve/pkg/attributes"
+	metricsStore "github.com/rancher/steve/pkg/stores/metrics"
+	"github.com/rancher/steve/pkg/stores/partition"
 )
 
-const watchTimeoutEnv = "CATTLE_WATCH_TIMEOUT_SECONDS"
+const (
+	watchTimeoutEnv      = "CATTLE_WATCH_TIMEOUT_SECONDS"
+	errNamespaceRequired = "metadata.namespace or apiOp.namespace are required"
```
tomleb, line 43: This error will most likely end up in the UI and I doubt users will know what apiOp is or mean. We probably want something more generic like "namespace is required" or "metadata.namespace is required".

tomleb (commented):
Looks good, thanks for the added tests, just some comments/nits (note they apply to both regular proxy store & sql store).

pkg/stores/proxy/proxy_store_test.go
gehrkefc: Added one more test to accomplish that.

pkg/stores/proxy/proxy_store.go
gehrkefc: Agreed! Error message updated!

ericpromislow (dismissed):
LGTM

pkg/stores/proxy/proxy_store.go
```diff
@@ -422,20 +427,30 @@ func (s *Store) Create(apiOp *types.APIRequest, schema *types.APISchema, params
...
-	if ns == "" && apiOp.Namespace != "" {
-		ns = apiOp.Namespace
-		input.SetNested(ns, "metadata", "namespace")
+
+	if attributes.Namespaced(schema) && namespace == "" {
+		if apiOp.Namespace == "" {
+			return nil, nil, validation.ErrorCode{
+				Status: http.StatusUnprocessableEntity,
+				Code:   errNamespaceRequired,
+			}
```
tomleb, line 442: Sorry I didn't pick this up on my previous review:

```json
{
  "type": "error",
  "links": {},
  "code": "metadata.namespace is required",
  "message": "",
  "status": 422
}
```

It think to be more in line with the current errors we return it should look like:

```json
{
  "type": "error",
  "links": {},
  "code": "InvalidBodyContent",
  "message": "metadata.namespace is required",
  "status": 422
}
```

For example, we do the following in other places in Steve: https://github.com/rancher/steve/blob/bdb38275a6852620aa0f8de2f83420274dfd1f31/pkg/schema/definitions/handler.go#L23-L30.

So it would look something like this:

```go
return nil, nil apierror.NewAPIError(invalidBodyContent, "metadata.namespace is required")
```

where invalidBodyContent is already defined here: https://github.com/rancher/wrangler/blob/f9725ad6a04e725835441bc897b7b637f4c7e5ee/pkg/schemas/validation/error.go#L29. I'm not sure if we need a custom ErrorCode or can re-use the InvalidBodyContent so I'd suggest having a look at current usage of InvalidBodyContent if that makes sense here.

gehrkefc: @tomleb Just updated. Another look ?

tomleb (approved):
LGTM

ericpromislow (approved):
latest change ('changed return to use apierror.NewAPIError') looks fine.,

PR #273, opened 2024-09-04
Display error output from dynamiclistener.Server when in debug mode.

ericpromislow:
Related to both issues [#46956](https://github.com/rancher/rancher/issues/46956)  (v2.10)
and [#46957](https://github.com/rancher/rancher/issues/46957)  (v2.9)

This change let's the user of the Steve CLI decide whether to see panic output with `--debug` on,
or squelched otherwise.

### DIFF
--- a/main.go
+++ b/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"os"
 
+	"github.com/rancher/dynamiclistener/server"
 	"github.com/rancher/steve/pkg/debug"
 	stevecli "github.com/rancher/steve/pkg/server/cli"
 	"github.com/rancher/steve/pkg/version"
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -38,5 +39,5 @@ func run(_ *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	return s.ListenAndServe(ctx, config.HTTPSListenPort, config.HTTPListenPort, nil)
+	return s.ListenAndServe(ctx, config.HTTPSListenPort, config.HTTPListenPort, &server.ListenOpts{DisplayServerLogs: debugconfig.Debug})
```
tomleb, line 42: Those logs should pretty much always be set. I shouldn't have to run with `--debug` to see panics imo. They are error logs.

ericpromislow: Then we should drop that `DisplayServerLogs` option and always turn on "logging" (remember that only error-level output comes out of this process)

tomleb: No, we still need the option to keep the old behavior for other projects (k3s, rke2). Steve, Rancher and webhook will turn it on by default (eg: `DisplayServerLogs: true`) but the other projects won't (or they might, but we don't decide this).

tomleb (approved):
LGTM

MbolotSuse (approved):

PR #277, opened 2024-09-11
Add dumb watch mechanism for Vai

tomleb:

Support watch with Vai using a debouncer


pkg/stores/sqlproxy/proxy_store.go
```diff
@@ -676,53 +533,43 @@ func (s *Store) ListByPartitions(apiOp *types.APIRequest, schema *types.APISchem
...
 	}
 
-	go func() {
-		defer close(result)
-		<-ctx.Done()
-		eg.Wait()
-		cancel()
-	}()
+	// XXX: Why was this needed at all??
+	apiOp = apiOp.Clone().WithContext(ctx)
```
moio, line 553: This was copied over from the non-SQLite implementation of Watch, which has this line:

https://github.com/rancher/steve/blob/99e479ba0f0883bf7337b3eba8a67200dcc7509c/pkg/stores/partition/store.go#L295

That seems to come from a distant past:

https://github.com/rancher/steve/commit/a9c39ef89b26ad9ef52d9fdfef5249da894ad7ad#diff-74cdf37ae3cf46d709e040c4c6e22a30e4b62586165b5f60c71f2e3f02514d98R132

The only reason I can see is we do not want to change the passed in `apiOp`, as it is passed by pointer. As Stores can be nested, I can see a calling Store not to expect the called Store to modify the passed `apiOp` to add a context to the nested `Request` the `apiOp` wraps.

So my understanding is that cloning here has the effect to have the context be applied to callees exclusively, but not callers. Whether this really prevents any bugs I am not sure.

pkg/stores/sqlproxy/proxy_store.go
```diff
@@ -676,53 +533,43 @@ func (s *Store) ListByPartitions(apiOp *types.APIRequest, schema *types.APISchem
...
-		return s.Watch(apiOp, schema, wr)
+	inf, err := s.CacheFor(&tablelistconvert.Client{ResourceInterface: client}, schema)
+	if err != nil {
+		return nil, err
 	}
 
-	apiOp.Namespace = partition.Namespace
-	if partition.All {
-		return s.Watch(apiOp, schema, wr)
+	debounceListener := newDebounceListener(5 * time.Second)
```
moio, line 567: Of course we will have to make sure the listener time is configurable - at least via configuration object or environment variable, in future possibly on a request-by-request basis.

moio (commented):
Questions:

 - will changes to resources a user does not have access to still result in the resource counter increasing?
 - do you have any plans to handle special Steve-only resources such as counters, that come with no revisions? How does watch work currently, can we keep the mechanism as-is?

tomleb:
At this point we're mostly still experimenting so it doesn't make sense to have a PR opened. Closing.

PR #284, opened 2024-09-24
sql: silence informer errors on unwatchable resources

moio:
Listable-but-unwatchable resources such as `nodes metrics.k8s.io/v1beta1` allow an Informer to be started, but it will by default log all watch failures.

Detect unwatchable types and tell Lasso so that errors can be silenced

partially addresses https://github.com/rancher/rancher/issues/46030

_best reviewed commit-by-commit_


tomleb (approved):
LGTM.

crobby (approved):
lgtm

PR #286, opened 2024-09-30
Add tests for concurrent AccessControl store usage

aruiz14:
## Issue: https://github.com/rancher/rancher/issues/47318

This PR does not change the behavior of the code, just a small refactor to make it more testable and an intentionally failing test (later disabled) to prove the problem.

### DIFF
--- a/pkg/accesscontrol/access_store_test.go
+++ b/pkg/accesscontrol/access_store_test.go
@@ -301,6 +301,7 @@ func (c *spyCache) observeAdd(k interface{}) {
 }
 
 func TestAccessStore_AccessFor_concurrent(t *testing.T) {
+	t.Skipf("TODO - Add a fix for this test")
 	testUser := &user.DefaultInfo{Name: "test-user"}
 	asCache := &spyCache{accessStoreCache: cache.NewLRUExpireCache(100)}
 	store := &AccessStore{
--- a/pkg/accesscontrol/access_store_test.go
+++ b/pkg/accesscontrol/access_store_test.go
@@ -3,7 +3,9 @@ package accesscontrol
 import (
 	"fmt"
 	"slices"
+	"sync"
 	"testing"
+	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


moio (approved):
LGTM

pkg/accesscontrol/access_store_test.go
```diff
@@ -274,6 +276,81 @@ func TestAccessStore_AccessFor(t *testing.T) {
...
+							{Namespace: All, ResourceName: All}: true,
+						},
+					},
+				}
+			},
+		},
+		cache: asCache,
+	}
+
+	const n = 5 // observation showed cases with up to 5 (or more) concurrent queries for the same user
```
tomleb, line 330: Can you elaborate on what it is we think is the wrong behavior / concern?

I can see that `l.cache.Add` is being called multiple times, but it will always have the same ID so the underlying cache (`*LRUCache`) will only have one of those (for the same key).

So is the issue that we're computing multiple times the result sets for nothing?

Another issue I'm seeing with the current code is there's a race between computing the cacheKey and creating the resulting `AccessSet`. The cacheKey is created with set `A` of RBAC resources, and there might be some changes to RBAC resources that that would lead to the `AccessSet` being created with a set `B` of RBAC resources. Is this what this test is after?

tomleb (commented):
The early nil check is a nice change, makes it easier to follow!

Mostly question about the purpose of this test / PR as I wasn't sure I understood completely.

pkg/accesscontrol/access_store_test.go
aruiz14: You are correct in both observations.
1. This test aims to tackle the first one, with the additional concern of not only the multiple computation for the same key, but also how it's used afterwards (e.g. `AccessSets` seem to be embedded into the `Schemas` type). There seems to be some sort of indirection that makes the original problem very hard to troubleshoot.
2. Indeed this is a separate problem I also notice but it does not have a straightforward solution. I could take a look after this and try to find a viable alternative. In any case, I believe the solution to problem 1) will mitigate the impact of this other problem to some extent.

tomleb (approved):
LGTM.

tomleb:
Merging as asked by @aruiz14

pkg/accesscontrol/access_store_test.go
aruiz14: Following up on problem 2):
https://github.com/rancher/steve/pull/292

PR #298, opened 2024-10-15
Fix ExtensionAPIServer interface

tomleb:
# Issue https://github.com/rancher/rancher/issues/47030

There wasn't a type assertion for the `ExtensionAPIServer` interface and unfortunately the interface did not match the "default" implementation.

We're fixing it here and adding an assertion to prevent the same problem happening in the future.


gehrkefc (approved):
LGTM

maxsokolovsky (approved):

PR #299, opened 2024-10-17
#47483 - Adding NonResourceURLs support to AccessStore

gehrkefc:
This is related to issue: [#47483](https://github.com/rancher/rancher/issues/47483)


pkg/accesscontrol/policy_rule_index.go
```diff
@@ -85,25 +87,29 @@ func (p *policyRuleIndex) get(subjectName string) *AccessSet {
...
-						accessSet.Add(verb,
-							schema.GroupResource{
-								Group:    group,
-								Resource: resource,
-							}, Access{
-								Namespace:    namespace,
-								ResourceName: resourceName,
-							})
+		if len(rule.Resources) > 0 {
+			for _, group := range rule.APIGroups {
```
tomleb, line 91: Could we perhaps extract some of this to their own methods so we reduce the nesting a bit?

pkg/accesscontrol/policy_rule_index.go
```diff
@@ -85,25 +87,29 @@ func (p *policyRuleIndex) get(subjectName string) *AccessSet {
...
+									Resource: resource,
+								}, Access{
+									Namespace:    namespace,
+									ResourceName: resourceName,
+								})
+						}
 					}
 				}
 			}
+		} else if roleRef.Kind == clusterRoleKind {
```
tomleb, line 111: Should this check that the binding is a clusterrolebinding and not a rolebinding? Seems like for clusterrolebinding we're setting namespace as `All` (which is `*`)

pkg/accesscontrol/access_set.go
```diff
@@ -56,6 +65,13 @@ func (a *AccessSet) Merge(right *AccessSet) {
 			m[k] = v
 		}
 	}
+
+	for k, v := range right.nonResourceSet {
+		_, ok := a.nonResourceSet[k]
+		if !ok {
+			a.nonResourceSet[k] = v
+		}
+	}
```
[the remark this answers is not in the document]
gehrkefc: Ok. Makes sense. I'll add code to merge rules.

pkg/accesscontrol/access_set_test.go
```diff
@@ -0,0 +1,303 @@
...
+				assert.Contains(t, rules, tt.expected, "rule not found")
+			} else {
+				assert.NotContains(t, rules, tt.expected, "unexpected rule found")
+			}
+		})
+
+	}
+}
+
+func TestAccessSet_GrantsNonResource(t *testing.T) {
```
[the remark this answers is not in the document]
gehrkefc: Ok, I'll add it.

pkg/ext/apiserver_authorization_test.go
```diff
@@ -342,3 +345,152 @@ func (s *ExtensionAPIServerSuite) TestAuthorization() {
 		})
 	}
 }
+
+func TestAuthorization_NonResourceURLs(t *testing.T) {
```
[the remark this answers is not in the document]
gehrkefc: Are these urls being served in any way ?

pkg/ext/apiserver_authorization.go
```diff
@@ -22,20 +22,25 @@ func NewAccessSetAuthorizer(asl accesscontrol.AccessSetLookup) *AccessSetAuthori
...
+	accessSet := a.asl.AccessFor(attrs.GetUser())
+
 	if !attrs.IsResourceRequest() {
-		// XXX: Implement
-		return authorizer.DecisionDeny, "AccessSetAuthorizer does not support nonResourceURLs requests", nil
+		if accessSet.GrantsNonResource(verb, path) {
+			return authorizer.DecisionAllow, "", nil
+		}
+
+		return authorizer.DecisionDeny, "", nil
```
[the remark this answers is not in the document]
gehrkefc: Ok, Will do.

pkg/accesscontrol/policy_rule_index.go
gehrkefc: I'm sorry, what do you mean exactly ?

pkg/ext/apiserver_authorization_test.go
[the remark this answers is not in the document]
tomleb: Yes these urls are expected to be served to user so they can access the public documentation (OpenAPI spec)

pkg/accesscontrol/policy_rule_index.go
tomleb: Basically, we're doing nested-ifs and nested-loops 6 levels deep which affects the readability. My suggestion would be to split it a little in two functions to make it more readable. eg: the nonResourceURL case and the resourceURL case.

pkg/accesscontrol/access_set.go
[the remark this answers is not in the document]
gehrkefc: I changed my mind. Changed it to be `map[nonResourceKey]struct{}`. I've analyzed it and it makes more sense this way. Considering we'll have a look check the rules inside the function from kubernetes.

pkg/ext/apiserver_authorization_test.go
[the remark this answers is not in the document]
gehrkefc: Ok, some bindings using read-only and read-write users (already present) and created 6 new tests.

pkg/accesscontrol/policy_rule_index.go
gehrkefc: Created another function just to handle resource based rules. It became this way:

https://github.com/rancher/steve/blob/026f994f5803054edb249ac5f2080457b4b50c18/pkg/accesscontrol/policy_rule_index.go#L89-L94

pkg/accesscontrol/access_set_test.go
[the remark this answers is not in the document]
gehrkefc: Added.

pkg/accesscontrol/policy_rule_index.go
gehrkefc: I'm checking if the kind is "ClusterRole" and not setting the namespace, does it mean "All" in k8s terms ?

pkg/ext/apiserver_authorization.go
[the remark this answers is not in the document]
gehrkefc: Checked it integrating with rancher and added a comment with the response message.

pkg/accesscontrol/access_set_test.go
```diff
@@ -0,0 +1,190 @@
...
+			verb:   "post",
+			url:    "/healthz",
+			keys:   map[nonResourceKey]struct{}{},
+			expect: false,
+		},
+	}
+
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
+			accessSet := &AccessSet{nonResourceSet: tt.keys}
```
tomleb, line 128: nit: WDYT of using the exported method to construct the AccessSet? That way if the implementation changes we most likely won't have to change the test also.

pkg/ext/apiserver_authorization_test.go
```diff
@@ -311,6 +314,66 @@ func (s *ExtensionAPIServerSuite) TestAuthorization() {
 			},
 			expectedStatusCode: http.StatusForbidden,
 		},
+		{
```
tomleb, line 317: I find it a bit confusing that `read-only` means read-only but only for all resources AND for openapi v2, but not v3. I think it would be less confusing to make up new users specifically for those cases. eg: a `openapi-v2-only-read` user and a `openapi-read-only` user, or something similar.

tomleb (commented):
Looks good, a couple of things for tests but otherwise :+1:

pkg/ext/apiserver_authorization_test.go
gehrkefc: Ok, user changed!

pkg/accesscontrol/access_set_test.go
gehrkefc: Agree, just changed it.

pkg/accesscontrol/policy_rule_index.go
gehrkefc: I have this test at GrantsNonResource, should I replicate it here ?

pkg/accesscontrol/access_set_test.go
```diff
@@ -0,0 +1,190 @@
...
+	}{
+		{
+			name:  "valid case",
+			verbs: []string{"get", "post"},
+			urls:  []string{"/healthz", "/metrics"},
+			want: []nonResourceKey{
+				{"get", "/healthz"},
+				{"get", "/metrics"},
+				{"post", "/healthz"},
+				{"post", "/metrics"},
```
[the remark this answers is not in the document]
gehrkefc: Ok, I had this in GrantsNonResource. Added it here too.

pkg/accesscontrol/access_set.go
```diff
@@ -120,6 +160,25 @@ func (a *AccessSet) Add(verb string, gr schema.GroupResource, access Access) {
 	}
 }
 
+func (a *AccessSet) AddNonResouceURLs(verbs, urls []string) {
```
ericpromislow, line 163: typo: should be `AddNonResourceURLs`

pkg/accesscontrol/access_set_test.go
```diff
@@ -0,0 +1,210 @@
...
+			want:  nil,
+		},
+	}
+
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
+			accessSet := &AccessSet{}
+			accessSet.AddNonResouceURLs(tt.verbs, tt.urls)
+
+			if len(tt.want) > 0 {
```
ericpromislow, line 62: Shouldn't there be an `else` part here to assert that `accessSet.nonResourceSet` doesn't contain certain values?  I should add that usually testing a negative doesn't give as much information as testing a positive, so this might not be a good use of time. Just asking

ericpromislow (changes requested):
Just one typo

pkg/accesscontrol/access_set.go
gehrkefc: Just fixed it. Thanks!

pkg/accesscontrol/access_set_test.go
gehrkefc: Ok, just added it. Thank you!

tomleb (approved):
LGTM, let's wait after code freeze before merging this

ericpromislow (approved):
LGTM

PR #317, opened 2024-11-07
Index arbitrary labels

ericpromislow:
Related to [#46333](https://github.com/rancher/rancher/issues/46333)


pkg/resources/virtual/common/common.go
```diff
@@ -30,9 +30,35 @@ func (d *DefaultFields) TransformCommon(obj *unstructured.Unstructured) (*unstru
...
+func (d *DefaultFields) TransformLabels(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	err := addLabelFields(obj)
+	if err != nil {
+		return nil, fmt.Errorf("unable to add label fields: %w", err)
+	}
 	return obj, nil
 }
 
+func addLabelFields(raw *unstructured.Unstructured) error {
+	labels := raw.GetLabels()
```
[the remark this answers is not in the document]
ericpromislow: True, good catch.

pkg/resources/virtual/common/common.go
```diff
@@ -74,6 +100,15 @@ func addIDField(raw *unstructured.Unstructured) *unstructured.Unstructured {
 	return raw
 }
 
+// updateTypeField replaces the _type field with the contents of the field named "type", if it exists
+func fixTypeField(raw *unstructured.Unstructured) *unstructured.Unstructured {
```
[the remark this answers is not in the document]
ericpromislow: This function is part of the transformFunc chain at https://github.com/ericpromislow/steve/blob/f8103ff621f792ea1e6a9399d087f170b1a54f8f/pkg/resources/virtual/virtual.go#L48

pkg/resources/virtual/common/common.go
```diff
@@ -74,6 +100,15 @@ func addIDField(raw *unstructured.Unstructured) *unstructured.Unstructured {
...
+	currentTypeValue, ok := raw.Object["type"]
```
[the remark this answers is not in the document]
ericpromislow: Turns out this function was replaced by `TransformEventObject`, which returns a nil error because it's part of that same object-transform loop.

pkg/stores/sqlpartition/listprocessor/processor_test.go
```diff
@@ -266,6 +266,36 @@ func TestParseQuery(t *testing.T) {
 			return nil
 		},
 	})
+	tests = append(tests, testCase{
+		description: "ParseQuery() with a labels filter param " +
```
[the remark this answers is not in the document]
ericpromislow: ok. I'm still a slave to the 80-character guideline.

pkg/resources/virtual/common/common.go
```diff
@@ -30,9 +30,35 @@ func (d *DefaultFields) TransformCommon(obj *unstructured.Unstructured) (*unstru
...
 		return nil, fmt.Errorf("unable to add summary fields: %w", err)
 	}
+	err = addLabelFields(obj)
+	if err != nil {
+		return nil, fmt.Errorf("unable to add label fields: %w", err)
+	}
+	return obj, nil
+}
+
+// TransformLabels caches the labels
```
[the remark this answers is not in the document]
ericpromislow: Yes, and I see the code was calling `TransformLabels` twice. The second call has no effect, but is confusing.

pkg/stores/sqlpartition/queryparser/labels.go
```diff
@@ -0,0 +1,143 @@
+package queryparser
```
bigkevmcd, line 1: Should this not retain its original license header?

ericpromislow: good point. I made a few changes but let me look into it

pkg/stores/sqlpartition/queryparser/selector.go
bigkevmcd, line 1: It feels like we could/should provide some indication of what's changed from the upstream?

In case we need to pull in other upstream changes?

bigkevmcd (changes requested):
Mostly fine, I would much prefer to see some real tests against SQLite rather than the continued use of mocks.

ericpromislow:
"I would much prefer to see some real tests against SQLite rather than the continued use of mocks."

This is covered in tech-debt item
https://github.com/rancher/rancher/issues/48263
(I edited the subject to include steve)

pkg/stores/sqlpartition/listprocessor/processor.go
```diff
@@ -53,6 +56,72 @@ type Cache interface {
 	ListByOptions(ctx context.Context, lo informer.ListOptions, partitions []partition.Partition, namespace string) (*unstructured.UnstructuredList, int, string, error)
 }
 
+// k8sRequirementToOrFilter - convert one k8s Requirement to a list of Filter's:
+
+//type Requirement struct {
```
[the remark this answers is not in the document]
ericpromislow: I just had those in place while writing the code. Forgot to delete.

pkg/resources/virtual/common/common.go
```diff
@@ -33,6 +33,24 @@ func (d *DefaultFields) TransformCommon(obj *unstructured.Unstructured) (*unstru
 	return obj, nil
 }
 
+// TransformLabels transfers the object's kubernetes labels to the cached part of `obj`
+func (d *DefaultFields) TransformLabels(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	addLabelFields(obj)
+	return obj, nil
+}
+
+func addLabelFields(raw *unstructured.Unstructured) {
```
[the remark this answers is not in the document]
ericpromislow: I broke the code for `addLabelFields` out into a separate function because of the way it modifies its `*unstructured.Unstructured` .

I added some uniut tests, and I see that they pass if `addLabelFields` is a no-op. All it does is copy the hash from `obj.GetLabels()` into a new hash, and assigns it to that same field. This only makes sense if later code modifies that hash in place, and the original object depends on the original values, but it's not obvious to me why that would be the case.

So I'm going to just commit, and then delete `addLabelFields` in a subsequent commit, and I'lI wait to here if I should reinstate it.

pkg/stores/sqlpartition/listprocessor/processor.go
```diff
@@ -53,6 +56,72 @@ type Cache interface {
...
+		selection.GreaterThan:  informer.Gt,
+	}
+	v, ok := h[k8sOp]
+	if ok {
+		return v, nil
+	}
+	return "", fmt.Errorf("unknown k8sOp: %s", k8sOp)
+}
+
+func isQuotedStringTarget(values []string) bool {
```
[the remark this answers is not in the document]
ericpromislow: Sure

pkg/stores/sqlpartition/listprocessor/processor.go
```diff
@@ -162,6 +228,14 @@ func getLimit(apiOp *types.APIRequest) int {
 	return limit
 }
 
+func splitQuery(query string) []string {
```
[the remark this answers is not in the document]
ericpromislow: Sure. It's not obvious (and before we were just doing `strings.Split(".", query)` which doesn't work for complex label names.

pkg/stores/sqlpartition/listprocessor/processor_test.go
```diff
@@ -505,18 +817,29 @@ func TestParseQuery(t *testing.T) {
...
+			if x != nil {
+				fmt.Printf("stop here")
+			}
 			lo, err := ParseQuery(test.req, test.nsc)
 			if test.errExpected {
 				assert.NotNil(t, err)
+				if test.errorText != "" {
+					assert.Contains(t, test.errorText, err.Error())
+				}
+				//assert.Equal(t, err, errors.New("unable to parse requirement: existence tests are valid only for labels; not valid for field 'a5In1'"))
```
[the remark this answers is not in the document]
ericpromislow: It should be deleted. It made sense at one point but became obsolete.

pkg/stores/sqlpartition/queryparser/selector.go
ericpromislow: I considered maintaining this file as a patch off the upstream file, but I didn't see any actual changes to the language at all in the `apimachinery/pkg/labels/selector.go` file since 2017, which implies it was moved from another file in that year. Assuming that future upstream changes to the label-selection language will be few and far between, it seemed simpler to copy and modify rather than create a patch file and add build complexity.

pkg/stores/sqlpartition/listprocessor/processor.go
```diff
@@ -53,6 +56,61 @@ type Cache interface {
...
+// Determine if the value field is surrounded by a pair of single- or double-quotes
+// This is a difference we implement from the CLI: if the target value of a (not) equal
+// test is quoted, we use the full string. Otherwise we do a substring match
+// (which is implemented as 'SELECT ... some-field ... LIKE "%VALUE%" ...' in the query)
+func isQuotedStringTarget(values []string) bool {
+	if len(values) != 1 || len(values[0]) == 0 {
+		return false
+	}
+	s1 := values[0][0:1]
+	if !strings.Contains(`"'`, s1) {
```
moio, line 87: When filtering for exact matches, do we want the string to be single quoted, double quoted, or support both? I am asking because the current behavior seem to be single quote:

https://github.com/rancher/steve/blame/d05a48f9cfa6a43a5fbb840cc5da00a92b229009/pkg/stores/sqlpartition/listprocessor/processor.go#L80

and IIUC this change would make it either single or double quote. Alas, we never documented the choice on the README.


@richard-cox do you have a preference?

@ericpromislow can you amend the README so that whatever we decide is documented please?

pkg/stores/sqlpartition/queryparser/labels.go
```diff
@@ -0,0 +1,164 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+This file is derived from
+https://github.com/kubernetes/apimachinery/blob/master/pkg/labels/labels.go
```
moio, line 19: Looking at the diff between the file pointed by this link and the file in this PR I see they have:

 - a `FormatLabels` func we don't
 - more validation in `ConvertSelectorToLabelsMap` (but we do not use that func)

Is this correct? If so I would suggest:

```suggestion
This file is derived from
https://github.com/kubernetes/apimachinery/blob/90df4d1d2d40ea9b3a522bec6e3577237358de00/pkg/labels/labels.go

 - FormatLabels was dropped
 - validateLabelKey calls from ConvertSelectorToLabelsMap were dropped
```

That will help if ever this needs bumping

pkg/stores/sqlpartition/queryparser/selector.go
```diff
@@ -0,0 +1,965 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+This file is derived from
+https://github.com/kubernetes/apimachinery/blob/master/pkg/labels/selector.go
```
moio, line 19: Similar to the point above, I think it helps to point to the exact commit and be descriptive about changes.

pkg/stores/sqlproxy/proxy_store.go
```diff
@@ -85,6 +86,7 @@ var (
 			{"spec", "nodeName"}},
 		gvkKey("", "v1", "Service"): {
 			{"spec", "clusterIP"},
+			{"spec", "targetPort"},
```
moio, line 89:
```suggestion
```

This could be a leftover from an earlier rebase - the field got removed by https://github.com/rancher/steve/pull/433

moio (dismissed):
Thanks @ericpromislow, I do not see big issues either. Most comments from my part are really suggestions and should be quick to address.

Ready to approve as soon as they are dealt with.

👀 from @tomleb would also be helpful

pkg/stores/sqlpartition/listprocessor/processor.go
```diff
@@ -162,6 +217,21 @@ func getLimit(apiOp *types.APIRequest) int {
...
 
+// splitQuery takes a single-string metadata-labels filter and converts it into an array of 3 accessor strings,
+// where the first two strings are always "metadata" and "labels", and the third is the label name.
+// This is more complex than doing something like `strings.Split(".", "metadata.labels.fieldName")
+// because the fieldName can be more complex - in particular it can contain "."s) and needs to be
+// bracketed, as in `metadata.labels[rancher.io/cattle.and.beef]".
+// The `labelsRegex` looks for the bracketed form.
+func splitQuery(query string) []string {
+	m := labelsRegex.FindStringSubmatch(query)
+	if m != nil && len(m) >= 4 {
```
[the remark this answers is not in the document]
ericpromislow: True, if that regexp matches `m` should have exactly 4 members (the first being the fuliy matched string for those jumping in here and not reading all the code).

pkg/stores/sqlpartition/queryparser/labels.go
ericpromislow: ok

pkg/stores/sqlpartition/queryparser/selector.go
ericpromislow: Sure

pkg/stores/sqlproxy/proxy_store.go
ericpromislow: Looks like the field was preserved during a rebase-commit that I resolved incorrectly by keeping that field.

pkg/stores/sqlpartition/queryparser/selector.go
moio: discussion continued in https://github.com/rancher/steve/pull/317#discussion_r1918198613, resolving this thread

pkg/stores/sqlpartition/queryparser/selector.go
```diff
@@ -0,0 +1,965 @@
...
+		string(selection.In), string(selection.NotIn),
+		string(selection.Equals), string(selection.DoubleEquals), string(selection.NotEquals),
+		string(selection.GreaterThan), string(selection.LessThan),
+	}
+	validRequirementOperators = append(binaryOperators, unaryOperators...)
+	labelSelectorRegex        *regexp.Regexp
+)
+
+func init() {
+	labelSelectorRegex = regexp.MustCompile(`^metadata.labels[[.]`)
```
moio, line 74:
```suggestion
	labelSelectorRegex = regexp.MustCompile(`^metadata\.labels\[.`)
```

This:
 - escapes `.` (else `"metadataAlabels[abc"` will match)
 - escapes `[` (go still considers the regexp valid, but it's FMPOV prone to confuse readers as it looks like the start of a character class)
 - removes `[.]` in favor of `.`, which match the same set of strings

Please note this regex will still match incomplete selectors like `"metadata.labels[asd"`. If this is sufficient (IIUC it is, but please double check), then no other change is needed and this comment can be resolved after the suggestion is accepted.

moio (dismissed):
Looks good from my perspective.

We definitely need:

- @richard-cox's perspective on https://github.com/rancher/steve/pull/317/files?w=0#r1918171835 / https://github.com/rancher/steve/pull/317/files?w=0#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R120-R136

- another look from somebody in Frameworks (@tomleb ?) especially from a maintenance perspective

pkg/stores/sqlpartition/listprocessor/processor.go
richard-cox: Currently
- `'` denotes exact match
- no quotes denotes partial match

From my interpretation of docs the proposal would be the following?
- `'` denotes exact match
- `"` denotes partial match
- no quotes syntax has been removed?

I think ideally we would continue as currently. all other params, include those that contain text, do not require `"`

ericpromislow: No, quotes are still optional. And I've implemented single quotes indicate an exact match, and double quotes don't (they're just used to delimit the contents, but because of the syntax of k8s labels, they're never required because characters like spaces aren't allowed).

pkg/stores/sqlpartition/listprocessor/processor.go
```diff
@@ -53,6 +56,71 @@ type Cache interface {
...
+		return false, false
+	}
+	if !strings.HasSuffix(values[0], s1) {
+		return false, false
+	}
+	return true, s1[0] == '\''
+}
+
+// k8sRequirementToOrFilter - convert one k8s Requirement to a list of Filter's:
+
```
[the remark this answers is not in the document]
ericpromislow: ok

pkg/stores/sqlpartition/queryparser/selector.go
ericpromislow: I thought I only needed to check for the start of a label reference, but adding `"metadata.labels[missing/close-bracket",` to the list of bad strings in `TestSelectorParse` gave a failing test, meaning nothing was complaining about it. So I tightened up the regex to require a full match of a valid label key.

pkg/stores/sqlpartition/queryparser/selector.go
```diff
@@ -0,0 +1,965 @@
...
+		string(selection.In), string(selection.NotIn),
+		string(selection.Equals), string(selection.DoubleEquals), string(selection.NotEquals),
+		string(selection.GreaterThan), string(selection.LessThan),
+	}
+	validRequirementOperators = append(binaryOperators, unaryOperators...)
+	labelSelectorRegex        *regexp.Regexp
+)
+
+func init() {
+	labelSelectorRegex = regexp.MustCompile(`^metadata.labels(?:\.\w[-a-zA-Z0-9_./]*|\[.*])$`)
```
moio, line 74: I am still not too convinced about this RE

See some suspicious strings passing in https://go.dev/play/p/AN8J0aUS0T0

How about:

```suggestion
	labelSelectorRegex = regexp.MustCompile(`^metadata\.labels\[\w[a-zA-Z0-9_-\./]*\]`)
```

moio (changes requested):

pkg/stores/sqlpartition/listprocessor/processor.go
moio: Do we need double quotes at all?

pkg/stores/sqlproxy/proxy_store.go
```diff
@@ -306,11 +307,13 @@ func (s *Store) initializeNamespaceCache() error {
 	// get any type-specific fields that steve is interested in
 	fields = append(fields, getFieldForGVK(gvk)...)
 
-	// get the type-specifc transform func
+	// get the type-specific transform func
 	transformFunc := s.transformBuilder.GetTransformFunc(gvk)
 
 	// get the ns informer
-	nsInformer, err := s.cacheFactory.CacheFor(fields, transformFunc, &tablelistconvert.Client{ResourceInterface: client}, attributes.GVK(&nsSchema), false, true)
+	client2 := &tablelistconvert.Client{ResourceInterface: client}
```
tomleb, line 314: nit:
```suggestion
	tableClient := &tablelistconvert.Client{ResourceInterface: client}
```

pkg/stores/sqlproxy/proxy_store.go
```diff
@@ -749,8 +752,10 @@ func (s *Store) ListByPartitions(apiOp *types.APIRequest, schema *types.APISchem
 	fields := getFieldsFromSchema(schema)
 	fields = append(fields, getFieldForGVK(gvk)...)
 	transformFunc := s.transformBuilder.GetTransformFunc(gvk)
-
-	inf, err := s.cacheFactory.CacheFor(fields, transformFunc, &tablelistconvert.Client{ResourceInterface: client}, attributes.GVK(schema), attributes.Namespaced(schema), controllerschema.IsListWatchable(schema))
+	client2 := &tablelistconvert.Client{ResourceInterface: client}
+	attrs2 := attributes.GVK(schema)
+	ns2 := attributes.Namespaced(schema)
```
tomleb, line 757: I didn't see a second attributes or namespace in that function so we can remove the `2` suffix.

```suggestion
	tableClient := &tablelistconvert.Client{ResourceInterface: client}
	attrs := attributes.GVK(schema)
	ns := attributes.Namespaced(schema)
```

pkg/stores/sqlpartition/queryparser/selector.go
```diff
@@ -0,0 +1,965 @@
...
+	// LessThanToken represents less than
+	LessThanToken
+	// NotEqualsToken represents not equal
+	NotEqualsToken
+	// NotInToken represents not in
+	NotInToken
+	// OpenParToken represents open parenthesis
+	OpenParToken
+	SingleQuoteToken
+	DoubleQuoteToken
```
tomleb, line 368: I'm not sure those are necessary as they aren't used anywhere. For example, the tests passed even with them removed.

pkg/stores/sqlpartition/queryparser/selector.go
```diff
@@ -0,0 +1,965 @@
...
+	"==":    DoubleEqualsToken,
+	"=":     EqualsToken,
+	">":     GreaterThanToken,
+	"in":    InToken,
+	"<":     LessThanToken,
+	"!=":    NotEqualsToken,
+	"notin": NotInToken,
+	"(":     OpenParToken,
+	"'":     SingleQuoteToken,
+	"\"":    DoubleQuoteToken,
```
tomleb, line 386: I'm not sure those are necessary as they aren't used anywhere. For example, the tests passed even with them removed.

pkg/stores/sqlpartition/queryparser/selector.go
```diff
@@ -0,0 +1,965 @@
...
+	unaryOperators = []string{
+		string(selection.Exists), string(selection.DoesNotExist),
+	}
+	binaryOperators = []string{
+		string(selection.In), string(selection.NotIn),
+		string(selection.Equals), string(selection.DoubleEquals), string(selection.NotEquals),
+		string(selection.GreaterThan), string(selection.LessThan),
+	}
+	validRequirementOperators = append(binaryOperators, unaryOperators...)
+	labelSelectorRegex        *regexp.Regexp
```
tomleb, line 70:
```suggestion
	labelSelectorRegex        regexp.MustCompile(...)
```

This should work since you're in a `var` block

pkg/stores/sqlpartition/listprocessor/processor.go
tomleb: My first thought was it should probably be the operator that decides if its an exact match equality or a substring match. Something like: `=` is exact (stay closer to k8s), `=~` is substring. `!~` is not equal substring.

Imo we should have only one way per matching strategy, I don't see a reason to have both double quotes and no quotes denote substring match. That should simplify the logic a bit too.

tomleb (commented):
Mostly nits / styling, one question about the deepcopy, but overall LGTM

pkg/stores/sqlpartition/listprocessor/processor.go
ericpromislow: ok, double quotes are gone and trigger a syntax error

pkg/stores/sqlpartition/queryparser/selector.go
ericpromislow: From the playground snippet the regex is working as I intended -- it wants to see an identifier after the final `.`, or stuff in square brackets followed by the end of the string, but it doesn't care what goes in the square brackets. No error message is necessary -- it's enough for the user to see nothing matched and adjust the expression.  Rather than trying to match every possible error, I'm treating `[J*U@N+K]` as the same as if the user typed `metadata.labels.box' when they meant `metadata.labels.boy` .  They'll see that `box` resulted in zero hits, ask themselves why, and fix the search term.

pkg/stores/sqlpartition/queryparser/zz_generated.deepcopy.go
```diff
@@ -0,0 +1,48 @@
+//go:build !ignore_autogenerated
```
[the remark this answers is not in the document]
ericpromislow: I don't have any opinion on this. All this code came as-is from kubectl.

pkg/stores/sqlproxy/proxy_store.go
ericpromislow: sure

pkg/stores/sqlproxy/proxy_store.go
ericpromislow: I broke these out from the long call to `s.cacheFactory.CacheFor` because the overly long line was harder to digest, and frustrating when stepping through this code in the debugger.

pkg/stores/sqlpartition/queryparser/selector.go
ericpromislow: Yes, dropped in an earlier commit

pkg/stores/sqlpartition/queryparser/selector.go
ericpromislow: Yup, dropped

pkg/stores/sqlpartition/queryparser/selector.go
ericpromislow: Yes

ericpromislow:
> My first thought was it should probably be the operator that decides if its an exact match equality or a substring match. Something like: = is exact (stay closer to k8s), =~ is substring. !~ is not equal substring.

> Imo we should have only one way per matching strategy, I don't see a reason to have both double quotes and no quotes denote substring match. That should simplify the logic a bit too.

The requirements come form the dashboard, and they're the main consumer of this. So they'll need to weigh in on these operators.  But I added the double-quotes as a sort of symmetric thing and they're easily dropped.

pkg/stores/sqlpartition/queryparser/zz_generated.deepcopy.go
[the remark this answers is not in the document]
tomleb: So imo if we don't have anything to regenerate the file we shouldn't add the autogenerated file ourselves if we don't need it. In this case I don't think we need the deepcopy code (I might be wrong, you'll have to try to remove it see if things still compile), so I'd suggest just removing it.

pkg/stores/sqlpartition/queryparser/selector.go
tomleb: But I still see them so it looks like they were brought back? Unless I misunderstand. (eg: I'm fairly sure we can remove both `SingleQuoteToken` and `DoubleQuoteToken` here.

pkg/stores/sqlpartition/queryparser/selector.go
moio: OK, I misunderstood that `metadata.labels.something` is valid, sorry about that.

Nevertheless IIUC it still holds that `metadataAlabels.something` or `metadataXlabels.something` should not match because they do not have a literal dot between `metadata` and `labels`, and I would still recommend escaping characters that are not interpreted as special but could be, just for clarity.

I will suggest a different change.

moio:
> > My first thought was it should probably be the operator that decides if its an exact match equality or a substring match. Something like: = is exact (stay closer to k8s), =~ is substring. !~ is not equal substring.
> 
> > Imo we should have only one way per matching strategy, I don't see a reason to have both double quotes and no quotes denote substring match. That should simplify the logic a bit too.
> 
> The requirements come form the dashboard, and they're the main consumer of this. So they'll need to weigh in on these operators. But I added the double-quotes as a sort of symmetric thing and they're easily dropped.

From an architectural perspective “defaulting to Kubernetes standards and extending them when needed” seems like the most sensible thing to do, but I’ll leave the final decision to the UI team (cc @richard-cox).

moio (approved):

richard-cox:
> > > My first thought was it should probably be the operator that decides if its an exact match equality or a substring match. Something like: = is exact (stay closer to k8s), =~ is substring. !~ is not equal substring.
> > 
> > 
> > > Imo we should have only one way per matching strategy, I don't see a reason to have both double quotes and no quotes denote substring match. That should simplify the logic a bit too.
> > 
> > 
> > The requirements come form the dashboard, and they're the main consumer of this. So they'll need to weigh in on these operators. But I added the double-quotes as a sort of symmetric thing and they're easily dropped.
> 
> From an architectural perspective “defaulting to Kubernetes standards and extending them when needed” seems like the most sensible thing to do, but I’ll leave the final decision to the UI team (cc @richard-cox).

Dashboard side as long as we can execute equals or not equals operators and exact or partial matching of the value we're pretty free on the syntax. Implementation side the [usage](https://github.com/rancher/dashboard/blob/master/shell/plugins/steve/steve-pagination-utils.ts#L434-L436) is abstracted away from the api syntax, so it's an easy update either way

Changing the operator will make the type of match much more obvious. I had some minor concerns with tilde (this used to be a reserved character relating to the home directory and used when locating assets), though looks fine now https://datatracker.ietf.org/doc/html/rfc3986#section-2.3

For url brevity i would go with the following

- `=` exact
- `~` partial
- `!=` not exact
- `!~` not partial

ericpromislow:
> For url brevity i would go with the following
> 
>     = exact
>     ~ partial
>     != not exact
>     !~ not partial

is great.  `kubectl` supports `==` as a synonym for `=` so I'd go include that for the benefit of users used to using the `==` synonym in their k8s work.

Then we can drop quoting altogether, which would be great. Can I get a confirm from @moio to do this before I make further changes?

pkg/stores/sqlpartition/queryparser/selector.go
```diff
@@ -0,0 +1,966 @@
...
+	unaryOperators = []string{
+		string(selection.Exists), string(selection.DoesNotExist),
+	}
+	binaryOperators = []string{
+		string(selection.In), string(selection.NotIn),
+		string(selection.Equals), string(selection.DoubleEquals), string(selection.NotEquals),
+		string(selection.GreaterThan), string(selection.LessThan),
+	}
+	validRequirementOperators = append(binaryOperators, unaryOperators...)
+	labelSelectorRegex        = regexp.MustCompile(`^metadata.labels(?:\.\w[-a-zA-Z0-9_./]*|\[.*])$`)
```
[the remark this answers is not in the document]
ericpromislow: Makes sense -- I'd keep the `(?:...` part because you don't need to create a capture group.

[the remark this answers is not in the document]
moio: OK!

moio:
> Then we can drop quoting altogether, which would be great. Can I get a confirm from @moio to do this before I make further changes?

Sure, I confirm the plan makes sense to me as well! Thanks in advance

pkg/stores/sqlpartition/listprocessor/processor.go
```diff
@@ -53,6 +56,61 @@ type Cache interface {
 	ListByOptions(ctx context.Context, lo informer.ListOptions, partitions []partition.Partition, namespace string) (*unstructured.UnstructuredList, int, string, error)
 }
 
+func k8sOpToRancherOp(k8sOp selection.Operator) (informer.Op, bool, error) {
+	h := map[selection.Operator]informer.Op{
```
crobby, line 60: Is it possible to define this map somewhere outside of here to avoid redefining it on each call?

scripts/test.sh
```diff
@@ -11,3 +11,4 @@ version="1.$minor.x"
 
 export KUBEBUILDER_ASSETS=$(setup-envtest use -p path "$version")
 go test ./...
+# go test --race ./...
```
crobby, line 14: Why is this being added as a comment?

crobby (dismissed):

scripts/test.sh
ericpromislow: Because I wanted to hop between race and non-race mode and forgot to throw out that change :)

pkg/stores/sqlpartition/listprocessor/processor.go
```diff
@@ -54,6 +69,48 @@ type Cache interface {
...
+	return "", false, fmt.Errorf("unknown k8sOp: %s", k8sOp)
+}
+
+// Determine if the value field is surrounded by a pair of single- or double-quotes
+// This is a difference we implement from the kubernetes CLI: if the target value of a (not) equal
+// test is single-quoted, we use the full string. Otherwise we do a substring match
+// (which is implemented as 'SELECT ... some-field ... LIKE "%VALUE%" ...' in the query)
+//
+// The caller also needs to know if it should strip delimiting quotes, so this returns two bools
+func isQuotedStringTarget(values []string) (isQuoted bool, isSingleQuoted bool) {
```
moio, line 86: IIUC this is dead code now, can we drop it?

moio (approved):
LGTM

pkg/stores/sqlpartition/listprocessor/processor.go
ericpromislow: dropped

moio (approved):

crobby (approved):

PR #431, opened 2024-12-20
[v2.9] SQL cache backports

moio:
This backports #400, #430, #284, #433

#284 needed some minor adaptation, mostly due to the different mocking framework

I hope this helps


tomleb:
The branch here should be release/v0.3 according to https://github.com/rancher/steve/blob/main/VERSION.md. We'll be deleting the older branches later, after we fully transition.

moio:
Base changed

go.mod
```diff
@@ -34,11 +34,11 @@ require (
...
-	k8s.io/api v0.30.1
-	k8s.io/apiextensions-apiserver v0.30.1
-	k8s.io/apimachinery v0.30.1
-	k8s.io/apiserver v0.30.1
-	k8s.io/client-go v0.30.1
+	k8s.io/api v0.31.1
+	k8s.io/apiextensions-apiserver v0.31.0
+	k8s.io/apimachinery v0.31.1
+	k8s.io/apiserver v0.31.0
+	k8s.io/client-go v0.31.1
```
tomleb, line 41: I can see that rancher v2.9 is still on k8s v1.30. Steve must stay on the same version to make sure there aren't incompatibilities added later on. (eg: k8s v1.31 having features not present in v1.30, so the code in rancher/rancher wouldn't compile)

(We've hit those a couple of times and they're hard to spot in advance so by sticking to the right version it makes it a lot easier for us)

I suppose the bump to lasso is the culprit here forcing us to upgrade to v1.31. In that case we're going to have to either stop bumping lasso (but then that makes it hard to pull Vai changes) or pin the versions of kubernetes to the right version. (I'd go with the second option).

You can have a look at [webhooks' go.mod](https://github.com/rancher/webhook/blob/main/go.mod#L9-L35) for an example. That list isn't quite exhaustive yet, it's missing prometheus for example, but should be fine for now.

go.mod
```diff
@@ -19,11 +19,11 @@ require (
 	github.com/gorilla/websocket v1.5.1
 	github.com/pborman/uuid v1.2.1
 	github.com/pkg/errors v0.9.1
-	github.com/prometheus/client_golang v1.16.0
+	github.com/prometheus/client_golang v1.19.1
 	github.com/rancher/apiserver v0.0.0-20240708202538-39a6f2535146
 	github.com/rancher/dynamiclistener v0.6.0-rc2
 	github.com/rancher/kubernetes-provider-detector v0.1.5
-	github.com/rancher/lasso v0.0.0-20240923125127-ae858d002589
+	github.com/rancher/lasso v0.0.0-20241202185148-04649f379358
```
tomleb, line 26: (See the other comment I left for context)

This makes me wonder, is there anything forcing us to keep the sql cache in lasso itself? It feels like it would be much easier to move most of it to steve directly, that way no longer need to deal with bumping lasso and changes to Vai would only need to be done in two places (steve, + bump to Rancher) instead of three (lasso + steve + bump to rancher)

Thoughts? (cc: @ericpromislow since you've worked on this a lot too)

tomleb (commented):
One comment about kubernetes version bump and another about Vai in lasso.

go.mod
moio: > This makes me wonder, is there anything forcing us to keep the sql cache in lasso itself? It feels like it would be much easier to move most of it to steve directly, that way no longer need to deal with bumping lasso and changes to Vai would only need to be done in two places (steve, + bump to Rancher) instead of three (lasso + steve + bump to rancher)

I am not aware of any such constraints. It was a design choice to keep some of the complexity in lasso, under the premise it could have been more reusable there. FMPOV, until and unless we have an actual use case, logic can very well be kept all in one place.

go.mod
moio: addressed per your suggestion as 67e3385

go.mod
moio: Anyway we take this discussion, I believe it should not block this PR, right?

tomleb (commented):
One more change but otherwise looks good

go.mod
tomleb: Yes no need to block this.

tomleb (dismissed):

tomleb (approved):
LGTM

ericpromislow (approved):
Looks good. I got caught up on the target branch -- the sooner we can align all the release branch names the better.

PR #432, opened 2024-12-31
Partial extension API server store + control over printed columns

tomleb:
# Issue https://github.com/rancher/rancher/issues/48430

This PR addresses two issues with the current Extension API lib:

1. Forces all store operations even if they're not needed. For example, if a specific store only supports `create`, the discovery API and OpenAPI generated will still show that the store supports `get`, `list`, etc. This is misleading to the user, especially since the openapi spec is meant to be the documentation.
2. THere was no way to override the default column listing (either from kubectl get or from steve). The default shows the name and the age of the object.

# Optional operations

The upstream library auto-generates the verbs support by a `rest.Storage` based on the optional interfaces that it implements. Since previously the `rest.Storage` was the `delegate` which implemented all operations, it would generate all known verbs (`get`, `list`, etc).

To fix this, this PR removes most of the old abstraction of Delegate + Store. Instead, developers will need to create their store that implements the various optional interfaces exposed by the upstream library. We provide some utility functions to help implement some of those (eg: `CreateOrUpdate`, `ConvertToTable`) due to non-trivial logic, the rest are pretty simple to implement and will be left to the devs.

eg: Implementing a store with a single operation (`Create`) looks like the following:

```go
type partialStorage struct {
	gvk schema.GroupVersionKind
}

// New implements [rest.Storage]
func (t *partialStorage) New() runtime.Object {
	obj := &TestType{}
	obj.GetObjectKind().SetGroupVersionKind(t.gvk)
	return obj
}

// Destroy implements [rest.Storage]
func (t *partialStorage) Destroy() {
}

// GetSingularName implements [rest.SingularNameProvider]
func (t *partialStorage) GetSingularName() string {
	return "testtype"
}

// NamespaceScoped implements [rest.Scoper]
func (t *partialStorage) NamespaceScoped() bool {
	return false
}

// GroupVersionKind implements [rest.GroupVersionKindProvider]
func (t *partialStorage) GroupVersionKind(_ schema.GroupVersion) schema.GroupVersionKind {
	return t.gvk
}

func (s *partialStorage) Create(ctx context.Context, obj runtime.Object, createValidation rest.ValidateObjectFunc, options *metav1.CreateOptions) (runtime.Object, error) {
	if createValidation != nil {
		err := createValidation(ctx, obj)
		if err != nil {
			return obj, err
		}
	}

	return nil, nil
}
```

This is beneficial for the devs: the discovery API won't be misleading anymore, they won't need us to implement new verbs (eg: `DeleteCollection` wasn't in the old `Store` interface, `ConvertToTable` wasn't exposed either for developer to use, the old Store only supported cluster scoped stores). One issue is that we lose the whole `convertError` here that would convert non-apiserver (`k8s.io/apimachinery/pkg/api/errors`) errors to a fallback default one.

# Custom columns

Developers that want to define custom columns (eg: for the output of kubectl get) can now do so by implementing the [TableConvertor](https://github.com/kubernetes/apiserver/blob/v0.31.1/pkg/registry/rest/rest.go#L163-L165) interface.

To make it easier (because it has to work with both get and list), we provide a `ConvertToTable` utility that implements the complex part, leaving the developer only to implement a list of columns and a function to convert an object to a list of "cells".

```go
func (s *customColumnsStore) ConvertToTable(ctx context.Context, object runtime.Object, tableOptions runtime.Object) (*metav1.Table, error) {
	columns := []metav1.TableColumnDefinition{
		{
			Name: "Name",
			Type: "name",
		},
		{
			Name: "Foo",
			Type: "string",
		},
		{
			Name: "Bar",
			Type: "number",
		},
	}
	convertFn := func(obj *TestType) []string {
		return []string{
			"the name is " + obj.GetName(),
			"the foo value",
			"the bar value",
		}
	}

	return ConvertToTable(ctx, object, tableOptions, s.testStore.gvr.GroupResource(), columns, convertFn)
}

```


pkg/ext/store.go
```diff
@@ -29,6 +31,25 @@ type Context struct {
...
+type UpdateFunc[T runtime.Object] func(ctx Context, obj T, opts *metav1.UpdateOptions) (T, error)
+type GetFunc[T runtime.Object] func(ctx Context, name string, opts *metav1.GetOptions) (T, error)
+type ListFunc[TList runtime.Object] func(ctx Context, opts *metav1.ListOptions) (TList, error)
+type WatchFunc[T runtime.Object] func(ctx Context, opts *metav1.ListOptions) (<-chan WatchEvent[T], error)
+type DeleteFunc[T runtime.Object] func(ctx Context, name string, opts *metav1.DeleteOptions) error
+
+// ConvertFunc will convert a object to a list of cell in a metav1.Table (think kubectl get table output)
+type ConvertFunc[T runtime.Object] func(obj T) []string
+
+type Storage[T runtime.Object, TList runtime.Object] interface {
```
tomleb, line 44: Now we have a Store interface and a Storage interface. The old store interface forced us to implement all operations even if not necessary, so we'd like to move away from that.

I found the old store interface pretty convenient for tests because I'm building many different stores. I made it work with the new API with the new "StandardStore" struct, which takes in a Store. Maybe instead I should just remove the old Store interface just do:

`NewStandardStore(listFn, getFn, etc)` which would make it the same as the old store..

pkg/ext/apiserver_authentication_test.go
```diff
@@ -12,22 +13,32 @@ import (
...
 	"k8s.io/apiserver/pkg/authentication/authenticator"
 	"k8s.io/apiserver/pkg/authentication/user"
 	"k8s.io/apiserver/pkg/authorization/authorizer"
 	"k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/apiserver/pkg/registry/rest"
 	"k8s.io/apiserver/pkg/server/options"
 )
 
+var _ rest.Storage = (*authnTestStore)(nil)
+var _ rest.Lister = (*authnTestStore)(nil)
```
pmatseykanets, line 28: It's a matter of stylistic preference but I would do
```golang
var _ rest.Storage = &authnTestStore{}
var _ rest.Lister = &authnTestStore{}
```
Allocations should not be happening, if this is a concern.

pkg/ext/store.go
```diff
@@ -2,92 +2,260 @@ package ext
...
+		if err := fn(object); err != nil {
+			return nil, err
+		}
+	}
+
+	return table, nil
+}
+
+func cellStringToCellAny(cells []string) []any {
+	res := []any{}
```
pmatseykanets, line 87: nit: Maybe do this to avoid returning an empty slice instead of a nil slice when the input slice is nil.
```suggestion
	var res []any
```

pkg/ext/store.go
```diff
@@ -2,92 +2,260 @@ package ext
...
+	for _, cell := range cells {
+		res = append(res, cell)
+	}
+	return res
+}
+
+// CreateOrUpdate helps implement [rest.Updater] by handling most of the logic.
 //
-//	apierrors.NewNotFound(ctx.GroupVersionResource.GroupResource(), name)
+// createValidation is used to do some validation on the object before creating
```
pmatseykanets, line 96:
```suggestion
// createValidation is used to do validation on the object before creating
```

pkg/ext/store.go
```diff
@@ -2,92 +2,260 @@ package ext
...
-	Get(ctx Context, name string, opts *metav1.GetOptions) (T, error)
-	// List retrieves all resources matching the given ListOptions from the backing storage.
-	List(ctx Context, opts *metav1.ListOptions) (TList, error)
-	// Watch sends change events to a returned channel.
-	//
-	// The store is responsible for closing the channel.
-	Watch(ctx Context, opts *metav1.ListOptions) (<-chan WatchEvent[T], error)
-	// Delete deletes the resource of the given name from the backing storage.
-	Delete(ctx Context, name string, opts *metav1.DeleteOptions) error
+// updateValidation is used to do some validation on the object before updating it in the store.
```
pmatseykanets, line 102:
```suggestion
// updateValidation is used to do validation on the object before updating it in the store.
```

pkg/ext/store.go
```diff
@@ -2,92 +2,260 @@ package ext
...
+	forceAllowCreate bool,
+	options *metav1.UpdateOptions,
+	getFn func(ctx context.Context, name string, opts *metav1.GetOptions) (T, error),
+	createFn func(ctx context.Context, obj T, opts *metav1.CreateOptions) (T, error),
+	updateFn func(ctx context.Context, obj T, opts *metav1.UpdateOptions) (T, error),
+) (runtime.Object, bool, error) {
+	oldObj, err := getFn(ctx, name, &metav1.GetOptions{})
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return nil, false, err
```
pmatseykanets, line 119: Here and in all other error returns here - should we wrap and annotate errors? E.g.
```suggestion
			return nil, false, fmt.Errorf("error getting object %s: name, %w", )
```

pkg/ext/store.go
```diff
@@ -2,92 +2,260 @@ package ext
...
+	}()
+
+	return w, nil
+}
+
+func convertListOptions(options *metainternalversion.ListOptions, scheme *runtime.Scheme) (*metav1.ListOptions, error) {
+	var out metav1.ListOptions
+	err := scheme.Convert(options, &out, nil)
+	if err != nil {
+		return nil, fmt.Errorf("convert list options: %w", err)
```
pmatseykanets, line 249:
```suggestion
		return nil, fmt.Errorf("error converting list options: %w", err)
```

pkg/ext/store_test.go
```diff
@@ -0,0 +1,56 @@
...
+				},
+				Message: fmt.Sprintf("Internal error occurred: %v", assert.AnError),
+			}},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			output := convertError(tt.input)
+			assert.Equal(t, tt.output, output)
```
pmatseykanets, line 52:
```suggestion
			assert.Equal(t, tt.output, convertError(tt.input))
```

pkg/ext/store_test.go
```diff
@@ -0,0 +1,56 @@
...
+		})
+	}
+
```
pmatseykanets, line 55: Extra EOL
```suggestion
```

pkg/ext/fixtures_test.go
```diff
@@ -162,33 +170,196 @@ func (t *TestTypeOther) DeepCopyObject() runtime.Object {
...
+}
+
+// Update implements [rest.Updater]
+func (t *testStore[T, TList]) Update(ctx context.Context, name string, objInfo rest.UpdatedObjectInfo, createValidation rest.ValidateObjectFunc, updateValidation rest.ValidateObjectUpdateFunc, forceAllowCreate bool, options *metav1.UpdateOptions) (runtime.Object, bool, error) {
+	t.lock.Lock()
+	defer t.lock.Unlock()
+	return CreateOrUpdate(ctx, name, objInfo, createValidation, updateValidation, forceAllowCreate, options, t.get, t.create, t.update)
+}
+
+func (t *testStore[T, TList]) get(ctx context.Context, name string, opts *metav1.GetOptions) (*TestType, error) {
```
pmatseykanets, line 270:
```suggestion
func (t *testStore[T, TList]) get(_ context.Context, name string, _ *metav1.GetOptions) (*TestType, error) {
```

pkg/ext/fixtures_test.go
```diff
@@ -162,33 +170,196 @@ func (t *TestTypeOther) DeepCopyObject() runtime.Object {
...
+	obj, found := t.items[name]
+	if !found {
+		return nil, apierrors.NewNotFound(t.gvr.GroupResource(), name)
+	}
+	return obj, nil
+}
+
+func (t *testStore[T, TList]) create(ctx context.Context, obj *TestType, opts *metav1.CreateOptions) (*TestType, error) {
```
pmatseykanets, line 278:
```suggestion
func (t *testStore[T, TList]) create(_ context.Context, obj *TestType, _ *metav1.CreateOptions) (*TestType, error) {
```

pkg/ext/fixtures_test.go
```diff
@@ -162,33 +170,196 @@ func (t *TestTypeOther) DeepCopyObject() runtime.Object {
...
+	}
+	t.items[obj.Name] = obj
+	t.events <- WatchEvent[*TestType]{
+		Event:  watch.Added,
+		Object: obj,
+	}
+	return obj, nil
+}
+
+func (t *testStore[T, TList]) update(ctx context.Context, obj *TestType, opts *metav1.UpdateOptions) (*TestType, error) {
```
pmatseykanets, line 290:
```suggestion
func (t *testStore[T, TList]) update(_ context.Context, obj *TestType, _ *metav1.UpdateOptions) (*TestType, error) {
```

pmatseykanets (commented):
Looks great!

pkg/ext/store.go
tomleb: Any errors returned by the various store operations (`Create`, `Update`, `Delete`, `List`, etc) **must** implement [apierrors.APIStatus](https://github.com/kubernetes/apimachinery/blob/v0.31.1/pkg/api/errors/errors.go#L39-L43). If we don't then the library that's calling our store's methods will log an error because it won't recognize the error and the clients (kubectl) won't get the pretty printed error either.

If we just wrap like suggested here, we'd lose the apierrors.APIStatus that `getFn` returns, or createFn / updateFn returns. I could do a better wrapping though if getFn returned the expected APIStatus. Then we can add context to the error message without changing the APIStatus error.

Thoughts?

pmatseykanets: > Any errors returned by the various store operations (Create, Update, Delete, List, etc) must implement apierrors.APIStatus.

Does `create|updateValidation` **always** return `apierrors.APIStatus`?
Also [L152-L153](https://github.com/tomleb/rancher-steve/blob/249d3b8697f98dc6b38d2353357d69d309642e31/pkg/ext/store.go#L152-L153) doesn't conform with the **must** part.

> I could do a better wrapping though if getFn returned the expected APIStatus. Then we can add context to the error message without changing the APIStatus error.

👍

pkg/ext/store.go
tomleb: I usually prefer not to prefix every error with `error`, it's usually clear that we're getting an error.

eg: I prefer to avoid this

```
error converting list options: error this didn't work: error foo failed
```

pkg/ext/store_test.go
tomleb: Fixed

pkg/ext/fixtures_test.go
tomleb: Fixed.

pkg/ext/fixtures_test.go
tomleb: Fixed.

pkg/ext/fixtures_test.go
tomleb: Fixed

pkg/ext/store_test.go
tomleb: Sure, changed

pkg/ext/store.go
tomleb: > Does create|updateValidation always return apierrors.APIStatus?

I don't think so. We can add our own validation on top of the default provided so it's probably best to to the conversion as a fallback. Added that.

> Also [L152-L153](https://github.com/tomleb/rancher-steve/blob/249d3b8697f98dc6b38d2353357d69d309642e31/pkg/ext/store.go#L152-L153) doesn't conform with the must part.

Yeah, that should be fixed now. I'm doing convertError everywhere except for the getFn / createFn / updateFn calls. I expect them to return `apierrors.APIStatus` errors.

pkg/ext/store.go
tomleb: Reworded those comments so you can have a look at them again.

pkg/ext/store.go
tomleb: Reworded those comments so you can have a look at them again.

pkg/ext/store.go
tomleb: Sure

pkg/ext/apiserver_authentication_test.go
tomleb: I'll leave as is since it's mostly stylistic. The codebase already uses (*<thing>)(nil) in a few places so I don't see a reason to change it.

tomleb:
@andreas-kupries Thoughts on the watcher utility? I believe you're implementing watch for your PR. 

Basically, the `Watch()` method must return a watch interface that doesn't have any default implementing (afaik). So if we don't provide one for you, you'll have to implement it yourself.

pkg/ext/fixtures_test.go
```diff
@@ -162,33 +170,268 @@ func (t *TestTypeOther) DeepCopyObject() runtime.Object {
...
+		ch: make(chan watch.Event),
+	}
+	go func() {
+		// Not much point continuing the watch if the store stopped its watch.
+		// Double stopping here is fine.
+		defer w.Stop()
+
+		// Closing eventCh is the responsibility of the store.Watch method
+		// to avoid the store panicking while trying to send to a close channel
+		eventCh, err := watchFn(ctx, options)
```
andreas-kupries, line 418: In this test store the `eventCh` does not look to be closed at all ?

If I understand things correctly here the `eventCh` is the channel of backend events (i.e. for tokens the events on the backing secrets). My caller is given a watch structure `w` fulfilling the `watch.Interface` which returns a channel `C` to the caller to pull token events from. The newly-made go routine reads from `eventsCh`, converts and passes to `C` for handling upstream. 

Now looking at terminating a watch, I see two cases, from the top on down (1), and from bottom up (2).
Only for (1) do I have an idea how to handle `eventCh`.

1. When the caller stops watching I assume that it invokes `w.Stop()`. This marks the structure as closed, and closes `C` as well. The go routine notices the closed state of `C` on the next attempt to add an event and terminates it self. As part of that it calls `Stop()` again, which notices the closed state and does nothing. 
2. When the underlying `eventCh` somehow closes (x) the go routine terminates, runs `Stop()` which marks the structure as closed and closes `C`.  The closed `C` is then noted up stream in their (range C), this terminates event processing in the caller as well.

For (1) we can close `eventCh` in `Stop()`, same as we do for `C`.
For (2) `eventCh` is already closed, doing it in `Stop()` is likely a bad thing.

(Ad x) Is that possible ?

andreas-kupries (commented):
Looks ok, although I have questions. The big one is about `Watch` handling.

pkg/ext/apiserver.go
```diff
@@ -201,34 +200,43 @@ func (s *ExtensionAPIServer) ServeHTTP(w http.ResponseWriter, req *http.Request)
...
+// watch verbs.
 //
-//	gvk := schema.GroupVersionKind{
-//		Group: "ext.cattle.io",
-//		Version: "v1",
-//		Kind: "Token",
-//	}
-//	InstallStore(s, &Token{}, &TokenList{}, "tokens", "token", gvk, store)
+// A store MUST implement the following interfaces: [rest.Storage], [rest.Scoper], [rest.GroupVersionKindProvider]
+// and [rest.SingularNameProvider].
```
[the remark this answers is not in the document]
tomleb: Yes. In this case, the file imports `k8s.io/apiserver/pkg/registry/rest` and `rest` is the package name, so Go will know `rest.Storage` is from that package. This makes it work with your IDE (eg: go to definition should work on that comment), and it would also be rendered as a link in the go docs. (Though it can be a bit confusing without seeing the imports)

tomleb:
> > @andreas-kupries Thoughts on the watcher utility?
> 
> I did not see any ? (Except maybe some of the code for the test store ?)

My bad, that was not a very clear question. I have unexported the watch utility so you'd have to implement your own. Is that something you'd rather do or would you rather it be in this steve. We can't know how future stores will need to implement watch, so we can't really implement a watch utility that will work for most cases. If you were fine with the current watch then I can export it and you'll be able to use it.

> > I believe you're implementing watch for your PR.
> 
> Yes.
> 
> > Basically, the `Watch()` method must return a watch interface that doesn't have any default implementing (afaik). So if we don't provide one for you, you'll have to implement it yourself.
> 
> I see no issue with implementing the structure and interface myself. Then, when after stores also implement `Watch` we can see how much of that can be moved into the utilities.
> 
> Big question I have (see my review) is how to handle the `eventCh` exactly, to close/terminate it correctly.

 :+1:

pkg/ext/fixtures_test.go
tomleb: > If I understand things correctly here the eventCh is the channel of backend events (i.e. for tokens the events on the backing secrets)

iirc, the channel was for the tokens itself. So the store was meant to implement the secrets -> tokens conversion and send that in the channel. We were abstracting away the upstream `watch.Interface` interface.

> Now looking at terminating a watch, I see two cases, from the top on down (1), and from bottom up (2).

So I went back and looked at the [watch.Interface](https://github.com/kubernetes/apimachinery/blob/v0.31.1/pkg/watch/watch.go#L29) again and I see the following:

```go
// Stop tells the producer that the consumer is done watching, so the
// producer should stop sending events and close the result channel. ...

// Only the consumer should call Stop(), not the producer. If the producer
// errors and needs to stop the watch prematurely, it should instead send
// an error event and close the result channel.
```

In our case, the store (`rest.Watcher`) is the producer, so it shouldn't call `Stop()` at all, ever.

So I'm thinking either:
1. We implement some `watch.Interface` that does a watch on another resource, starts a goroutine and calls a conversion function (eg: secret -> tokens). And that becomes your watcher. This can be in the library assuming it might get used in other stores.
2. We let you implement this logic and if we see that it becomes used by others then we put it back in the library.

I'd favor (2) here for the same reason you mention in another comment ("we can see how much of that can be moved into the utilities")

If we ignore the current utilities implementation (which is wrong, I'll fix that) are you better placed now to implement that interface? You'll still need a channel, but that channel will only ever be closed in the `Stop()` method. Then your backend watch (secrets) should also be called in the `Stop()` method.

Something like:

```go
type TokenWatcher struct {
	secretWatch watch.Interface
	ch          chan watch.Event
	// TODO: Need mutex / atomic.Bool
	closed bool
}

func (w *TokenWatcher) Stop() {
	w.closed = true
	w.secretWatch.Stop()
	close(w.ch)
}

func (w *TokenWatcher) ResultChan() <-chan watch.Event {
	return w.ch
}

func (t *tokenStore) Watch(ctx context.Context, internaloptions *metainternalversion.ListOptions) (watch.Interface, error) {
	options, err := ConvertListOptions(internaloptions)
	if err != nil {
		return nil, err
	}

	secretWatch, err := secretClient.Watch(ctx, options)
	if err != nil {
		return nil, err
	}

	watcher := &TokenWatcher{
		secretWatch: secretWatch,
		ch:          make(chan watch.Event),
	}
	go func() {
		for secretEvent := range secretWatch.ResultChan() {
			tokenEvent := transform(secretEvent)
			// TODO: Need mutex
			if !watcher.closed {
				watcher.ch <- tokenEvent
			}
		}
	}()
	return watcher, nil
}
```

pkg/ext/apiserver.go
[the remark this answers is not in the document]
andreas-kupries: Ok. For me the important part is that I now know where to find the interfaces, and their methods, for when it comes to rewriting the token store.

pkg/ext/fixtures_test.go
andreas-kupries: Agreed on 1 versus 2, 2 prefered.

>  are you better placed now to implement that interface? 

Yes. The info from the `watch.Interface` itself clears things up nicely. It means that what the testStore` watch is doing is very much wrong regarding calls of Stop.

The example needs a fix, I believe, in the form of 
```
go func() {
	for secretEvent := range secretWatch.ResultChan() {
		tokenEvent := transform(secretEvent)
		// TODO: Need mutex
		if !watcher.closed {
			watcher.ch <- tokenEvent
		}
	}
++++   secretWatch.Stop() // pass stop signal from above to the watcher below.
}()
```

The Stop() from above arrives in the go routine, triggers its terminating, and in turn terminates the lower watcher.

And the `TODO mutex` can be hidden in the methods of `TokenWatcher`, i.e., `Stop()` and an `addEvent` as the test store had.

pkg/ext/apiserver_authorization_test.go
```diff
@@ -67,6 +75,54 @@ func (t *authzTestStore) List(ctx Context, opts *metav1.ListOptions) (*TestTypeL
...
+		err := createValidation(ctx, obj)
+		if err != nil {
+			return obj, err
+		}
+	}
+
+	objT, ok := obj.(*TestType)
+	if !ok {
+		var zeroT *TestType
+		return nil, fmt.Errorf("expected %T but got %T", zeroT, obj)
```
pmatseykanets, line 105: Should probably `convertError()` here.

pmatseykanets (dismissed):

pkg/ext/apiserver_authorization_test.go
tomleb: Fixed

pkg/ext/fixtures_test.go
```diff
@@ -162,33 +170,268 @@ func (t *TestTypeOther) DeepCopyObject() runtime.Object {
...
+		err := createValidation(ctx, obj)
+		if err != nil {
+			return obj, err
+		}
+	}
+
+	objT, ok := obj.(*TestType)
+	if !ok {
+		var zeroT T
+		return nil, fmt.Errorf("expected %T but got %T", zeroT, obj)
```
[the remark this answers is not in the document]
tomleb: Fixed

tomleb:
@andreas-kupries Fixed the watcher implementation and documented request.UserFrom to get the user info (which was done previously by the `Context` object

Ready for a re-review.

pmatseykanets (approved):
👍

andreas-kupries (approved):

crobby (approved):
I think this looks great.  Tom gave me a nice talkthrough and I really like the tests.  They seem very thorough and helpful.

PR #434, opened 2025-01-08
Imperative api pls

joshmeranda:
https://github.com/rancher/rancher/issues/47540

Replaces https://github.com/rancher/steve/pull/316

> Implements the default authentication.
>
> This was original part of https://github.com/rancher/steve/pull/287 but was split up to make it easier to review.


tomleb:
@bigkevmcd adding you since you had reviewed the other one previously.

tomleb (approved):
LGTM

One thing I'd like to have as test in the future is running the API server with the APIService configured so that we can verify a few things when going through the main kubeapiserver (eg: ensuring user, groups and extras headers are properly propagated). This would require running a pod as an integration test (or potentially we can do that when RDP work is done).

pkg/ext/apiserver_suite_test.go
```diff
@@ -43,6 +219,7 @@ func (s *ExtensionAPIServerSuite) TearDownSuite() {
 	s.cancel()
 	err := s.testEnv.Stop()
 	s.Require().NoError(err)
+	os.RemoveAll(s.certTempPath)
```
bigkevmcd, line 222: If this fails to stop, will the `certTempPath` be left around?

Maybe `s.T().TempDir()`?

pkg/ext/apiserver_test.go
```diff
@@ -129,13 +129,12 @@ func TestStore(t *testing.T) {
...
 
 	store := newMapStore()
-	extensionAPIServer, cleanup, err := setupExtensionAPIServer(t, scheme, &TestType{}, &TestTypeList{}, store, func(opts *ExtensionAPIServerOptions) {
+	extensionAPIServer, err := setupExtensionAPIServer(t, scheme, &TestType{}, &TestTypeList{}, store, func(opts *ExtensionAPIServerOptions) {
 		opts.Listener = ln
 		opts.Authorizer = authorizer.AuthorizerFunc(authzAllowAll)
 		opts.Authenticator = authenticator.RequestFunc(authAsAdmin)
 	}, nil)
 	require.NoError(t, err)
-	defer cleanup()
```
bigkevmcd, line 138: 👍🏻

pkg/ext/apiserver_test.go
```diff
@@ -711,11 +709,11 @@ func setupExtensionAPIServer[
 	err = extensionAPIServer.Run(ctx)
```
bigkevmcd, line 709: `setupExtensionAPIServer` both returns an error in some cases, **and** fails the test in other cases.

I suspect it doesn't need to return the error, it could just `require.NoError()`  and _maybe_ `t.Helper()` is appropriate in this case too?

pkg/ext/apiserver_suite_test.go
```diff
@@ -19,11 +163,43 @@ type ExtensionAPIServerSuite struct {
...
+	// Configures the aggregation layer according to
+	// https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer/#enable-kubernetes-apiserver-flags
 	apiServer := &envtest.APIServer{}
+	apiServer.Configure().Append("requestheader-allowed-names", "system:auth-proxy")
+	apiServer.Configure().Append("requestheader-extra-headers-prefix", "X-Remote-Extra-")
+	apiServer.Configure().Append("requestheader-group-headers", "X-Remote-Group")
+	apiServer.Configure().Append("requestheader-username-headers", "X-Remote-User")
+	apiServer.Configure().Append("requestheader-client-ca-file", caFilepath)
+	apiServer.Configure().Append("proxy-client-cert-file", certFilepath)
+	apiServer.Configure().Append("proxy-client-key-file", keyFilepath)
```
bigkevmcd, line 202: 👍🏻

pkg/ext/apiserver_test.go
joshmeranda: Sounds fair enough, we never really do anything with the errors from `setupExtensionAPISever` any way so we can just fail if we encounter one. And yes `t.Helper()` will be useful to us!

tomleb (approved):
LGTM

pkg/ext/apiserver_authentication.go
```diff
@@ -0,0 +1,257 @@
...
+// [dynamiccertificates.CAContentProvider].
+type UnionAuthenticator struct {
+	unionAuthenticator     authenticator.Request
+	unionCAContentProvider dynamiccertificates.CAContentProvider
+}
+
+// NewUnionAuthenticator creates a [UnionAuthenticator].
+//
+// The authenticators will be tried one by one, in the order they are given, until
+// one succeed or all fails.
```
crobby, line 38: "one succeeds or all fail" ??

crobby (approved):
Looks good.  I like the examples included in the comments.

PR #447, opened 2025-01-13
 SQLite backed cache: Support sorting mgmt clusters on value in a specific condition

ericpromislow:
Supports the UI to display managed clusters in a "saner" way.

Sample URL:

```
$ curl -sk https://localhost:5111/v1/management.cattle.io.clusters?sort=-spec.internal,-status.connected,spec.displayName' | jq '.data[] | [.metadata.name, .spec.internal, .status.connected, .spec.displayName]' 
[ "local", true, true,   "local" ]
[ "c-m-qc5fl797", false, true, "eptest-20250113a" ]
[ "c-m-kkjfv6r2",  false,  true,  "eptest-20250113e" ]
[ "c-m-7rp8c77v",  false,   false,  "eptest-20250113b2" ]
```

This sort order puts the internal clusters first, then ready (secondarily sorting by the cluster's friendly name), then the 
non-ready ones.

Note putting a hyphen before boolean fields does a reverse (or DESC) sort, so the true ones appear before the false ones.

Related to [#48092](https://github.com/rancher/rancher/issues/48092)

Depends on: 
[#122](https://github.com/rancher/lasso/pull/122) - Only because lasso is doing some sql code-generation that steve should be doing.


pkg/stores/queryhelper/safesplit.go
```diff
@@ -0,0 +1,16 @@
...
+// SafeSplit breaks a regular "a.b.c" path to the string slice ["a", "b", "c"]
+// but if the final part is in square brakets, like "metadata.labels[cattleprod.io/moo]",
+// it returns ["metadata", "labels", "cattleprod.io/moo"
+func SafeSplit(fieldPath string) []string {
+	squareBracketLocation := strings.Index(fieldPath, "[")
+	if squareBracketLocation == -1 {
+		return strings.Split(fieldPath, ".")
+	}
+	s := strings.Split(fieldPath[0:squareBracketLocation], ".")
+	s = append(s, fieldPath[squareBracketLocation+1:len(fieldPath)-1])
```
moio, line 14: Does the case where `string = "a.b[c"` need any special treatment? (no ending square bracket)

How about `a[[`?

README.md
```diff
@@ -187,9 +187,9 @@ The list can be negated to exclude results:
 
 #### `sort`
```
moio, line 188: You might want to add a paragraph about the chance in the syntax of accessing properties with dot in their name, as the square brackets syntax is new.

pkg/stores/partition/listprocessor/processor.go
```diff
@@ -99,27 +101,28 @@ const (
...
-	field := ""
-	if s.primaryOrder == DESC {
-		field = "-" + field
-	}
-	field += strings.Join(s.primaryField, ".")
-	if len(s.secondaryField) > 0 {
-		field += ","
-		if s.secondaryOrder == DESC {
-			field += "-"
+	dottableRegex := regexp.MustCompile(`[^a-zA-Z0-9_]`)
```
moio, line 110: What is the reason to use particular regex was chosen instead of checking for the presence of a dot, like in SafeSplit?

pkg/stores/partition/listprocessor/processor.go
```diff
@@ -189,24 +192,23 @@ func ParseQuery(apiOp *types.APIRequest) *ListOptions {
...
+					sortOrder = DESC
+					field = field[1:]
+				}
 			}
-			if secondaryField != "" {
-				sortOpts.secondaryField = strings.Split(secondaryField, ".")
+			if len(field) == 0 {
+				// Current semantics: if we have a sort query like
+				// sort=,metadata.someOtherField
+				// we *DON'T sort
```
moio, line 207: I suggest we take the chance to fix this - either raising an error or producing reliably the same sorting would be OK FMPOV.

moio (commented):
Thanks @ericpromislow, this looks good to undraft to me.

As a side note, I see there are three concerns being addressed in this PR:
 - adding support for an arbitrary number of sort conditions
 - adding support for fields containing a dot
 - adding a transformation to the v3 cluster object in order to contain a virtual `.connected` field in its status

I understand the reasoning behind keeping all of these in one PR, as the end-user concern/ask is actually unique. OTOH, an argument could be made to to split the change into three PRs addressing one concern each, especially because the first two affect non-Vai use cases too.

I will leave it the Frameworks team to make that decision.

From my perspective, most things look good - apart from suggestions there is only a couple of questions that are hopefully quick to settle.

Thanks!

go.mod
```diff
@@ -8,6 +8,7 @@ replace (
 	github.com/crewjam/saml => github.com/rancher/saml v0.2.0
 	github.com/knative/pkg => github.com/rancher/pkg v0.0.0-20181214184433-b04c0947ad2f
 	github.com/matryer/moq => github.com/rancher/moq v0.0.0-20190404221404-ee5226d43009
+	github.com/rancher/lasso => ../lasso
```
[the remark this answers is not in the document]
ericpromislow: of course

pkg/stores/queryhelper/safesplit.go
ericpromislow: Those will just result in failures to match data. I don't see a need to do error-checking here.

pkg/stores/partition/listprocessor/processor.go
ericpromislow: I had a choice between either specifying a string that contained something other than an identifier character (alnum or underscore), or specifying a set of non-ident characters (which for k8s labels are `./-`). I figured the alnum characters would be more straightforward. But the name is wrong -- if the term contains a character that is not in that character set, it is NOT dottable. Looking at the unit tests now ...

pkg/stores/partition/listprocessor/processor.go
ericpromislow: The good news is when I made this change a unit test failed.

go.mod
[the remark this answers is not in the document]
ericpromislow: gone

pkg/stores/partition/listprocessor/processor.go
```diff
@@ -350,24 +352,29 @@ func matchesAll(obj map[string]interface{}, filters []OrFilter) bool {
...
-				return leftSecond < rightSecond
+		leftNode := list[i].Object
+		rightNode := list[j].Object
+		for i, field := range s.Fields {
+			//itfs := make([]interface{}, len(field))
+			//for j, f := range field {
+			//	itfs[j] = f
+			//}
+			//leftValue := convert.ToString(data.GetValueN(leftNode, itfs...))
+			//rightValue := convert.ToString(data.GetValueN(rightNode, itfs...))
```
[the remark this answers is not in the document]
ericpromislow: This code was there because my IDE didn't like the `field...` arg in `data.GetValueN()`, but it runs fine, so I can drop that commented-out code.

moio (dismissed):
Looks good from my perspective once the README fully reflects the syntax change introduced.

I guess it can be merged as soon as @tomleb also approves.

Thanks for this contribution!

README.md
ericpromislow: Done.

moio (approved):

moio (approved):

PR #450, opened 2025-01-14
[main&2.10.3] Add schema links and resource methods for resource verb patch

crobby:
**NOTE:  overriding the apiserver version until https://github.com/rancher/apiserver/pull/128 is merged/tagged**

Issue:  https://github.com/rancher/rancher/issues/40712

Previously, we have not factored-in nor included the patch resource permission in our object links or schema for resource methods.

With this change, if a user has patch permission on an object, it will be reflected in both the links for an object as well as the resource methods on the schema.

-----
Testing:
Manual testing done
Unit tests added

--------

Helpful testing information:

To see the schema for something like apps.deployments, you can go to:  <rancher url>/v1/schemas/apps.deployments
In the resultant json, if the user has patch permissions, you should see something like...
```
"resourceMethods": [ 4 items
"GET",
"DELETE",
"PUT",
"PATCH"
],
```
included in the yaml blob.

Likewise, if you look at the api view for an object, you will see the patch link defined when a user has patch permissions on that object.
`{
  "patch": "https://localhost:9443/apis/apps/v1/namespaces/default/deployments/nginx-deployment",
  "self": "https://localhost:9443/v1/apps.deployments/default/nginx-deployment",
  "update": "https://localhost:9443/apis/apps/v1/namespaces/default/deployments/nginx-deployment",
  "view": "https://localhost:9443/apis/apps/v1/namespaces/default/deployments/nginx-deployment"
}
`

### DIFF
--- a/pkg/resources/common/formatter.go
+++ b/pkg/resources/common/formatter.go
@@ -98,6 +98,7 @@ func formatter(summarycache *summarycache.SummaryCache, asl accesscontrol.Access
 		}
 		hasUpdate := accessSet.Grants("update", gvr.GroupResource(), resource.APIObject.Namespace(), resource.APIObject.Name())
 		hasDelete := accessSet.Grants("delete", gvr.GroupResource(), resource.APIObject.Namespace(), resource.APIObject.Name())
+		hasPatch := accessSet.Grants("patch", gvr.GroupResource(), resource.APIObject.Namespace(), resource.APIObject.Name())
 
 		selfLink := selfLink(gvr, meta)
 
@@ -117,6 +118,15 @@ func formatter(summarycache *summarycache.SummaryCache, asl accesscontrol.Access
 			}
 		} else {
 			delete(resource.Links, "remove")
+		}
+		if hasPatch {
+			if attributes.DisallowMethods(resource.Schema)[http.MethodPut] {
+				resource.Links["patch"] = "blocked"
+			} else {
+				resource.Links["patch"] = u
+			}
+		} else {
+			delete(resource.Links, "patch")
 		}
 
 		if unstr, ok := resource.APIObject.Object.(*unstructured.Unstructured); ok {
--- a/pkg/resources/common/formatter.go
+++ b/pkg/resources/common/formatter.go
@@ -120,10 +120,8 @@ func formatter(summarycache *summarycache.SummaryCache, asl accesscontrol.Access
 			delete(resource.Links, "remove")
 		}
 		if hasPatch {
-			if attributes.DisallowMethods(resource.Schema)[http.MethodPut] {
+			if attributes.DisallowMethods(resource.Schema)[http.MethodPatch] {
 				resource.Links["patch"] = "blocked"
-			} else {
-				resource.Links["patch"] = u
 			}
 		} else {
 			delete(resource.Links, "patch")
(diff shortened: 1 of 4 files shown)


pkg/resources/common/formatter.go
```diff
@@ -122,6 +123,15 @@ func formatter(summarycache *summarycache.SummaryCache, asl accesscontrol.Access
 		} else {
 			delete(resource.Links, "remove")
 		}
+		if hasPatch {
+			if attributes.DisallowMethods(resource.Schema)[http.MethodPatch] {
+				resource.Links["patch"] = "blocked"
+			} else {
```
ericpromislow, line 129: Please delete the block
```
} else {
  resource.Links["patch"] = u
  ```
  
  This code caused the slack-only issue fixed by https://github.com/rancher/steve/pull/451

File: pkg/resources/common/formatter_test.go [hunk 2]
```diff
@@ -935,6 +936,79 @@ func Test_formatterLinks(t *testing.T) {
...
+				Object: &v1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "example-deployment",
+						Namespace: "example-ns",
+					},
+				},
+			},
+			currentLinks: map[string]string{
+				"default": "defaultVal",
+				"view":    "/apis/apps/v1/namespaces/example-ns/deployments/example-deployment",
```
ericpromislow, line 966: `"view"` should be `"patch"` , assuming you delete the two lines from `formatter.go`

ericpromislow (changes requested):
This needs to be brought into alignment with main after a recent fix to steve.

crobby:
> This needs to be brought into alignment with main after a recent fix to steve.

It looks like apiserver sets the links for self, update, and remove, but not for patch.  I think it's probably something that I'll need to add there first, and then the patch link should be there so I can make this look like the other verbs here in Steve

pkg/resources/common/formatter.go
crobby: Thanks.  I've made a change to apiserver to return the patch link.  This PR is using my forked copy of that for now, but it seems to work correctly.

pkg/resources/common/formatter_test.go
crobby: test updated to reflect the new world order

crobby:
Ok, I've gone made a PR for apiserver to include patch links.
After gluing them all together with Rancher, I now see....
Screenshot from 2025-01-16 11-09-13
as expected.

go.mod
```diff
@@ -8,6 +8,7 @@ replace (
 	github.com/crewjam/saml => github.com/rancher/saml v0.2.0
 	github.com/knative/pkg => github.com/rancher/pkg v0.0.0-20181214184433-b04c0947ad2f
 	github.com/matryer/moq => github.com/rancher/moq v0.0.0-20190404221404-ee5226d43009
+	github.com/rancher/apiserver => github.com/crobby/apiserver v0.5.2-0.20250116110826-e76f13e6c0bd
```
ericpromislow, line 11: I typically leave PRs in draft mode when I'm referencing local repos until they get merged in.

ericpromislow (dismissed):
Everything but the references to crobby/apiserver looks good.

go.mod
crobby: fair enough

joshmeranda (approved):

PR #308, opened 2024-10-28
Finish the AccessSet changes to the main formatter.

ericpromislow:
Related to rancher/rancher#39129. 

The original PR, steve/pull/158 drifted too far from changes in main, so it's easier to create a new PR:

* Bring in and update DefaultSchemaTemplatesForStore
* Move from based on `master` to `main`
* Update k8s version

This includes and  completes changes for [steve/158](https://github.com/rancher/steve/pull/158/commits)

PR #451, opened 2025-01-15
Fix overriding update and remove links with the wrong link

tomleb:
Before https://github.com/rancher/steve/pull/308 we were not overriding the update / remove links, so they were pointing to the right path (`/v1/...`). The new code was now overriding those links but to the wrong path (`/api/...` or `/apis/...`) which points to the native k8s resource. This caused issue in the UI.

The fix is to not override the links that already exist, we either block or delete them.

### DIFF
--- a/pkg/resources/common/formatter.go
+++ b/pkg/resources/common/formatter.go
@@ -107,8 +107,6 @@ func formatter(summarycache *summarycache.SummaryCache, asl accesscontrol.Access
 		if hasUpdate {
 			if attributes.DisallowMethods(resource.Schema)[http.MethodPut] {
 				resource.Links["update"] = "blocked"
-			} else {
-				resource.Links["update"] = u
 			}
 		} else {
 			delete(resource.Links, "update")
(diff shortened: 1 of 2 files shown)


tomleb:
@ericpromislow Was there a reason to override those links? I see they are [set conditionally in apiserver](https://github.com/rancher/apiserver/blob/074b1305ff087b2878a1f01bb06f9f08493a9912/pkg/writer/encoding.go#L95-L104). Is there a case where they wouldn't be allowed in apiserver (so no links) but allowed in steve?

bigkevmcd (approved):
This is the same as I had working, so works for me!

ericpromislow:
> @ericpromislow Was there a reason to override those links? I see they are [set conditionally in apiserver](https://github.com/rancher/apiserver/blob/074b1305ff087b2878a1f01bb06f9f08493a9912/pkg/writer/encoding.go#L95-L104). Is there a case where they wouldn't be allowed in apiserver (so no links) but allowed in steve -> would result in no links?

So Michael B walked me through his solution for a problem which turned into steve PR 308, which just incorporated the AccessSetLookup, but I don't have enough context to tell you what the reason was, nor what's going on in the apiserver.

ericpromislow (approved):
Looks good, makes sense.

PR #455, opened 2025-01-20
Fix sort field parsing

tomleb:
Without this fix, queries with sort params will return the following error:

```
{"type":"error","links":{},"code":"ServerError","message":"sort fields length 1 != sort orders length 0","status":500,"type":"error"}
```


ericpromislow (approved):
This PR will be obsoleted by PR 447 but it's fine for now

moio (approved):

PR #458, opened 2025-01-23
 support unwatchables in vai

ericpromislow:
Related to: [#46768](https://github.com/rancher/rancher/issues/46768) :

- Create a synthetic watcher for watching for changes in the specified resource
- When changes are detected, write `watch.Event` objects which get processed by the same reflector that processes changes in watchable resources
- Includes a single unit test to exercise the core functionality of the synthetic watcher.

### DIFF
--- a/pkg/sqlcache/informer/informer.go
+++ b/pkg/sqlcache/informer/informer.go
@@ -35,15 +35,21 @@ var newInformer = cache.NewSharedIndexInformer
 
 // NewInformer returns a new SQLite-backed Informer for the type specified by schema in unstructured.Unstructured form
 // using the specified client
-func NewInformer(ctx context.Context, client dynamic.ResourceInterface, fields [][]string, transform cache.TransformFunc, gvk schema.GroupVersionKind, db db.Client, shouldEncrypt bool, namespaced bool) (*Informer, error) {
+func NewInformer(ctx context.Context, client dynamic.ResourceInterface, fields [][]string, transform cache.TransformFunc, gvk schema.GroupVersionKind, db db.Client, shouldEncrypt bool, namespaced bool, watchable bool) (*Informer, error) {
+	watchFunc := func(options metav1.ListOptions) (watch.Interface, error) {
+		return client.Watch(ctx, context.Background(), options)
+	}
+	if !watchable {
+		watchFunc = func(options metav1.ListOptions) (watch.Interface, error) {
+			return newSyntheticWatcher().watch(ctx, client, options, 5*time.Second)
+		}
+	}
 	listWatcher := &cache.ListWatch{
 		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
 			a, err := client.List(ctx, options)
 			return a, err
 		},
-		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
-			return client.Watch(ctx, options)
-		},
+		WatchFunc: watchFunc,
 	}
 
 	example := &unstructured.Unstructured{}
--- a/pkg/sqlcache/informer/informer.go
+++ b/pkg/sqlcache/informer/informer.go
@@ -43,7 +43,8 @@ func NewInformer(ctx context.Context, client dynamic.ResourceInterface, fields [
 	}
 	if !watchable {
 		watchFunc = func(options metav1.ListOptions) (watch.Interface, error) {
-			return newSyntheticWatcher().watch(ctx, client, options, defaultRefreshTime)
+			ctx, cancel := context.WithCancel(context.Background())
+			return newSyntheticWatcher(ctx, cancel).watch(client, options, defaultRefreshTime)
 		}
 	}
 	listWatcher := &cache.ListWatch{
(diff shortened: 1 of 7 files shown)


pkg/sqlcache/informer/synthetic_watcher.go
```diff
@@ -0,0 +1,139 @@
...
+	return rw, nil
+}
+
+type objectHolder struct {
+	version            string
+	unstructuredObject *unstructured.Unstructured
+}
+
+// receive periodically calls client.List(), and converts the returned items into Watch Events
+func (rw *SyntheticWatcher) receive(client dynamic.ResourceInterface, options metav1.ListOptions, interval time.Duration) { //
```
moio, line 42:
```suggestion
func (rw *SyntheticWatcher) receive(client dynamic.ResourceInterface, options metav1.ListOptions, interval time.Duration) {
```

pkg/stores/sqlpartition/listprocessor/processor_test.go
```diff
@@ -518,6 +518,15 @@ func TestParseQuery(t *testing.T) {
...
 	})
+	//tests = append(tests, testCase{
+	//	description: "ParseQuery: filter starting with a comma should return an error.",
+	//	req: &types.APIRequest{
+	//		Request: &http.Request{
+	//			URL: &url.URL{RawQuery: "filter=,"},
+	//		},
+	//	},
+	//	errExpected: true,
+	//})
```
moio, line 529:
```suggestion
```

pkg/sqlcache/informer/informer.go
```diff
@@ -34,15 +34,21 @@ var newInformer = cache.NewSharedIndexInformer
...
 // NewInformer returns a new SQLite-backed Informer for the type specified by schema in unstructured.Unstructured form
 // using the specified client
-func NewInformer(client dynamic.ResourceInterface, fields [][]string, transform cache.TransformFunc, gvk schema.GroupVersionKind, db sqlStore.DBClient, shouldEncrypt bool, namespaced bool) (*Informer, error) {
+func NewInformer(client dynamic.ResourceInterface, fields [][]string, transform cache.TransformFunc, gvk schema.GroupVersionKind, db sqlStore.DBClient, shouldEncrypt bool, namespaced bool, watchable bool) (*Informer, error) {
+	watchFunc := func(options metav1.ListOptions) (watch.Interface, error) {
+		return client.Watch(context.Background(), options)
+	}
+	if !watchable {
+		watchFunc = func(options metav1.ListOptions) (watch.Interface, error) {
+			return newSyntheticWatcher().watch(client, options, 5*time.Second)
```
moio, line 43: This seems to be a reasonable default.

I recommend extracting it into a constant at the top of the file, in case we need to tune it (eg. by GVK) at some point.

pkg/sqlcache/informer/synthetic_watcher.go
```diff
@@ -0,0 +1,139 @@
...
+
+func newSyntheticWatcher() *SyntheticWatcher {
+	return &SyntheticWatcher{
+		stopChan:   make(chan struct{}),
+		doneChan:   make(chan struct{}),
+		resultChan: make(chan watch.Event, 0),
+	}
+}
+
+func (rw *SyntheticWatcher) watch(client dynamic.ResourceInterface, options metav1.ListOptions, interval time.Duration) (*SyntheticWatcher, error) {
```
moio, line 31: It'd be nice if we could make this code context-aware, in order for it to shut down properly in case Steve is shutting down.

In practice this means having a context parameter here and passing it down the forever-loop goroutine.

pkg/sqlcache/informer/informer.go
ericpromislow: Well, a variable with that default value and a setter to call to modify it if necessary.

pkg/sqlcache/informer/synthetic_watcher.go
ericpromislow: I think this should be part of a separate PR.  

First, wouldn't we want to pass the context to `NewInformer` and then pass it to all of `client.Watch()` in the default `WatchFunc` and `client.List()` in the `ListFunc` as well as the synthetic `WatchFunc`?  I'll add it for now, but I would expect contexts are more meaningful if the invoker has access to them.

moio (dismissed):
I think this could work, as those types are few and do not contain a lot of data TTBOMK.

The downside is the full list is kept in memory at all times - a different design could be conceived to avoid that (and offload the list to SQLite).

But that sounds a bit like premature optimization, so we might want to wait to introduce such complexity only if we find out this becomes a real problem.

Similar considerations can be done about the polling time.

ericpromislow:
I think there are only three resources that fit in this category:
```
$ kubectl api-resources -o wide | grep -v -e '\<watch\>' | grep list
componentstatuses                            cs           v1                                false        ComponentStatus                             get,list                                                     
nodes                                                     metrics.k8s.io/v1beta1            false        NodeMetrics                                 get,list                                                     
pods                                                      metrics.k8s.io/v1beta1            true         PodMetrics                                  get,list                      
```

Of course this is subject to change, but that's the nature of rancher.

moio:
> I think there are only three resources that fit in this category: [...]

Exactly, plus any CRDs or other types users explicitly give these permissions. But those should be rare cases where not a lot of resources are involved.

We can react if and when more surfaces.

Thanks!

moio:
I fear my PR adding contexts has now broken this, and a new round of rebasing is necessary.

Sorry about that 😢

moio (approved):
Rebase looks good, confirming approval

ericpromislow:
Sample curl command: `curl -svkL 'https://localhost:5111/v1/metrics.k8s.io.nodes'`

PR #462, opened 2025-01-24
vai: do not discard closing statement errors

moio:
Errors in deferred `CloseStmt` calls are currently ignored.

This patch captures such errors and returns them (possibly joined with another error if it happened).

Part of https://github.com/rancher/rancher/issues/47825

### DIFF
--- a/pkg/sqlcache/informer/indexer.go
+++ b/pkg/sqlcache/informer/indexer.go
@@ -3,6 +3,7 @@ package informer
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"fmt"
 	"reflect"
 	"strings"
@@ -147,7 +148,7 @@ func (i *Indexer) AfterUpsert(key string, obj any, tx transaction.Client) error 
 /* Satisfy cache.Indexer */
 
 // Index returns a list of items that match the given object on the index function
-func (i *Indexer) Index(indexName string, obj any) ([]any, error) {
+func (i *Indexer) Index(indexName string, obj any) (result []any, err error) {
 	i.indexersLock.RLock()
 	defer i.indexersLock.RUnlock()
 	indexFunc := i.indexers[indexName]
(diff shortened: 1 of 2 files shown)


pkg/sqlcache/informer/indexer.go
```diff
@@ -184,7 +185,13 @@ func (i *Indexer) Index(indexName string, obj any) ([]any, error) {
 	// HACK: sql.Statement.Query does not allow to pass slices in as of go 1.19 - create an ad-hoc statement
 	query := fmt.Sprintf(selectQueryFmt, db.Sanitize(i.GetName()), strings.Repeat(", ?", len(values)-1))
 	stmt := i.Prepare(query)
-	defer i.CloseStmt(stmt)
+
+	defer func() {
+		cerr := i.CloseStmt(stmt)
+		if cerr != nil {
+			err = errors.Join(err, cerr)
```
tomleb, line 192: Does the error message from `cerr` provide enough context to know that it came from this close statement call (or just allow us to troubleshoot)?

moio: Hopefully, but that really depends on the SQLite library implementation.

I could add the statement as context if you prefer that way.

tomleb: if it's not too much trouble yes that'd be great to have a bit more context, just in case.

ericpromislow (dismissed):
Straightforward

pkg/sqlcache/informer/indexer.go
moio: addressed as 7d290ce8

TIL: the current library actually never returns errors on close - but I still feel like this code is good to have, as it might change and/or we might switch to another library which might

current https://gitlab.com/cznic/sqlite/-/blob/v1.34.5/sqlite.go#L472-479
possible https://github.com/mattn/go-sqlite3/blob/7658c06970ecf5588d8cd930ed1f2de7223f1010/sqlite3.go#L1896-L1915

moio:
Rebased on top of current `main`

tomleb (approved):

crobby (approved):

PR #463, opened 2025-01-24
vai: document resync period

moio:
Currently all Vai informers are started with a zero interval period, meaning they never force resync all objects.

It was suggested this should be made configurable and aligned with the behavior of other informers in lasso, and this PR implements that.

Requires https://github.com/rancher/lasso/pull/125 and a lasso bump (which I am adding as soon as it is merged).

Addresses the discussion in https://github.com/rancher/lasso/pull/65#discussion_r1607252875, part of https://github.com/rancher/rancher/issues/47825

### DIFF
--- a/pkg/sqlcache/informer/informer.go
+++ b/pkg/sqlcache/informer/informer.go
@@ -48,8 +48,15 @@ func NewInformer(client dynamic.ResourceInterface, fields [][]string, transform 
 	example := &unstructured.Unstructured{}
 	example.SetGroupVersionKind(gvk)
 
-	// avoids the informer to periodically resync (re-list) its resources
-	// currently it is a work hypothesis that, when interacting with the UI, this should not be needed
+	// TL;DR: this disables the Informer periodic resync - but this is inconsequential
+	//
+	// Long version: Informers use a Reflector to pull data from a ListWatcher and push it into a DeltaFIFO.
+	// Concurrently, they pop data off the DeltaFIFO to fire registered handlers, and also to keep an updated
+	// copy of the known state of all objects (in an Indexer).
+	// The resync period option here is passed from Informer to Reflector to periodically (re)-push all known
+	// objects to the DeltaFIFO. That causes the periodic (re-)firing all registered handlers.
+	// In this case we are not registering any handlers to this particular informer, so re-syncing is a no-op.
+	// We therefore just disable it right away.
 	resyncPeriod := time.Duration(0)
 
 	sii := newInformer(listWatcher, example, resyncPeriod, cache.Indexers{})


tomleb:
I initially thought the resync was doing an actual LIST+WATCH resync with the apiserver, but it is simply taking whatever is in its cache and re-running handlers with it. (Confirmed by adding print statements in the various Store.Add, Store.Update, etc call and not seeing them with a custom configured interval of 2 seconds)

Given that there aren't any handlers registered with the vai informer, do we actually need this now? Or is it just a "future-proofing" kind of change?

moio:
> I initially thought the resync was doing an actual LIST+WATCH resync with the apiserver, but it is simply taking whatever is in its cache and re-running handlers with it. (Confirmed by adding print statements in the various Store.Add, Store.Update, etc call and not seeing them with a custom configured interval of 2 seconds)

Whoa.

I have to admit I was under the wrong assumption you mentioned to this day. Probably misguided by a document that circulated years ago as part of the Rancher developer onboarding: https://github.com/aiyengar2/k8s-docs/blob/main/docs/controllers/02_informers.md#introduction-to-the-informer-pattern

I checked client-go's code and indeed it seems like the periodic resync is only used to notify event handlers and setting the internal `Reflector` resync time. And all Reflector does is to call Resync on its associated Store - but is a no-op on "simple" stores like ours (which is modeled after `ThreadSafeStore`).

So that only leaves event handlers, no periodic re-listing.

Then this change is completely useless.

I will use this PR as an occasion to document the behavior well.

pkg/sqlcache/informer/informer.go
```diff
@@ -48,8 +48,17 @@ func NewInformer(client dynamic.ResourceInterface, fields [][]string, transform
...
-	// avoids the informer to periodically resync (re-list) its resources
-	// currently it is a work hypothesis that, when interacting with the UI, this should not be needed
+	// TL;DR: this disables the Informer periodic resync - but this is inconsequential
+	//
+	// Long version: there are two consequences:
+	//
+	// 1: this disables the Informer's underlying Reflector periodic resync. Consequently, the Reflector never calls
+	// its associated Store's Resync() function. In this case we are using sqlstore.Store, which has a no-op Resync(),
+	// so this is inconsequential. (sqlstore's Resync is correctly a no-op: Resync is intended only for extensions of
+	// the Store concept such as DeltaFIFO)
```
tomleb, line 58: > Consequently, the Reflector never calls
	// its associated Store's Resync() function. In this case we are using sqlstore.Store, which has a no-op Resync(),
	// so this is inconsequential.

This isn't quite accurate (and can't blame you for it, the interface is re-used in a few places).

Looking at [sharedIndexInformer.Run](https://github.com/kubernetes/client-go/blob/4765adee376ddbdc4a899ae7c164bf20a4c10089/tools/cache/shared_informer.go#L471-L478):

```go
		fifo := NewDeltaFIFOWithOptions(DeltaFIFOOptions{
			// s.indexer is our custom SQL store
			KnownObjects:          s.indexer,
			EmitDeltaTypeReplaced: true,
			Transformer:           s.transform,
		})

		cfg := &Config{
			Queue:             fifo,
                        ....
		}
```

and [controller.Run](https://github.com/kubernetes/client-go/blob/4765adee376ddbdc4a899ae7c164bf20a4c10089/tools/cache/controller.go#L141-L151):

```go
	r := NewReflectorWithOptions(
              ...
		c.config.Queue,
             ....
	)
```

The reflector's store is actually the DeltaFIFO which does implement [Resync](https://github.com/kubernetes/client-go/blob/4765adee376ddbdc4a899ae7c164bf20a4c10089/tools/cache/delta_fifo.go#L720), but that Resync function just has Get or List access to our SQL store, not Resync or anything else.

So I suppose we could get rid of that first consequence, there's really only one consequence: handlers aren't being fired. (Or rather, objects in the cache aren't being forcingly added to the handler's queue).

tomleb (approved):

PR #464, opened 2025-01-24
Make schema type definitions recursive.

crobby:


pkg/schema/definitions/visitor.go
```diff
@@ -20,23 +20,34 @@ func (s *schemaFieldVisitor) VisitArray(array *proto.Array) {
 	// it was kept this way to provide backwards compat with previous endpoints.
```
ericpromislow, line 20: There's something I don't understand about this PR:

1. The PR title says to make schema type defs recursive
2. The are no changes in the two functions that do anything recursive -- the only recursive bit is in the description field
3. Each function has a comment saying : "this currently is not recursive and provides little information for nested types"

I don't understand the problem enough to see which part this PR is actually solving.

crobby: Fair points.  I think the word "recursive" may have been misused in the issue related to this PR and then I make it worse by using that word again.  I think it's more about accurately reflecting nested definitions.  Once I get something slightly stable, I plan to check with the UI team since they're probably the principal consumers of these schemas to see if what I've done is matching what they are expecting.

crobby:
Note, this is on hold until at least Rancher 2.12.0, will keep it in draft form until then

PR #466, opened 2025-01-28
Fix race in accessstore test

aruiz14:
Refers to https://github.com/rancher/rancher/issues/48902

Setting the same `id` variable from multiple goroutines can be considered a race condition (only affects the test code, not the actual code being tested).

In reality, this was not a problem, because additional checks were performed to ensure that all goroutines would observe. the same value.

### DIFF
--- a/pkg/accesscontrol/access_store_test.go
+++ b/pkg/accesscontrol/access_store_test.go
@@ -315,24 +315,27 @@ func TestAccessStore_AccessFor_concurrent(t *testing.T) {
 
 	const n = 5 // observation showed cases with up to 5 (or more) concurrent queries for the same user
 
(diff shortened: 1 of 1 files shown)


crobby (commented):
Should this be linked to issue https://github.com/rancher/rancher/issues/48902

aruiz14:
> Should this be linked to issue [rancher/rancher#48902](https://github.com/rancher/rancher/issues/48902)

Definitely, thanks!

ericpromislow (approved):
LGTM - Agreed that the data race was only in test code, but we should strive for 0 data races in the test code as well

crobby (approved):

PR #469, opened 2025-01-30
sql: use a closure to wrap transactions

moio:
This introduces the a `WithTransaction` function, which is then used for all transactional work in Steve.

Example:
```go
err := s.WithTransaction(ctx, true, func(tx transaction.Client) error {
		// do things with transaction such as
		_, err := tx.Exec(createTableQuery)
		if err != nil {
			return &db.QueryError{QueryString: createTableQuery, Err: err}
		}
		return nil
	})
```

Because `WithTransaction` takes care of all `Begin`s, `Commit`s and `Rollback`s, it eliminates the problem where forgotten open transactions can block all other operations (with long stalling and `SQLITE_BUSY` errors).

This also:

- merges together the disparate `DBClient` interfaces in one only `db.Client` interface with one unexported non-test implementation. I found this much easier to follow
- refactors the transaction package in order to make it as minimal as possible, and as close to the wrapped `sql.Tx` and `sql.Stmt` functions as possible, in order to reduce cognitive load when working with this part of the codebase
- simplifies tests accordingly
- adds a couple of known files to `.gitignore`

Partly addresses https://github.com/rancher/rancher/issues/48384

**Best reviewed commit-by-commit** - the vast majority of the changes are actually mechanical. The core of this change is "just" commit 8d7272eba070b557c85de6ebe8148a7e67170588
    
Credits to @tomleb for [suggesting the approach](https://github.com/rancher/lasso/pull/121#pullrequestreview-2515872507).


pkg/sqlcache/db/client.go
```diff
@@ -29,8 +30,59 @@ const (
 	informerObjectCachePerms fs.FileMode = 0o600
 )
 
-// Client is a database client that provides encrypting, decrypting, and database resetting.
-type Client struct {
+// Client is a database client that provides encrypting, decrypting, and database resetting
```
ericpromislow, line 33: "Client defines a database client…" since it's now an interface

pkg/sqlcache/db/client.go
```diff
@@ -29,8 +30,59 @@ const (
...
+	c.connLock.RUnlock()
+	if err != nil {
+		return err
+	}
+
+	err = f(transaction.NewClient(tx))
+
+	if err != nil {
+		rerr := tx.Rollback()
+		err = errors.Join(err, rerr)
```
ericpromislow, line 72: Shouldn't we be appending `rerr` and `cerr` only if they're non-nil?

pkg/sqlcache/db/client.go
```diff
@@ -114,19 +157,19 @@ func NewClient(c Connection, encryptor Encryptor, decryptor Decryptor) (*Client,
...
 
 // Prepare prepares the given string into a sql statement on the client's connection.
-func (c *Client) Prepare(stmt string) *sql.Stmt {
+func (c *client) Prepare(stmt string) *sql.Stmt {
 	c.connLock.RLock()
 	defer c.connLock.RUnlock()
 	prepared, err := c.conn.Prepare(stmt)
 	if err != nil {
-		panic(errors.Errorf("Error preparing statement: %s\n%v", stmt, err))
+		panic(fmt.Errorf("Error preparing statement: %s\n%v", stmt, err))
```
ericpromislow, line 165: Sort of existing code, but I believe the `%v` should be `%w` to process wrapped errors correctly.

pkg/sqlcache/informer/listoption_indexer_test.go
```diff
@@ -15,12 +15,16 @@ import (
...
 	"testing"
 
+	"github.com/rancher/steve/pkg/sqlcache/db"
 	"github.com/rancher/steve/pkg/sqlcache/partition"
 	"github.com/stretchr/testify/assert"
 	"go.uber.org/mock/gomock"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
+	//"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
```
ericpromislow, line 25: Do we still need these commented imports?

ericpromislow (changes requested):
Just some cosmetic suggestions -- makes sense when I walk through the commits as recommended, and looks good - making this straightforward. And I didn't think we were always rolling back on failures, in some cases we were leaving transactions dangling. Which I guess might work sometimes....

pkg/sqlcache/db/client.go
moio: Addressed in e966723c

pkg/sqlcache/informer/listoption_indexer_test.go
moio: No, sorry for letting that slip in.

Addressed in e966723c

pkg/sqlcache/db/client.go
moio: [Join discards nils](https://pkg.go.dev/errors#Join), so my understanding is the current code should be correct.

pkg/sqlcache/db/client.go
moio: You are right, addressed in 31c03f8a

moio:
> And I didn't think we were always rolling back on failures, in some cases we were leaving transactions dangling. Which I guess might work sometimes....

Right, that's the whole point of this PR - [we had cases in which dangling transactions blocked new ones](https://github.com/rancher/rancher/issues/48384). This should really prevent that from happening!

moio:
Rebased on top of current `main` and force pushed.

tomleb (approved):
Looks great, I like how you hid the `Commit()` / `Rollback()` functionality to ensure no one is calling it.

ericpromislow (approved):
LGTM

PR #507, opened 2025-02-18
[v0.4] Add tagged dependencies

vatsalparekh:


go.mod
```diff
@@ -22,8 +22,8 @@ require (
 	github.com/rancher/apiserver v0.4.2
 	github.com/rancher/dynamiclistener v0.6.1
 	github.com/rancher/kubernetes-provider-detector v0.1.5
-	github.com/rancher/lasso v0.0.0-20241202185148-04649f379358
-	github.com/rancher/norman v0.0.0-20241001183610-78a520c160ab
+	github.com/rancher/lasso v0.2.0
```
crobby, line 25: I think we can use lasso v0.2.1

go.mod
```diff
@@ -22,8 +22,8 @@ require (
...
+	github.com/rancher/norman v0.4.0
```
crobby, line 26: Latest norman tag is v0.4.2

crobby (changes requested):
We should probably use the latest tags for both lasso and norman

tomleb (approved):

PR #512, opened 2025-02-19
Fix "Unable to cancel request for *client.addQuery" warnings

aruiz14:
These warning messages appear frequently in Rancher logs, and happen when attempting to cancel a request but one of the `http.RoundTripper`s does not have a `CancelRequest` method (not part of the `RoundTripper` interface.

All the `http.RoundTripper`s in `client-go`'s `transport` package have this method, and use a private `tryCancelRequest` method for propagating such cancellation. `addQuery` would need to either provide its own `CancelRequest` or have a way to jump to the next RoundTripper, in order to not interrupt the cancellation to the rest of the chain.
The latter can be done by implementing the [`RoundTripperWrapper` interface](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/net#RoundTripperWrapper), as `tryCancelRequest` [already supports it](https://github.com/kubernetes/client-go/blob/v0.31.0/transport/transport.go#L353-L354).

### DIFF
--- a/pkg/client/factory.go
+++ b/pkg/client/factory.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/rancher/apiserver/pkg/types"
 	"github.com/rancher/steve/pkg/attributes"
+	utilnet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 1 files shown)


tomleb (approved):

bigkevmcd (approved):
This is basically because we fall through in the request cancellation logic, this will fix the issue.

PR #568, opened 2025-03-22
Fix namespace access control in steve

rohitsakala:
Issue https://github.com/rancher/rancher/issues/39378
Issue https://github.com/rancher/rancher/issues/39129

# Summary 
- When you create a `Role` with resources being `namespaces`, the access.Namespace field doesn't apply. Resources is `namespaces` and you cannot apply access.Namespace on `namespaces`.  So, defaulting it to `*`. 
- Also, the namespace becomes the resourceName.

# Understanding RBAC in Kubernetes for Namespaces 
- I will explain this with an example. Suppose, I have created a namespace default, created a `Service Account` in that namespace, created a `ClusterRole` on namespaces with `resourceNames` as default with `get`, `patch` and `update` verbs. Then, I created a `rolebinding` with `ClusterRole` and `Service Account`. 
- The user can perform edit on namespace on `default` only. 
- Now, consider the same scenario except that the resources is now `nodes` instead of `namespaces`. You cannot edit the node `kind`. 
- You might understand here that there is a special case for `namespaces`. So what is the proof of this ? The proof lies in the codebase of `kubernetes` 

- There is a special case where [if the resource is `namespaces` ](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go#L194), it sets the `requestInfo.Namespace` as the `ResourceName`. In `RBAC` codebase, there is a condition where if the `requestInfo.Namespace` is not empty, `rolebindings` are considered. So, the special condition enables the edit for the user in the case of `namespaces` resources. 

- In all other cases, [it doesn't set it as `ResourceName`](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go#L204). So, the `nodes` case fails.

### DIFF
--- a/pkg/accesscontrol/policy_rule_index.go
+++ b/pkg/accesscontrol/policy_rule_index.go
@@ -98,14 +98,31 @@ func addResourceAccess(accessSet *AccessSet, namespace string, rule rbacv1.Polic
 			}
 			for _, resourceName := range names {
 				for _, verb := range rule.Verbs {
+					access := Access{
+						Namespace:    namespace,
+						ResourceName: resourceName,
+					}
+
+					// The first condition namespace != All is to determine if it is a RoleBinding.
+					// The second and third conditions are to check if the resource is for "namespaces" in core group.
+					// In kubernetes, rule are valid if they satisfy the following
+					// 	- Should be `namespaces` GR
+					//  - From RoleBindings in `namespace`
+					//  - From Rule with  ResourceName `*`` or the `namespace` itself.
+					// Ref: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go#L194
+					// If the ResourceName is `All` || namespace itself then only the current namespace is considered as Resourcename
+					// In the case of Rolebinding for the resource "namespaces" in core group, access.Namespace
+					// is set to All since namespace on the resource "namespaces" is not valid.
+					if namespace != All && resource == "namespaces" && group == "" && (resourceName == All || resourceName == namespace) {
+						access.Namespace = All
+						access.ResourceName = namespace
+					}
+
 					accessSet.Add(verb,
 						schema.GroupResource{
 							Group:    group,
 							Resource: resource,
-						}, Access{
-							Namespace:    namespace,
-							ResourceName: resourceName,
-						})
+						}, access)
 				}
 			}
 		}
--- a/pkg/accesscontrol/policy_rule_index_test.go
+++ b/pkg/accesscontrol/policy_rule_index_test.go
@@ -1,11 +1,13 @@
 package accesscontrol
 
 import (
+	"reflect"
 	"slices"
 	"testing"
 
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 func Test_policyRuleIndex_roleBindingBySubject(t *testing.T) {
@@ -234,5 +236,153 @@ func makeCRB(name string, roleRef rbacv1.RoleRef, subjects []rbacv1.Subject) *rb
 		},
 		RoleRef:  roleRef,
 		Subjects: subjects,
+	}
+}
+
+func Test_addResourceAccess(t *testing.T) {
+	tests := []struct {
+		name      string
+		namespace string
+		rule      rbacv1.PolicyRule
+		want      AccessSet
+	}{
+		{
+			name:      "RoleBinding namespaces resource with empty names",
+			namespace: "test-ns",
+			rule: rbacv1.PolicyRule{
+				APIGroups:     []string{""},
+				Resources:     []string{"namespaces", "deployments"},
+				ResourceNames: []string{},
+				Verbs:         []string{"get"},
+			},
+			want: AccessSet{
+				set: map[key]resourceAccessSet{
+					{
+						verb: "get", gr: schema.GroupResource{Group: "", Resource: "namespaces"}}: {
+						Access{Namespace: "*", ResourceName: "test-ns"}: true,
+					},
+					{
+						verb: "get", gr: schema.GroupResource{Group: "", Resource: "deployments"}}: {
+						Access{Namespace: "test-ns", ResourceName: "*"}: true,
+					},
+				},
+			},
+		},
+		{
+			name:      "ClusterRoleBinding namespaces resource with empty names",
+			namespace: "*",
+			rule: rbacv1.PolicyRule{
+				APIGroups:     []string{""},
+				Resources:     []string{"namespaces", "deployments"},
+				ResourceNames: []string{},
+				Verbs:         []string{"get"},
+			},
+			want: AccessSet{
+				set: map[key]resourceAccessSet{
+					{verb: "get", gr: schema.GroupResource{Group: "", Resource: "namespaces"}}: {
+						Access{Namespace: "*", ResourceName: "*"}: true,
+					},
+					{
+						verb: "get", gr: schema.GroupResource{Group: "", Resource: "deployments"}}: {
+						Access{Namespace: "*", ResourceName: "*"}: true,
+					},
+				},
+			},
+		},
+		{
+			name:      "RoleBinding namespaces resource with specific names",
+			namespace: "test-ns",
+			rule: rbacv1.PolicyRule{
+				APIGroups:     []string{""},
+				Resources:     []string{"namespaces"},
+				ResourceNames: []string{"specific-ns"},
+				Verbs:         []string{"get"},
+			},
+			want: AccessSet{
+				set: map[key]resourceAccessSet{
+					{verb: "get", gr: schema.GroupResource{Group: "", Resource: "namespaces"}}: {
+						Access{Namespace: "test-ns", ResourceName: "specific-ns"}: true,
+					},
+				},
+			},
+		},
+		{
+			name:      "RoleBinding namespaces resource with its own namespace",
+			namespace: "test-ns",
+			rule: rbacv1.PolicyRule{
+				APIGroups:     []string{""},
+				Resources:     []string{"namespaces"},
+				ResourceNames: []string{"test-ns"},
+				Verbs:         []string{"get"},
+			},
+			want: AccessSet{
+				set: map[key]resourceAccessSet{
+					{verb: "get", gr: schema.GroupResource{Group: "", Resource: "namespaces"}}: {
+						Access{Namespace: "*", ResourceName: "test-ns"}: true,
+					},
+				},
+			},
+		},
+		{
+			name:      "RoleBinding other resource with empty names",
+			namespace: "test-ns",
+			rule: rbacv1.PolicyRule{
+				APIGroups:     []string{"apps"},
+				Resources:     []string{"deployments"},
+				ResourceNames: []string{},
+				Verbs:         []string{"get"},
+			},
+			want: AccessSet{
+				set: map[key]resourceAccessSet{
+					{verb: "get", gr: schema.GroupResource{Group: "apps", Resource: "deployments"}}: {
+						Access{Namespace: "test-ns", ResourceName: "*"}: true,
+					},
+				},
+			},
+		},
+		{
+			name:      "ClusterRoleBinding other resource with empty names",
+			namespace: "*",
+			rule: rbacv1.PolicyRule{
+				APIGroups:     []string{"apps"},
+				Resources:     []string{"deployments"},
+				ResourceNames: []string{},
+				Verbs:         []string{"get"},
+			},
+			want: AccessSet{
+				set: map[key]resourceAccessSet{
+					{verb: "get", gr: schema.GroupResource{Group: "apps", Resource: "deployments"}}: {
+						Access{Namespace: "*", ResourceName: "*"}: true,
+					},
+				},
+			},
+		},
+		{
+			name:      "RoleBinding other resource with specific names",
+			namespace: "test-ns",
+			rule: rbacv1.PolicyRule{
+				APIGroups:     []string{"apps"},
+				Resources:     []string{"deployments"},
+				ResourceNames: []string{"my-deploy"},
+				Verbs:         []string{"get"},
+			},
+			want: AccessSet{
+				set: map[key]resourceAccessSet{
+					{verb: "get", gr: schema.GroupResource{Group: "apps", Resource: "deployments"}}: {
+						Access{Namespace: "test-ns", ResourceName: "my-deploy"}: true,
+					},
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			accessSet := &AccessSet{}
+			addResourceAccess(accessSet, tt.namespace, tt.rule)
+			if !reflect.DeepEqual(*accessSet, tt.want) {
+				t.Errorf("addResourceAccess() got = %v, want %v", *accessSet, tt.want)
+			}
+		})
 	}
 }


ericpromislow (approved):
Code looks good, changes work

ericpromislow (changes requested):
Looks like there's a problem here -- I can no longer view a cluster when I'm a member of the cluster and a readonly member of its only custom project:
```
 HTTP Error 403:
namespaces is forbidden: User "u-npsfh" cannot list resource "namespaces" in API group "" at the cluster scope 
```

tomleb:
Can you add some tests? I haven't looked at the changes yet.

rohitsakala:
> Looks like there's a problem here -- I can no longer view a cluster when I'm a member of the cluster and a readonly member of its only custom project:
> 
> ```
>  HTTP Error 403:
> namespaces is forbidden: User "u-npsfh" cannot list resource "namespaces" in API group "" at the cluster scope 
> ```

Fixed it.

pkg/accesscontrol/policy_rule_index.go
```diff
@@ -93,19 +93,29 @@ func addResourceAccess(accessSet *AccessSet, namespace string, rule rbacv1.Polic
 	for _, group := range rule.APIGroups {
 		for _, resource := range rule.Resources {
 			names := rule.ResourceNames
-			if len(names) == 0 {
-				names = []string{All}
+			if namespace != All && resource == "namespaces" && group == "" {
```
crobby, line 96: Maybe add a comment here to help our future selves understand what is going on here.

crobby (commented):
This change seems reasonable and I think it has a chance at fixing more issues than just the originally targeted issue.  
Needs some tests added, but I think this is promising.

tomleb:
To help with the review, could you provide RBAC manifests that used to not work and now do?

pkg/accesscontrol/policy_rule_index.go
ericpromislow: Yes, what's so special about `namespace != All` that you need to set `names = []string{namespace}` ?  What if `names` is set to a non-empty array of some other value? Can it just bet thrown out?

pkg/accesscontrol/policy_rule_index.go
```diff
@@ -93,19 +93,29 @@ func addResourceAccess(accessSet *AccessSet, namespace string, rule rbacv1.Polic
...
+				names = []string{namespace}
+			} else {
```
ericpromislow, line 98: This can be written as
```
} else if len(names) == 0 {
names = []string{All}
}
```

ericpromislow:
This change fixes issues 39378 (as specified) and also 39129. Please add an explanatory comment and maybe the refactoring of the `if ... else` (mainly because I never like to see `else { if (...) { ... } }` at the end of a block) and I can approve.

rohitsakala:
> To help with the review, could you provide RBAC manifests that used to not work and now do?

@tomleb The RBAC manifests are not changing. The bug was in the logic of the access control codebase.

pkg/accesscontrol/policy_rule_index.go
rohitsakala: I added code comments and changed the code. please re-review.

pkg/accesscontrol/policy_rule_index.go
```diff
@@ -94,8 +94,22 @@ func addResourceAccess(accessSet *AccessSet, namespace string, rule rbacv1.Polic
 		for _, resource := range rule.Resources {
 			names := rule.ResourceNames
 			if len(names) == 0 {
-				names = []string{All}
+				// The first condition namespace != All is to deduct if it is a RoleBinding.
+				// The second and third conditions are to check if the resource is for "namespaces" in core group.
+				// If names is empty, it means the resource is for all namespaces [else condition] and in the case
+				// of RoleBinding on the resource "namespaces" in core group, it means the the resourceName will be
```
ericpromislow, line 100: typo `the the` => `the`

pkg/accesscontrol/policy_rule_index.go
```diff
@@ -94,8 +94,22 @@ func addResourceAccess(accessSet *AccessSet, namespace string, rule rbacv1.Polic
 		for _, resource := range rule.Resources {
 			names := rule.ResourceNames
 			if len(names) == 0 {
-				names = []string{All}
+				// The first condition namespace != All is to deduct if it is a RoleBinding.
```
ericpromislow, line 97: s/deduct/determine/

ericpromislow (changes requested):
Just some trivial typos in the comment.

The unit test explains what this is supposed to do

crobby (commented):
I think this is doing what is expected.  Once Eric's comments are addressed, we might be good to go.

pkg/accesscontrol/policy_rule_index.go
```diff
@@ -94,8 +94,22 @@ func addResourceAccess(accessSet *AccessSet, namespace string, rule rbacv1.Polic
...
+				// The second and third conditions are to check if the resource is for "namespaces" in core group.
+				// If names is empty, in the case of ClusterRoleBinding names is set to All.
+				// If names is empty, in the case of RoleBinding on the resource "namespaces" in core group, names is set to namespace
+				// in which the Rolebinding resides as kubernetes allows it and sets the requestInfo.Namespace to ResourceName.
+				// Ref: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go#L194
+				// In the case of Rolebinding, namespace is set to All in the case of resource "namespaces" in core group since namespace on
+				// the resource "namespaces" is not valid.
+				if namespace != All && resource == "namespaces" && group == "" {
+					names = []string{namespace}
+					namespace = All
```
pjbgf, line 107: We can't override the `namespace` here without opening up for privesc when multiple resources are being requested. Otherwise, we may be changing the namespace scope to `*` for other Resources in this request.

Instead the logic needs to change so that the `accessSet.Add` for this specific edge case is handled here, without changing the `namespace` var for everybody else.

pjbgf (changes requested):
We need to amend the handling of the `namespace` var.

pkg/accesscontrol/policy_rule_index_test.go
```diff
@@ -236,3 +238,142 @@ func makeCRB(name string, roleRef rbacv1.RoleRef, subjects []rbacv1.Subject) *rb
...
+		name      string
+		namespace string
+		rule      rbacv1.PolicyRule
+		want      AccessSet
+	}{
+		{
+			name:      "RoleBinding namespaces resource with empty names",
+			namespace: "test-ns",
+			rule: rbacv1.PolicyRule{
+				APIGroups:     []string{""},
```
pjbgf, line 253: Please add test cases with multiple Resources and APIGroups.

tomleb:
> We need to amend the handling of the `namespace` var.

Not sure if that's what you're referring to but I was just sending a message to Rohit that I think the fix should be in where we parse the request info to give the proper Namespace value based on what k8s does. That way no special handling in the RBAC code.

Basically, if the request is for the namespaces GVR, Kubernetes sets the namespace var even if it's a cluster scope resource. (Linked by Rohit above: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go#L194). I checked and they also have a test for this here: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo_test.go#L51.

So if we fix this (in rancher/apiserver) then I expect the AccessStore in Steve to work correct and this bug to be fixed. (Though we might uncover other bugs, which makes me a bit nervous).

pkg/accesscontrol/policy_rule_index.go
rohitsakala: Agreed.

rohitsakala:
> Not sure if that's what you're referring to but I was just sending a message to Rohit that I think the fix should be in where we parse the request info to give the proper Namespace value based on what k8s does. That way no special handling in the RBAC code.
> 
> Basically, if the request is for the namespaces GVR, Kubernetes sets the namespace var even if it's a cluster scope resource. (Linked by Rohit above: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go#L194). I checked and they also have a test for this here: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo_test.go#L51.
> 
> So if we fix this (in rancher/apiserver I believe) then I expect the AccessStore in Steve to work correct and this bug to be fixed. (Though we might uncover other bugs, which makes me a bit nervous).

Let me walk through the codebase from UI -> `rancher/steve || rancher/apiserver`.

-  The `UI` requests for the URL `v1/namespaces` by `GET`. UI checks the `link` of the response to understand if the user has access to a verb or not ? If the `link` to `update` is present, then the UI assumes the user can `update` that particular namespace.

```
        "id": "default",
            "type": "namespace",
            "links": {
                "self": "https://localhost:8443/v1/namespaces/default",
                "view": "https://localhost:8443/api/v1/namespaces/default",
                "update": "./...."
            },
            "apiVersion": "v1",
```

- The request first tries to get the schema for that particular user at [#1](https://github.com/rancher/steve/blob/main/pkg/schema/factory.go#L35). Here, at `c.as.AccessFor(user)`, it gets the AccessSet for that particular user for all resources.

- It builds the `AccessSet` at [#2](https://github.com/rancher/steve/blob/main/pkg/accesscontrol/access_store.go#L75) and during the process, it loops over the `rules` of `RoleBinding` to add them to the `AccessSet` at [#3](https://github.com/rancher/steve/blob/main/pkg/accesscontrol/policy_rule_index.go#L82) and then calls `addResourceAccess`. 

- In `addResourceAccess` , in our use case the rule is `patch & update` on `namespaces`. Let's see how it gets turned into `AccessSet`. Since the `resourceName` is `nil` in our `RoleBinding`, `names` becomes `*`. The resulting `AccessSet` contains the following information. 

```
update = Access{Namespace : "default", ResourceName : "*"}
patch = Access{Namespace : "default", ResourceName : "*"}
```

- After getting the `namespaces`, time comes for building the response and while adding the links at [formatter func ](https://github.com/rancher/steve/blob/main/pkg/resources/common/formatter.go#L76), the [Grants func](https://github.com/rancher/steve/blob/main/pkg/resources/common/formatter.go#L101) check happens where it tries to match `namespace` `resource name` `group`  `verb` and `name` with the one in `AccessSet`.

```
verb : `patch`
gr : `namespaces`
namespace : `""`
name : `default`
```

is equal to 

```
patch = Access{Namespace : "default", ResourceName : "*"}

verb : `patch`
gr: `namespaces`
namespace : `default`
name: `*` 
```

- These both are not equal at `namespace` only. Grants accepts `*` in both `name` or `namespace`. So `name` doesn't cause the non equality. . So, due to `namespace` alone, `Grants` doesn't give access though `AccessSet` has it. We can set `namespace` to `name : default` just like how `kubernetes` does it, but it is not fixing the bug.

- Why ? Because the addition of `links` depend on `ResourceMethods` and it depends [on this codebase](https://github.com/rancher/steve/blob/main/pkg/schema/factory.go#L105). where it removes the entries that were created above inside `AccessSet` and the comment explains why so ?. I made an exception for `namespaces` GR and the bug got fixed. 

- We need both the exceptions. 1: at `Grants` func like kubernetes. 2: An exception for `namespaces` GR at  [this codebase](https://github.com/rancher/steve/blob/main/pkg/schema/factory.go#L105). 

- The 2nd exception is similar to what we have in this PR. I feel the solution in this PR makes sense.

- Why do I feel it is more right ?

- In the PR, I am specifically doing semantically similar to what kubernetes is doing. 
  - The current PR adds `namespace` as `resourceName` to every `namespaces` GR so that the requested `resourceName` on `namespaces` GR gets satisfied (similar to kubernetes request)
    - Kubernetes adds requested `resourceName` as `namespace` for every request of `namespaces` GR.
  - Two, the [codebase](https://github.com/rancher/steve/blob/main/pkg/schema/factory.go#L105) removes the `AccessSet` as `invalid` and we are adding the 2nd exception on the items of the `AccessSet`. It is better to fix the `AccessSet`. 
  - Three, the above entries `update = Access{Namespace : "default", ResourceName : "*"}` semantically look wrong. The `ResourceName` should be `default` but it is `*`.
  - Four, because of the diagram below. 
  
  
Screenshot 2025-04-02 at 6 20 37 PM

In kubernetes world, it the RBAC checking happens by looking into the `rules` of the bindings, but in Rancher world, the RBAC checking happens by looking into the `AccessSet` which is built on top of bindings. We are needing to writing 2nd exception because our `AccessSet` is not build properly. So, I would like to build the `AccessSet` properly like in this PR.

  
I can be completely wrong here since I am a newbie. Let me know your thoughts @tomleb @rancher/rancher-squad-frameworks.

rohitsakala:
> Not sure if that's what you're referring to but I was just sending a message to Rohit that I think the fix should be in where we parse the request info to give the proper Namespace value based on what k8s does. That way no special handling in the RBAC code.
> 
> Basically, if the request is for the namespaces GVR, Kubernetes sets the namespace var even if it's a cluster scope resource. (Linked by Rohit above: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go#L194). I checked and they also have a test for this here: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo_test.go#L51.
> 
> So if we fix this (in rancher/apiserver I believe) then I expect the AccessStore in Steve to work correct and this bug to be fixed. (Though we might uncover other bugs, which makes me a bit nervous).

Let me walk through the codebase from UI -> `rancher/steve || rancher/apiserver`.

-  The `UI` requests for the URL `v1/namespaces` by `GET`. UI checks the `link` of the response to understand if the user has access to a verb or not ? If the `link` to `update` is present, then the UI assumes the user can `update` that particular namespace.

```
        "id": "default",
            "type": "namespace",
            "links": {
                "self": "https://localhost:8443/v1/namespaces/default",
                "view": "https://localhost:8443/api/v1/namespaces/default",
                "update": "./...."
            },
            "apiVersion": "v1",
```

- The request first tries to get the schema for that particular user at [#1](https://github.com/rancher/steve/blob/main/pkg/schema/factory.go#L35). Here, at `c.as.AccessFor(user)`, it gets the AccessSet for that particular user for all resources.

- It builds the `AccessSet` at [#2](https://github.com/rancher/steve/blob/main/pkg/accesscontrol/access_store.go#L75) and during the process, it loops over the `rules` of `RoleBinding` to add them to the `AccessSet` at [#3](https://github.com/rancher/steve/blob/main/pkg/accesscontrol/policy_rule_index.go#L82) and then calls `addResourceAccess`. 

- In `addResourceAccess` , in our use case the rule is `patch & update` on `namespaces`. Let's see how it gets turned into `AccessSet`. Since the `resourceName` is `nil` in our `RoleBinding`, `names` becomes `*`. The resulting `AccessSet` contains the following information. 

```
update = Access{Namespace : "default", ResourceName : "*"}
patch = Access{Namespace : "default", ResourceName : "*"}
```

- After getting the `namespaces`, time comes for building the response and while adding the links at [formatter func ](https://github.com/rancher/steve/blob/main/pkg/resources/common/formatter.go#L76), the [Grants func](https://github.com/rancher/steve/blob/main/pkg/resources/common/formatter.go#L101) check happens where it tries to match `namespace` `resource name` `group`  `verb` and `name` with the one in `AccessSet`.

```
verb : `patch`
gr : `namespaces`
namespace : `""`
name : `default`
```

is equal to 

```
patch = Access{Namespace : "default", ResourceName : "*"}

verb : `patch`
gr: `namespaces`
namespace : `default`
name: `*` 
```

- These both are not equal at `namespace` only. Grants accepts `*` in both `name` or `namespace`. So `name` doesn't cause the non equality. . So, due to `namespace` alone, `Grants` doesn't give access though `AccessSet` has it. We can set `namespace` to `name : default` just like how `kubernetes` does it, but it is not fixing the bug.

- Why ? Because the addition of `links` depend on `ResourceMethods` and it depends [on this codebase](https://github.com/rancher/steve/blob/main/pkg/schema/factory.go#L105). where it removes the entries that were created above inside `AccessSet` and the comment explains why so ?. I made an exception for `namespaces` GR and the bug got fixed. 

- We need both the exceptions. 1: at `Grants` func like kubernetes. 2: An exception for `namespaces` GR at  [this codebase](https://github.com/rancher/steve/blob/main/pkg/schema/factory.go#L105). 

- The 2nd exception is similar to what we have in this PR. I feel the solution in this PR makes sense.

- Why do I feel it is more right ?

- In the PR, I am specifically doing semantically similar to what kubernetes is doing. 
  - Kubernetes adds requested `resourceName` as `namespace` for every request of `namespaces` GR which means it considers all `rules`  related to `namespaces` GR  of `RoleBindings` which means that `namespaces` GR rules are applied to the `resourceName`.
  - The current PR adds `namespace` as `resourceName` to all rules specific to `namespaces` GR in the `RoleBindings` which means the AccessSet is built in a way that `namespace` GR rules are applied to the `namespace` as well. 
  - Two, the [codebase](https://github.com/rancher/steve/blob/main/pkg/schema/factory.go#L105) removes the entries in `AccessSet` as `invalid` and we are adding the 2nd exception on the items of the `AccessSet`. It is better to fix the `AccessSet`. 
  - Three, the above entries `update = Access{Namespace : "default", ResourceName : "*"}` semantically looks wrong. The `ResourceName` should be `default` but it is `*`.
  - Four, because of the diagram below. 
  
Screenshot 2025-04-02 at 6 20 37 PM

In kubernetes world, it the RBAC checking happens by looking into the `rules` of the bindings, but in Rancher world, the RBAC checking happens by looking into the `AccessSet` which is built on top of bindings. We are needing to writing 2nd exception because our `AccessSet` is not build properly. So, I would like to build the `AccessSet` properly like in this PR.

  
I can be completely wrong here since I am a newbie. Let me know your thoughts @tomleb @rancher/rancher-squad-frameworks.

pkg/accesscontrol/policy_rule_index_test.go
rohitsakala: Added.

pjbgf:
No concerns on the current changes, but I'd defer to @tomleb and team as to whether this is the best way to resolve the problem at hand.

crobby (approved):

tomleb:
Yeah I think semantically it might end up being the same. I think ideally we'd keep closer to how upstream is addressing this case, but if it works then that's fine too. :+1:

ericpromislow (approved):
Problems are solved and the explanation here is great.

pjbgf (approved):

PR #594, opened 2025-04-08
Include namespace-specific permission in schemas

crobby:
Issue:  https://github.com/rancher/rancher/issues/34731 and others from the https://github.com/rancher/rancher/issues/45827 epic.
https://github.com/rancher/rancher/issues/48788
https://github.com/rancher/rancher/issues/47393

### DIFF
--- a/pkg/resources/common/formatter.go
+++ b/pkg/resources/common/formatter.go
@@ -175,7 +175,9 @@ func formatter(summarycache common.SummaryCache, asl accesscontrol.AccessSetLook
 						perms[verb] = url
 					}
 				}
-				permissions[res] = perms
+				if len(perms) > 0 {
+					permissions[res] = perms
+				}
 			}
 
 			if unstr, ok := resource.APIObject.Object.(*unstructured.Unstructured); ok {
--- a/pkg/resources/common/formatter.go
+++ b/pkg/resources/common/formatter.go
@@ -9,7 +9,6 @@ import (
 	"github.com/rancher/steve/pkg/attributes"
 	"github.com/rancher/steve/pkg/resources/virtual/common"
 	"github.com/rancher/steve/pkg/schema"
-	metricsStore "githu
 	metricsStore "github.com/rancher/steve/pkg/stores/metrics"
 	"github.com/rancher/steve/pkg/stores/proxy"
 	"github.com/rancher/steve/pkg/summarycache"
--- a/pkg/resources/common/formatter.go
+++ b/pkg/resources/common/formatter.go
@@ -236,11 +236,8 @@ func getNamespaceFromResource(obj types.APIObject) string {
 		}
 	}
 
-	// Otherwise, if the id has a slash, we will interpret that
-	parts := strings.SplitN(obj.ID, "/", 2)
-	if len(parts) == 2 {
-		return parts[0] + "-" + parts[1]
-	}
-
-	return obj.ID
+	// Otherwise, if the id has a slash, we will interpret that.
+	// This is used to determine a project's namespace when there is no backingNamespace present.
+	// For cases where there is no slash, we use the object's ID, which is the same as the namespace.
+	return strings.Replace(obj.ID, "/", "-", 1)
 }
--- a/pkg/resources/common/formatter.go
+++ b/pkg/resources/common/formatter.go
@@ -146,6 +146,27 @@ func formatter(summarycache *summarycache.SummaryCache, asl accesscontrol.Access
 			excludeValues(request, unstr)
 		}
 
+		if permsQuery := request.Query.Get("checkPermissions"); permsQuery != "" {
+			id := resource.APIObject.ID
+			ns := getNamespaceFromResource(id)
+			gvr := attributes.GVR(resource.Schema)
+			permissions := map[string]map[string]bool{}
+			for _, res := range strings.Split(permsQuery, ",") {
+				perms := map[string]bool{}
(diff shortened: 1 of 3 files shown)


pkg/resources/common/formatter_test.go
```diff
@@ -1072,3 +1075,223 @@ func Test_formatterLinks(t *testing.T) {
...
+		resourcePermissions map[string][]string
+		schema              *types.APISchema
+		apiObject           types.APIObject
+		want                map[string]map[string]string
+	}{
+		{
+			name:                "get update patch on project and get on projectroletemplatebindings",
+			topLevelPermissions: []string{"get", "update", "patch"},
+			resourcePermissions: map[string][]string{
+				"projectroletemplatebindings": {"get", "list", "watch"},
```
ericpromislow, line 1097: Would it be possible to use camelCase for this PRTB key and make it easier to read?

ericpromislow (changes requested):
Just some code-related requests. Looks fine otherwise.

pkg/resources/common/formatter.go
```diff
@@ -184,3 +220,27 @@ func excludeValues(request *types.APIRequest, unstr *unstructured.Unstructured)
...
+	if statusRaw, ok := unstr.Object["status"]; ok {
+		if statusMap, ok := statusRaw.(map[string]interface{}); ok {
+			if backingNamespace, ok := statusMap["backingNamespace"].(string); ok && backingNamespace != "" {
+				return backingNamespace
+			}
+		}
+	}
+
+	// Otherwise, if the id has a slash, we will interpret that
+	parts := strings.SplitN(obj.ID, "/", 2)
```
rohitsakala, line 240: Is it always true that the second part will be a `namespace` after `/` ?

pkg/resources/common/formatter.go
```diff
@@ -184,3 +220,27 @@ func excludeValues(request *types.APIRequest, unstr *unstructured.Unstructured)
...
+	if len(parts) == 2 {
+		return parts[0] + "-" + parts[1]
+	}
+
+	return obj.ID
```
rohitsakala, line 245: We are returning obj.ID. How is obj.ID a namespace ?

pkg/resources/common/formatter.go
```diff
@@ -146,6 +156,32 @@ func formatter(summarycache *summarycache.SummaryCache, asl accesscontrol.Access
...
+				gr := schema2.GroupResource{Group: gvr.Group, Resource: gvr.Resource}
+				perms := map[string]string{}
+
+				for _, verb := range []string{"create", "update", "delete", "list", "get", "watch", "patch"} {
+					if accessSet.Grants(verb, gr, ns, "") {
+						url := request.URLBuilder.RelativeToRoot(buildBasePath(gvr, ns, ""))
+						perms[verb] = url
+					}
+				}
+				permissions[res] = perms
```
rohitsakala, line 178: If perms is empty, we add an empty map.

rohitsakala (changes requested):

pkg/resources/common/formatter.go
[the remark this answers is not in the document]
crobby: makes sense.  I like it

pkg/resources/common/formatter.go
crobby: Updated code to Eric's recommendation and added a comment to explain.

pkg/resources/common/formatter.go
crobby: thanks.  updated

pkg/resources/common/formatter.go
```diff
@@ -184,3 +222,24 @@ func excludeValues(request *types.APIRequest, unstr *unstructured.Unstructured)
...
+			if backingNamespace, ok := statusMap["backingNamespace"].(string); ok && backingNamespace != "" {
+				return backingNamespace
+			}
+		}
+	}
+
+	// Otherwise, if the id has a slash, we will interpret that.
+	// This is used to determine a project's namespace when there is no backingNamespace present.
+	// For cases where there is no slash, we use the object's ID, which is the same as the namespace.
+	return strings.Replace(obj.ID, "/", "-", 1)
```
rohitsakala, line 244: I hit this URl https://localhost:8443/v1/apps.deployments?checkPermissions=namespaces

and I get the following result 

Screenshot 2025-05-06 at 11 48 52 AM


There is no such namespace and the link also doesn't work. 

My URL is logically wrong, but when I put something logically wrong, it should not show these wrong links or `resourcePermissions` shouldn't be shown. Let me know your thoughts on this.

rohitsakala (changes requested):

pkg/resources/common/formatter.go
crobby: Interesting.  I did leave it intentionally open to being able to request things that might not make sense, but that is a surprising result.  I'll take a look to see if I can make it so it doesn't return those obviously incorrect resourcePermissions.  Definitely open to ideas if you have any.

crobby: Ah...I bet you were logged-in as admin.  For admin, the answer to any permission request is always, "yes".

crobby: full circle here..../v1/apps.deployments will give a list of all deployments......then the ensuing checks will all return true for admin.  Definitely not the case for normal users.

ericpromislow (approved):
LGTM

rohitsakala (approved):

PR #595, opened 2025-04-10
Log error from http client

rohitsakala:


pkg/ui/handler.go
```diff
@@ -166,8 +154,14 @@ func (u *Handler) IndexFile() http.Handler {
...
 func serveIndex(resp io.Writer, url string) error {
-	r, err := insecureClient.Get(url)
+	client := &http.Client{
+		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+		},
+	}
+	r, err := client.Get(url)
 	if err != nil {
+		logrus.Errorf("failed to download %s: %v", url, err)
```
tomleb, line 164: We already log the error here: https://github.com/rancher/steve/blob/da9bc2a11474d661f7c7213ff03d8896360eabce/pkg/ui/handler.go#L112 so instead of logging it twice, maybe best to just return it?

The error is ignored here: https://github.com/rancher/steve/blob/da9bc2a11474d661f7c7213ff03d8896360eabce/pkg/ui/handler.go#L161, we could add a logging statement there instead.

crobby:
Is there an issue that can be linked to this PR?

tomleb (approved):

pkg/ui/handler.go
```diff
@@ -158,15 +146,23 @@ func (u *Handler) IndexFileOnNotFound() http.Handler {
 func (u *Handler) IndexFile() http.Handler {
 	return u.indexMiddleware(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
 		if path, isURL := u.path(); isURL {
-			_ = serveIndex(rw, path)
+			err := serveIndex(rw, path)
+			if err != nil {
+				logrus.Errorf("failed to download %s: %v", path, err)
```
ericpromislow, line 151: `%v` should be `%w` to wrap the error object correctly

ericpromislow (changes requested):
Use %w, ...

ericpromislow (approved):
ok it's good

PR #602, opened 2025-04-23
Add support for modifying the created Factory clients

bigkevmcd:
This adds support for modifying the rest.Config used by the client factory.

Initial support for modifying the QPS/Burst values is provided.

### DIFF
--- a/pkg/client/factory.go
+++ b/pkg/client/factory.go
@@ -2,7 +2,10 @@ package client
 
 import (
 	"fmt"
+	"log"
 	"net/http"
+	"os"
+	"strconv"
 	"time"
 
 	"github.com/rancher/apiserver/pkg/types"
@@ -49,8 +52,7 @@ func (a *addQuery) WrappedRoundTripper() http.RoundTripper {
 
 func NewFactory(cfg *rest.Config, impersonate bool) (*Factory, error) {
 	clientCfg := rest.CopyConfig(cfg)
-	clientCfg.QPS = 10000
-	clientCfg.Burst = 100
+	updateConfigFromEnvironment(clientCfg)
 
 	watchClientCfg := rest.CopyConfig(clientCfg)
 	watchClientCfg.Timeout = 30 * time.Minute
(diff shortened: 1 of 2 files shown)


crobby (approved):
lgtm

ericpromislow (approved):
LGTM

bigkevmcd:
Per @tomleb I've moved the determination of the values out of steve (via functional-opts) which changes the branch significantly 😄

crobby (approved):
Still lgtm, thanks for the explanation.

pkg/client/factory.go
```diff
@@ -50,9 +54,41 @@ func (a *addQuery) WrappedRoundTripper() http.RoundTripper {
...
+		qps:   defaultQPS,
+		burst: defaultBurst,
+	}
+}
+
+// WithQPSAndBurst configures the rest.Config used for creating the clients in
+// the factory with the provided burst and qps configuration.
+//
+// See https://pkg.go.dev/k8s.io/client-go/rest#Config for more.
+func WithQPSAndBurst(qps float32, burst int) FactoryOption {
```
ericpromislow, line 73: With this commit I don't see anything outside test code calling this function. If it's intended to be called from `rancher/rancher/pkg/wrangler/context.go:NewContext` this increases the coupling between the `r/r/pkg/wrangler` code and the steve code.  The change in the first commit takes the config and figures out what to do -- I prefer that to this approach, but can always be convinced.

bigkevmcd: I really don't mind either way.

I'm not sure that it increases the coupling, it was already very tightly coupled (no matter what Rancher passes in as the values via the config, they get overwritten).

This reduces that coupling somewhat, allowing rancher (or any consumer) to configure the values that it wants to see.

But, I'm happy to revert, we want this to allow customers to configure how much we hammer their APIs...so, one way or another we need some sort of change.

bigkevmcd: We'd shift this to Rancher...let it decide what values (likely via the same route).

PR #607, opened 2025-04-24
 Refactor HelmData formatter to avoid importing helm dependencies

aruiz14:
- For Helm v3: just embed a raw JSON message as it's read from the encoded data, there is no need to interpret (unmarshal) it.
- For Helm v2: since it uses Protobuf binary encoding, decoding is not as simple as with JSON, as we would need the message descriptors. The simplest was to just copy the generated `pb.go` files from the legacy library and vendor them (also adapting the import paths).

Also, since the test code also relied on the helm library, I first refactored test code, ensuring it continued to pass on every case (before and after changes in the production logic).


aruiz14:
I need to run some more checks, since it panics registering types already registered.

ericpromislow (changes requested):
I see that `pkg/resources/formatters/internal/legacytypes/README.md` explains where all the protobuf files come from. Are there any other instructions needed?  I assume not, that these files are compiled from some spec and won't need to be changed as new versions of protobuf are released.  If this isn't true, pls treat the `Approve` as a `Request Changes`.

Ignoring what's going on in the protobuf files, this is a straightforward PR.

ericpromislow (approved):
Looks great

raulcabello (approved):
LGTM.  I agree that we won’t need to regenerate the .pb.go files as Helm v2 won't change.

PR #610, opened 2025-04-24
Sort-indirect PR broken into smaller parts: part 1/6 - put common types in their own package

ericpromislow:

The problem having these in the informer package is that eventually code in other packages will need to import `informer` only for constants or types, but some members of the informer package may already depend on those. Best to move type definitions into their own simpler package.

This is a replacement for PR 600 -- this is the first of at least 5 cascading PRs.

This one is large (+466 -450) but mechanical.  Moving types into its own `pkg/sqlcache/sqltypes/types.go` package.


crobby (approved):
Thanks for splitting this up.
This part looks good, straightforward.

tomleb (approved):

PR #612, opened 2025-04-24
Sort-indirect PR broken into smaller parts: part 3/6 - pass listOptions by reference

ericpromislow:
# DO NOT MERGE (depends on PR #611 )

The `ListOptions` structure gets passed to a lot of functions (with more on the way),
so pass it by reference instead of making copies.

### DIFF
--- a/pkg/sqlcache/informer/informer.go
+++ b/pkg/sqlcache/informer/informer.go
@@ -30,7 +30,7 @@ type Informer struct {
 }
 
 type ByOptionsLister interface {
-	ListByOptions(ctx context.Context, lo sqltypes.ListOptions, partitions []partition.Partition, namespace string) (*unstructured.UnstructuredList, int, string, error)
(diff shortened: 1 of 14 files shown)


crobby (approved):
lg

tomleb (approved):
I'll approve because it's pretty mechanical and I don't want to block the rest but I'd be curious what the concrete benefit of this would be. Do we expect some kind of performance improvement from the lack of copying? Not a fan of passing pointers because we can now mutate that fields anywhere it's passed in.

ericpromislow:
My analysis is that in the original Vai work there weren't many functions and this structure wasn't getting copied repeatedly. But now it is.

I'm not sure about the benefits, but the Manning book 100 mistakes go programmers make include inadvertent copying of structures as one of them. In their example the error was where you do 
```
for i, x := range []struct foo{....} {
  // x is a copy of foo[i], not a reference
}
```

and I wondered if we should avoid repeated copying when passing parameters.

PR #614, opened 2025-04-24
Sort-indirect PR broken into smaller parts: part 5/6 - generate the SQL for indirect queries

ericpromislow:
# DO NOT MERGE
## depends on PR 613

The key file here is `pkg/sqlcache/informer/query_generator.go` but the accompanying tests should give a reasonable idea of what's going on. It looks like a huge PR, but much of the size is due to moving code from `listoption_indexer*.go` to `query_generator*.go` .


crobby (approved):
The comprehensive tests definitely help clarify things here.  I don't see anything that stands out as definitely wrong.  lgtm

PR #615, opened 2025-04-25
Sort-indirect PR broken into smaller parts: part 6/6 - verify the generated SQL

ericpromislow:
# DO NOT MERGE (depends on PR #614 )

It's all tests here - the database data is in the fixtures/ dir in the form of sql statements, because that was the easiest way to get data into the database.

If you want to mess with the database you need to comment out the `os.Remove` statement in `teardownTests` and add a debug stmt giving the path to the tempfile.


pkg/sqlcache/informer/verify_generator_test.go
```diff
@@ -0,0 +1,667 @@
...
+				}
+			}
+		})
+	}
+}
+
+func TestMain(m *testing.M) {
+	err := setupTests()
+	if err != nil {
+		panic(fmt.Sprintf("Awp! verify_generator_test.go tests failed to setup: %s", err))
```
crobby, line 607: panic intententional?

ericpromislow: Given that `TestMain` doesn't return an error value, or anything, I assumed a panic was called for.

crobby (approved):

ericpromislow:
We're not taking this approach with an indirect-association operator. Instead indirect associations will be hardwired in the code using a table-driven approach.

PR #616, opened 2025-04-29
Sort-indirect PR broken into smaller parts: part 7/6 - whitelist external fields

ericpromislow:
# DO NOT MERGE (depends on PR #615 )

Use the whitelist data in proxy_store to verify access to the specified external fields.

This assumes that if the user has read access to table T1, and T1 is linked to T2, they also have 
read access to T2.


pkg/sqlcache/informer/query_generator.go
```diff
@@ -940,40 +939,54 @@ func (l *ListOptionIndexer) validateColumn(column string) error {
...
 
+func (l *ListOptionIndexer) validateDirective(isd *sqltypes.Sort) error {
+	if !isd.IsIndirect {
+		return nil
+	}
+	if len(isd.IndirectFields) != 4 {
+		return fmt.Errorf("expected indirect sort directive to have 4 indirect fields, got %d", len(isd.IndirectFields))
+	}
+	g, v, k := getGVKFieldsFromIndirectFields(isd.IndirectFields)
+	//gvk := makeGVK(isd.IndirectFields[0], isd.IndirectFields[1])
```
crobby, line 970: stray commented out code?

ericpromislow: Yes. That was a scaffold on the way to figuring out the minimal support needed to get the test code to access schemas.

ericpromislow:
We're not taking this approach with an indirect-association operator. Instead indirect associations will be hardwired in the code using a table-driven approach.

PR #617, opened 2025-04-30
[main] Integrate Imperative api into steve

joshmeranda:
# Issue

https://github.com/rancher/rancher/issues/47033

Adds new routes to the steve routing table, and infrastructure to allow merging the merges of two `http.Handler` objects.

Due to the size of the openapiv2 swagger data and the need to reconcile the data from the extension api server, requests to `/v1/openapiv2` are very slow (~1.5s). In most situations this won't be a problem since we don't need to reconcile the two responses if the extension api is enabled.


pkg/server/router/merge/merge.go
```diff
@@ -0,0 +1,215 @@
...
+		data, err := json.Marshal(merged)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal merged object: %w", err)
+		}
+
+		return data, nil
+	}
+}
+
+var APIGropuListMerger = typedMergeFunc(func(lhs metav1.APIGroupList, rhs metav1.APIGroupList) (metav1.APIGroupList, error) {
```
ericpromislow, line 113: I assume you mean `APIGroupListMerger`

pkg/server/router/merge/merge.go
```diff
@@ -0,0 +1,215 @@
...
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	if _, err := w.Write(merged); err != nil {
+		http.Error(w, fmt.Sprintf("failed to write response: %s", err), http.StatusInternalServerError)
+		return
+	}
+
+	// todo: write status
```
ericpromislow, line 213: Is there another issue for these TODOs?

pkg/server/router/merge/merge_openapi_v2.go
```diff
@@ -0,0 +1,142 @@
...
+func mergeSlicesOnValue[T comparable, C comparable](lhs, rhs []T, getter func(T) C) []T {
+	if len(lhs) == 0 {
+		return rhs
+	}
+
+	if len(rhs) == 0 {
+		return lhs
+	}
+
+	for _, t := range lhs {
```
ericpromislow, line 20: Shouldn't this be `for _, t := range rhs {` , leaving the rest of the loop as is? Otherwise it needs an explanatory comment. Right now this loop does nothing (unless `getter(x) => rand.Int()`)

pkg/server/router/merge/merge.go
joshmeranda: Oops, had these cleaned up in another branch that I didn't bring over. Should be good now!

pkg/server/router/merge/merge_openapi_v2.go
joshmeranda: Yep, good catch thanks!

joshmeranda:
@ericpromislow resolved most of those. I'll hold off on writing tests until we see what happens with https://github.com/SUSE/rancher-architecture/pull/14 since that would make this whole PR unnecessary and the tests for this will be  hard to write.

ericpromislow (changes requested):
Looks good. Would be good if there were unit tests exercising all (or most of) the possible paths of the merging functions, but I understand putting that off, and the code looks like it's implementing the intention.

PR #626, opened 2025-05-06
Generate SQL for searching in explicit arrays - part 1/4 (est) - generate SQL when we have a `Contains` op

ericpromislow:
The syntax will be `FIELD CONTAINS STRING`.

`FIELD` can either be a shorthand for `FIELD[1], FIELD[2], etc...` and we're looking for the row
where one of those values can be our search-string `STRING`.

Or `FIELD` can be a string representing a string array where the values are separated by or-bars (`"|"`)
and we want the row where one of those parts is `STRING`. We do this search by surrounding both
the `FIELD` and `STRING` with or-bars and looking for a substring match. This handles matching at
the start or end of `FIELD` as well as internally.

### DIFF
--- a/pkg/sqlcache/informer/listoption_indexer_test.go
+++ b/pkg/sqlcache/informer/listoption_indexer_test.go
@@ -1106,6 +1106,31 @@ func TestConstructQuery(t *testing.T) {
 		expectedErr:      nil,
 	})
 	tests = append(tests, testCase{
+		description: "TestConstructQuery: handles NOT-IN statements",
+		listOptions: sqltypes.ListOptions{Filters: []sqltypes.OrFilter{
+			{
+				[]sqltypes.Filter{
+					{
+						Field:   []string{"metadata", "queryField1"},
+						Matches: []string{"somevalue"},
+						Op:      sqltypes.NotIn,
+					},
+				},
+			},
+		},
+		},
+		partitions: []partition.Partition{},
+		ns:         "",
+		expectedStmt: `SELECT o.object, o.objectnonce, o.dekid FROM "something" o
+  JOIN "something_fields" f ON o.key = f.key
+  WHERE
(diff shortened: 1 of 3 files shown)


pkg/sqlcache/informer/listoption_indexer_test.go
```diff
@@ -1681,6 +1706,146 @@ func TestConstructQuery(t *testing.T) {
...
+					{
+						Matches: []string{"too", "many", "targets"},
+						Field:   []string{"metadata", "queryField1"},
+						Op:      sqltypes.Contains,
+					},
+				},
+			},
+		},
+		},
+		//partitions: []partition.Partition{},
```
crobby, line 1785: stray comment?  Same appears 2x below as well.

crobby (approved):
lg.  Just a couple of stray comments.

pkg/sqlcache/informer/listoption_indexer_test.go
ericpromislow: Yeah, that was me testing to see if those lines are no longer needed.

ericpromislow:
This revised PR drops the `CONTAINS` operator.  Now if you're attempting a test like `filter=FIELD = VALUE`, `FIELD` isn't indexed, but `FIELD[1], FIELD[2], ... FIELD[n]` are, I generate code that tries to match on all those values. Supported operators: `==`, `!=`, `~`, `!~`, `<`, `>`, `IN`, and `NOT IN`.

Also there's no array-type support for single strings that use or-bars as separators. You can use the `~` partial match operator to test those.

I didn't squash all the previous commits (yet).  This PR is now best done by doing a diff against `main`. But if it helps, I'll squash them all now, during review.

gehrkefc (approved):

crobby (approved):
lg

PR #627, opened 2025-05-07
Generate SQL for searching in explicit arrays - part 2/4 (est) - generate SQL when we have a Contains op on labels

ericpromislow:
## DO NOT MERGE

Depends on PR [#626](https://github.com/rancher/steve/pull/626)

Related to  [#49116](https://github.com/rancher/rancher/issues/49116)

Code gen for labels is slightly different than for non-label fields. Note also that we never assume there are array-like
labels with names `foo[1]`, `foo[2]`, etc.. Unlike internal fields, we don't know the names of the label fields in advance. So we only do `|`-split strings as arrays.


crobby (approved):
I'm still no SQL guru, but the tests do a nice job of showing what is expected and the actual code is pretty straightforward.

ericpromislow:
We've decided against a `CONTAINS` operator. Instead when doing an operation against a field known to be an array of values we'll generate the SQL for an implicit `CONTAINS` operator - this is done in a later PR.

PR #637, opened 2025-05-12
[main] Use Wrangler defined field for resourcePermissions, bump wrangler

crobby:
Issue:  https://github.com/rancher/rancher/issues/45827

Moving the resourcePermissions out of resourceFields.

The new version of wrangler defines the field to be its own top level entity.

This should fix the problem where the dashboard was trying to include resourcePermissions when creating objects like configmaps, among other issues.


rohitsakala (approved):

brandond (approved):

ericpromislow (approved):
Questions are answered

pkg/schema/factory.go
```diff
@@ -145,11 +145,11 @@ func (c *Collection) schemasForSubject(access *accesscontrol.AccessSet) (*types.
...
 		attributes.SetAccess(s, verbAccess)
 
-		if s.ResourceFields == nil {
-			s.ResourceFields = make(map[string]schemas.Field)
+		if s.Attributes == nil {
+			s.Attributes = make(map[string]interface{})
 		}
-		s.ResourceFields["resourcePermissions"] = schemas.Field{
-			Type:        "map[json]",
+		s.Attributes["resourcePermissions"] = schemas.Field{
```
brandond, line 151: You stated your intent to make `resourcePermissions` peer with `resourceMethods` and `resourceFields`, but this appears to instead be moving it under `attributes` instead of under `resourcePermissions` - where it clearly did not belong.

Are we just randomly picking a place in the schema for some new data? Where does this actually **BELONG**?

brandond (dismissed):

pkg/schema/factory.go
crobby: Fair point....
I should have been more clear.
This move puts resourcePermissions on the same level as "verbs" under attributes, which I think makes the most sense since its purpose is similar.

brandond: Has there been any higher-level discussion about where this field belongs? What is consuming this new field? I don't see that it is getting populated anywhere - this sets the type and description but where is it actually used?

brandond: ok but it sounds like that didn't address where specifically to put the field?

crobby: I've been in contact with @richard-cox for work on the epic, we've walked through things.  It was entirely my mistake for putting it under resourceFields at first and we didn't catch it until now.  I misunderstood the purpose of them.

brandond: OK. So stuffing this into the a key under the `Attributes map[string]interface{}` is the correct thing to do here, instead of making it a new peer to the other `Resource*` fields?

https://github.com/rancher/wrangler/blob/a1fbf41856b2cc65bf4f644b2b1a5d4fd88a24be/pkg/schemas/types.go#L10-L12

crobby: It currently contains other information related to things that the current user can/can't do.  disallowMethods, and verbs are other bits of information that are currently in the attributes field.

brandond: It looks like other things that interact with Attributes fields are all done with helpers in `attributes.go`:
https://github.com/rancher/steve/blob/ea01a40c8ddfb1d988501aa64c3d19911202d7e7/pkg/attributes/attributes.go#L166-L186

Should this perhaps be done via `SetResourcePermissions`/`ResourcePermissions` helper functions?

crobby: Sure, I suppose that would be a good way to go. Thanks.

crobby: Upon further review (and talking with Richard), it sounds like the most proper way forward here is to define a new field in Wrangler that is specific to resource permissions.  While attributes does indeed have some other related information, it sounds like the intended original purpose of the attributes field was more related to CRD columns.  Rather than add more "things" that run contrary to the original purpose, we will take care of this at the wrangler level. 

So this PR will become "remove the manual addition of the schema field and bump wrangler".
Set back to draft for now until the wrangler bump is ready.

crobby: @brandond Any other thoughts on this?

brandond: I am 👍🏻 on moving it to a proper field in `schemas.Schema`

brandond (approved):

PR #639, opened 2025-05-13
Generate SQL for searching in explicit arrays - part 2/2 - Verify parsed queries and generated SQL works correctly

ericpromislow:
# DO NOT MERGE

Depends on #626 

Related to [#49116](https://github.com/rancher/rancher/issues/49116)

The idea in this PR is to set up a database (of foods, but there are tables for other rancher-like things that
will get used in a later PR), make URL-like queries on them, translate the queries into SQL, run the SQL through
sqlite, and verify that we get the expected results.


ericpromislow:
We're taking a different approach for array-based searching. Not doing `metadata.fields`.  Different PR handles array-based searching for things like `spec.containers.image` where it's a single string field with `|`-separated substrings.

PR #641, opened 2025-05-14
Tweak array-indexing

ericpromislow:
# DO NOT MERGE

From discussion with Richard Cox we don't want the feature where a test like `metadata.fields == SOME-VALUE" will match any entry in `metadata.fields[0]`, `metadata.fields[1]`, `metadata.fields[2]`. For example, if you have an object where `metadata.fields = [true, false]` then both tests `metadata.fields == true` and `metadata.fields == false` will select that row, but that might select the row based on the wrong member of the `metadata.fields` array (giving a literal false positive if you're searching for `true` with `metadata.fields[1]` in mind, not `[0]`.

I think I'll move the updated README to a separate PR and close this issue.

Related to [#49116](https://github.com/rancher/rancher/issues/49116)

Updating the main steve readme to reflect the state of Sqlite-based caching and searching

I also updated a bit of code to avoid storing duplicate values when mapping multi-value string arrays
to a single string field in the database (the values are separated by "|", a character that isn't allowed
in the fields we index).

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,9 @@ Example, filtering by object name:
 /v1/{type}?filter=metadata.name=foo
 ```
 
-if a target value is surrounded by single-quotes, it succeeds only on an exact match:
+When SQLite caching is not enabled, matching works this way:
+
+If a target value is surrounded by single-quotes, it succeeds only on an exact match:
 
 Example, filtering by object name:
 
@@ -132,6 +134,11 @@ Example, filtering by object name:
 ```
 /v1/{type}?filter=metadata.name="can-be-a-substri"
 ```
+
+When SQLite caching is enabled, there are different operators.  `FIELD = VALUE` does exact matching,
+`FIELD ~ VALUE` does partial matching. You can use quotes around `VALUE` to delimit it; if it contains
+only alphanumeric characters and characters that aren't significant in filters (like '=', ',', etc.),
+they don't need to be quoted.
 
 One filter can list multiple possible fields to match, these are ORed together:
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -133,6 +135,11 @@ Example, filtering by object name:
 /v1/{type}?filter=metadata.name="can-be-a-substri"
 ```
 
+When SQLite caching is enabled, there are different operators.  `FIELD = VALUE` does exact matching,
+`FIELD ~ VALUE` does partial matching. You can use quotes around `VALUE` to delimit it; if it contains
+only alphanumeric characters and characters that aren't significant in filters (like '=', ',', etc.),
+they don't need to be quoted.
+
```
crobby, line 142: Maybe an example of this would help.  I'm sure it's correct, but I'm not totally clear.

crobby (approved):
Seems good.
Is it worthwhile to add a test to make sure the uniqueness bit works?

One other question/comment on the docs.

README.md
ericpromislow: I added examples for all the operators when sql is enabled. Might as well doc them here, and then the docs team can run with it if they so desire.

ericpromislow:
And it was definitely worthwhile adding a test -- it uncovered an existing bug in the underlying code.

tomleb:
@ericpromislow I've got a fix here https://github.com/rancher/steve/pull/642 for array indexing  when the value doesn't exist. Basically:

```go
						itemStr, ok := itemVal[subField].(string)
						if !ok {
							return nil, fmt.Errorf(failedToGetFromSliceFmt, subField, err)
						}
						result[index] = itemStr
```

we don't check that `itemVal[subField]` exist, so we return an error if it's not a string. What we want is to return a default value (eg: empty string). There's a tiny bit more context in the PR.

Do you want to include this in your PR or leave it separate? I'm open to both.

ericpromislow:
We're not doing this for `metadata.fields`, and have a different approach for `|`-separated substring fields

PR #642, opened 2025-05-15
Fix returning error for empty ingresses.spec.rules.host

tomleb:
# Issue https://github.com/rancher/rancher/issues/50309

Basically, for an ingress like this:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: aaa
spec:
  rules:
  - 
```
with no "host" defined (yeah..) we weren't able to insert it in the DB so the UI isn't able to fetch it either.

### DIFF
--- a/pkg/sqlcache/informer/listoption_indexer.go
+++ b/pkg/sqlcache/informer/listoption_indexer.go
@@ -905,11 +905,17 @@ func getField(a any, field string) (any, error) {
 					if !ok {
 						return nil, fmt.Errorf(failedToGetFromSliceFmt, subField, err)
 					}
-					itemStr, ok := itemVal[subField].(string)
-					if !ok {
-						return nil, fmt.Errorf(failedToGetFromSliceFmt, subField, err)
+
+					_, found := itemVal[subField]
+					if found {
+						itemStr, ok := itemVal[subField].(string)
+						if !ok {
+							return nil, fmt.Errorf(failedToGetFromSliceFmt, subField, err)
+						}
+						result[index] = itemStr
+					} else {
+						result[index] = ""
 					}
-					result[index] = itemStr
 				}
 				return result, nil
 			}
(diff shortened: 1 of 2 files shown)


pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -905,11 +905,17 @@ func getField(a any, field string) (any, error) {
...
 					}
-					itemStr, ok := itemVal[subField].(string)
-					if !ok {
-						return nil, fmt.Errorf(failedToGetFromSliceFmt, subField, err)
+
+					_, found := itemVal[subField]
+					if found {
+						itemStr, ok := itemVal[subField].(string)
+						if !ok {
+							return nil, fmt.Errorf(failedToGetFromSliceFmt, subField, err)
```
ericpromislow, line 913: `err` is null. Please drop the `%w` from `failedToGetFromSliceFmt` and remove the `err` args where it's used.

The rest is fine.

pkg/sqlcache/informer/listoption_indexer_test.go
```diff
@@ -1790,3 +1790,163 @@ func TestBuildSortLabelsClause(t *testing.T) {
...
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			result, err := getField(test.obj, test.field)
+			if test.expectedErr {
+				assert.Error(t, err)
+				return
+			}
+
+			assert.NoError(t, err)
```
ericpromislow, line 1948: I'd prefer `require.NoError` so if there is an error we don't try the test on the next line.

ericpromislow (changes requested):
Just a couple of minor requests

pkg/sqlcache/informer/listoption_indexer.go
tomleb: Good catch! Fixed.

pkg/sqlcache/informer/listoption_indexer_test.go
tomleb: Fixed

tomleb:
@crobby we decided to keep going with this one as separate fyi, if you could give another look, thanks!

crobby (approved):

ericpromislow (approved):
LGTM

PR #643, opened 2025-05-22
Remove limit and continue param for SQL cache

tomleb:

Part 1 of many PRs

**Difference between `page`/`pagesize` and `limit`/`continue`**

In K8s, `limit` & `continue` can be used to slow fetch a list of a specific resourceVersion in chunks. The `continue` param acts as a "cursor".

In Steve, `page`/`pagesize` also chunks the response but gives us that chunk at the time of the request, so no "cursor" functionality.

**Use case**

For Steve with SQL cache enabled, we no longer support the `limit` and `continue` param. There is going to be 2 use cases (For now anyway):
1. Getting a list for paginated page: In that case, we're interested in the "latest state" so we don't care about the "cursor" behavior.
2. Getting full lists: This is only used for _some_ resources, and we're okay with not supporting chunking here. The UI will simply get back the full list. Example: navlinks.

### DIFF
--- a/pkg/sqlcache/informer/listoption_indexer.go
+++ b/pkg/sqlcache/informer/listoption_indexer.go
@@ -423,29 +423,18 @@ func (l *ListOptionIndexer) constructQuery(lo *sqltypes.ListOptions, partitions 
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


ericpromislow (approved):
My favorite kind of PR, one that just removes code.

crobby (approved):

PR #647, opened 2025-05-29
Calculate AccessSets once per request instead of per resource

aruiz14:
Fixes https://github.com/rancher/rancher/issues/50484

# Analysis

As mentioned in the parent issue, we used a new Rancher 2.11.2 installation to create 500 Projects (3 member users and 2 namespaces on each), using a single user (the default admin), which resulted in this user being referenced by ~4000 RoleBinding resources (both within the Projects' top-level and child namespaces).

We started observing up to 17s response times, just listing namespaces under `/v1/namespaces`.
CPU profiling showed a considerable amount of the time was spent calculating `AccessSet` for the user, especially listing RoleBindings (as they were much more numerous compared to ClusterRoleBindings), followed by GC:
Screenshot 2025-05-29 at 17 35 42

Adding some debug logs and attaching a debugger I realized it was not a matter of that code being resource intensive, but `AccessFor(user)` was being called >1500 times for every single request, which then makes the problem obvious, as the caller is a Formatter, meaning that the accessSet is being calculated for every single object being printed.

# Solution
Since a request-level AccessSet is already being calculated while processing the Schemas, and stored as attribute, I changed the formatter to use that if available (falling back to the previous behavior otherwise)

# Result

Speedup is really noticeable just by browsing through the UI. The diagnostic package shows approximately x15 speedup in response times:
 - Before:
Screenshot 2025-05-29 at 16 26 16
 - After:
Screenshot 2025-05-29 at 16 29 04

### DIFF
--- a/pkg/accesscontrol/access_control.go
+++ b/pkg/accesscontrol/access_control.go
@@ -55,6 +55,9 @@ func SetAccessSetAttribute(schemas *types.APISchemas, accessSet *AccessSet) {
 }
 
 func AccessSetFromAPIRequest(req *types.APIRequest) *AccessSet {
+	if req == nil || req.Schemas == nil {
+		return nil
+	}
 	if v, ok := req.Schemas.Attributes[accessSetAttribute]; ok {
 		return v.(*AccessSet)
 	}
--- a/pkg/accesscontrol/access_control.go
+++ b/pkg/accesscontrol/access_control.go
@@ -28,7 +28,7 @@ func (a *AccessControl) CanDo(apiOp *types.APIRequest, resource, verb, namespace
 	}
 	group, resource := kv.Split(resource, "/")
 	accessSet := AccessSetFromAPIRequest(apiOp)
-	if accessSet.Grants(verb, schema.GroupResource{
+	if accessSet != nil && accessSet.Grants(verb, schema.GroupResource{
 		Group:    group,
 		Resource: resource,
 	}, namespace, name) {
--- a/pkg/accesscontrol/access_control.go
+++ b/pkg/accesscontrol/access_control.go
@@ -47,6 +47,7 @@ func (a *AccessControl) CanWatch(apiOp *types.APIRequest, schema *types.APISchem
 	return a.SchemaBasedAccess.CanWatch(apiOp, schema)
 }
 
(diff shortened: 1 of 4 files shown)


pkg/accesscontrol/policy_rule_index.go
```diff
@@ -181,8 +181,9 @@ func (p *policyRuleIndex) getRoleBindings(subjectName string) []*rbacv1.RoleBind
 
 // getRoleRefs gathers rules from roles granted to a given subject through RoleBindings and ClusterRoleBindings
 func (p *policyRuleIndex) getRoleRefs(subjectName string) subjectGrants {
-	var clusterRoleBindings []roleRef
-	for _, crb := range p.getClusterRoleBindings(subjectName) {
+	crbs := p.getClusterRoleBindings(subjectName)
+	clusterRoleBindings := make([]roleRef, 0, len(crbs))
+	for _, crb := range crbs {
```
aruiz14, line 186: This is a small optimization, as I also observed `growSlice` taking a fraction of the CPU.

bigkevmcd: If you're optimising for performance, this can use assignment rather than appending...

A very quick benchmark suggests that it should be around 5% faster 😄

pkg/accesscontrol/access_control.go
```diff
@@ -44,3 +46,20 @@ func (a *AccessControl) CanWatch(apiOp *types.APIRequest, schema *types.APISchem
 	}
 	return a.SchemaBasedAccess.CanWatch(apiOp, schema)
 }
+
+func SetAccessSetAttribute(schemas *types.APISchemas, accessSet *AccessSet) {
```
bigkevmcd, line 50: How about adding doc comments for these?

pkg/schema/factory.go
```diff
@@ -171,9 +171,7 @@ func (c *Collection) schemasForSubject(access *accesscontrol.AccessSet) (*types.
 		}
 	}
 
-	result.Attributes = map[string]interface{}{
-		"accessSet": access,
-	}
+	accesscontrol.SetAccessSetAttribute(result, access)
```
bigkevmcd, line 174: 👍🏻

bigkevmcd (approved):
Looks good!

PR #648, opened 2025-05-29
#48673 - Added Timestamp Cache Handling to metadata.fields

gehrkefc:
This PR is related to:

https://github.com/rancher/rancher/issues/48673


pkg/resources/virtual/virtual.go
```diff
@@ -37,6 +39,35 @@ func (t *TransformBuilder) GetTransformFunc(gvk schema.GroupVersionKind) cache.T
...
+				curValue, got, err := unstructured.NestedSlice(obj.Object, "metadata", "fields")
+				if !got {
+					return obj, nil
+				}
+
+				if err != nil {
+					return nil, err
+				}
+
+				duration, err := time.ParseDuration(curValue[index].(string))
```
tomleb, line 56: `time.ParseDuration` fails to parse something like `3d7h`:

```
E0529 21:05:14.760924   24656 reflector.go:166] "Unhandled Error" err="/home/tlebreux/.cache/go/mod-cache/k8s.io/client-go@v0.32.2/tools/cache/reflector.go:251: Failed to watch management.cattle.io/v3, Kind=RoleTemplate: unable to sync list result: couldn't enqueue object: NOPE: time: unknown unit \"d\" in duration \"3d7h\"" logger="UnhandledError"
E0529 21:05:15.600158   24656 reflector.go:166] "Unhandled Error" err="/home/tlebreux/.cache/go/mod-cache/k8s.io/client-go@v0.32.2/tools/cache/reflector.go:251: Failed to watch management.cattle.io/v3, Kind=RoleTemplate: unable to sync list result: couldn't enqueue object: NOPE: time: unknown unit \"d\" in duration \"3d7h\"" logger="UnhandledError"
``` 

Here's the k8s code that translates a time.Duration to this "human-readable duration":
- https://github.com/kubernetes/apimachinery/blob/59e9003f02d6f0c8fff53719a7a0604ec82ee9a9/pkg/util/duration/duration.go#L48

Unfortunately, it looks like the library doesn't have the reverse way, which means we'll have to implement our own and add tests to it.

Hummmm...

pkg/resources/common/duration.go
```diff
@@ -0,0 +1,44 @@
+package common
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"time"
+)
+
+var humanReadableDurationRe = regexp.MustCompile(
```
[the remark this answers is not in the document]
gehrkefc: Just fixed it. Thanks!

ericpromislow (commented):
I don't know the history of this code. Is it possible that some incoming strings still use the old full names (like "3days") and others use only the new ones?  Or during some version bump of k8s it moved from the full-name form to the single-letter form and we didn't adapt?

tomleb:
> I don't know the history of this code. Is it possible that some incoming strings still use the old full names (like "3days") and others use only the new ones? Or during some version bump of k8s it moved from the full-name form to the single-letter form and we didn't adapt?

This is an issue introduced by Vai and us keeping the column values in the cache. Pre-Vai doesn't have this issue because users making a request is proxied to Kubernetes, so you get something that's pretty up to date. 

Also, the issue description and comments contains a lot of information on why we need to do this.

gehrkefc:
@ericpromislow I'm based on this function from k8s: 

https://github.com/kubernetes/apimachinery/blob/v0.33.1/pkg%2Futil%2Fduration%2Fduration.go#L48

From what I understand, it generates only one letter representations.

pkg/resources/common/util.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+// GetIndexValueFromString looks for values between [ ].
+// e.g: $.metadata.fields[2], in this case it would return 2
+// In case it doens't find any value between brackets it
+// returns -1
+func GetIndexValueFromString(pathString string) int {
+	index := -1
+	idxStart := strings.Index(pathString, "[")
+	idxEnd := strings.Index(pathString, "]")
```
ericpromislow, line 18: This should be more like:
```
idxStart := strings.Index(pathString, "[")
if idxStart == -1 { return -1 }
idxEnd := strings.Index(pathString[idxStart+1:], "]")
if idxEnd == -1 return -1
idxEnd += idxStart + 1
idx, err := strconv.Atoi(pathString[idxStart+1 : idxEnd])
…
```

pkg/resources/common/util.go
```diff
@@ -0,0 +1,40 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/rancher/apiserver/pkg/types"
+	"github.com/rancher/steve/pkg/attributes"
+)
+
+// GetIndexValueFromString looks for values between [ ].
+// e.g: $.metadata.fields[2], in this case it would return 2
+// In case it doens't find any value between brackets it
```
ericpromislow, line 13: typo: "doesn't"

pkg/resources/common/formatter.go
```diff
@@ -212,6 +220,47 @@ func excludeFields(request *types.APIRequest, unstr *unstructured.Unstructured)
 	}
 }
 
+func convertMetadataFields(request *types.APIRequest, gvk schema2.GroupVersionKind, unstr *unstructured.Unstructured) {
```
ericpromislow, line 223: This function needs a comment explaining what it's supposed to do.

pkg/resources/common/formatter.go
```diff
@@ -212,6 +220,47 @@ func excludeFields(request *types.APIRequest, unstr *unstructured.Unstructured)
...
+				if err != nil {
+					logrus.Errorf("failed to get metadata.fields slice from unstr.Object: %s", err.Error())
+				}
+
+				if !got {
+					logrus.Debugf("couldn't find metadata.fields at unstr.Object")
+					return
+				}
+
+				millis, err := strconv.ParseInt(curValue[index].(string), 10, 64)
```
ericpromislow, line 245: I think your test data isn't old enough, because I'm running into errors with some apps I created this morning:
```
ERRO[0020] failed to convert timestamp value: strconv.ParseInt: parsing "6h7m": invalid syntax 
```

And when I read the code, why is this code not converting strings like `6h` into a numeric time and why are you writing the values back as human durations and not absolute timestamps?

It looks like the converter you wrote in virtual/virtual.go isn't kicking in.

ericpromislow (changes requested):
Some typos, but either I'm not understanding something or the transformer isn't working for me.

pkg/resources/common/gvkdatefields.go
```diff
@@ -0,0 +1,62 @@
+package common
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var DateFieldsByGVKBuiltins = map[schema.GroupVersionKind][]string{
```
ericpromislow, line 7: Tables like these should be in alphabetical order so they're easier to read and modify. Should. be straightforward to do this -- I have sort-lines capability in jetbrains, emacs, and vim

ericpromislow:
Also when I run without `--sql-cache`, I get this error message for each row getting processed:
```
ERRO[0006] failed to convert timestamp value: strconv.ParseInt: parsing "29s": invalid syntax 
```

pkg/resources/common/util.go
gehrkefc: Done. Thank you!

pkg/resources/virtual/virtual.go
```diff
@@ -37,6 +40,44 @@ func (t *TransformBuilder) GetTransformFunc(gvk schema.GroupVersionKind) cache.T
 	}
 	converters = append(converters, t.defaultFields.TransformCommon)
 
+	// Detecting if we need to convert date fields
```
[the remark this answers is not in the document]
gehrkefc: Done. Thank you!

pkg/resources/virtual/virtual.go
```diff
@@ -37,6 +40,44 @@ func (t *TransformBuilder) GetTransformFunc(gvk schema.GroupVersionKind) cache.T
...
+	for _, col := range columns {
```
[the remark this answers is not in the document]
gehrkefc: Thank you!

pkg/resources/common/gvkdatefields.go
gehrkefc: Sorted in alphabetical order and added a space to divide each group. Thank you!

pkg/resources/common/formatter.go
gehrkefc: Sure. Added docstring. Thanks!

pkg/resources/common/util.go
gehrkefc: Fixed it. Thanks!

pkg/resources/common/formatter.go
```diff
@@ -157,6 +164,7 @@ func formatter(summarycache common.SummaryCache, asl accesscontrol.AccessSetLook
 			includeFields(request, unstr)
 			excludeFields(request, unstr)
 			excludeValues(request, unstr)
+			convertMetadataFields(request, gvk, unstr)
```
ericpromislow, line 167: First, this function should have a more specific name, as it only converts timestamps in metadata fields.

Second, it should only be called when we're in sql-cache mode, because transforms are set up only in sql-cache mode (I'm not sure why we aren't doing it for non-vai processing). But that info is long-gone -- it's picked up when the server is created, and then dropped.

Here's the diff I have that stops complaining (and even trying to convert values) when it looks like we're not in sqlcache mode:

```
diff --git a/pkg/resources/common/formatter.go b/pkg/resources/common/formatter.go
index edb6cf7..5a7ac44 100644
--- a/pkg/resources/common/formatter.go
+++ b/pkg/resources/common/formatter.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"net/http"
+	"regexp"
 	"slices"
 	"strconv"
 	"strings"
@@ -27,6 +28,9 @@ import (
 	"k8s.io/apimachinery/pkg/util/duration"
 )
 
+var humanTimeRegex = regexp.MustCompile(`^(\d+[dhms])+$`)
+var inSqlMode = false
+
 func DefaultTemplate(clientGetter proxy.ClientGetter,
 	summaryCache *summarycache.SummaryCache,
 	asl accesscontrol.AccessSetLookup,
@@ -164,7 +168,9 @@ func formatter(summarycache common.SummaryCache, asl accesscontrol.AccessSetLook
 			includeFields(request, unstr)
 			excludeFields(request, unstr)
 			excludeValues(request, unstr)
-			convertMetadataFields(request, gvk, unstr)
+			if !inSqlMode {
+				convertMetadataFields(request, gvk, unstr)
+			}
 		}
 
 		if permsQuery := request.Query.Get("checkPermissions"); permsQuery != "" {
@@ -246,9 +252,21 @@ func convertMetadataFields(request *types.APIRequest, gvk schema2.GroupVersionKi
 					return
 				}
 
-				millis, err := strconv.ParseInt(curValue[index].(string), 10, 64)
+				timeValue, ok := curValue[index].(string)
+				if !ok {
+					logrus.Debugf("time field isn't a string")
+					return
+				}
+				millis, err := strconv.ParseInt(timeValue, 10, 64)
 				if err != nil {
-					logrus.Warnf("failed to convert timestamp value: %s", err.Error())
+					if humanTimeRegex.MatchString(timeValue) {
+						// If we're running in non-sqlcache mode, the human-friendly timestamps
+						// don't get converted to timestamps, so there's no point complaining
+						// that we can't process them here.
+						inSqlMode = true
+					} else {
+						logrus.Warnf("failed to convert timestamp value: %s", err.Error())
+					}
 					return
 				}
 
```

ericpromislow (changes requested):
I still think we should do something to avoid converting timestamps when we know they haven't been transformed, because transforms run only in vai mode.

Eventually hopefully we can drop the non-vai code, but for now our customers will be running both modes, and we should avoid emitting spurious warnings.

ericpromislow (approved):
Looks good now

crobby (approved):
lgtm

PR #652, opened 2025-05-31
Convert tests to real SQL store

tomleb:
The current ListByOptions tests aren't very useful and have a few flaws:
- The SQL store isn't even being used, instead we're using a mock, so no idea if the queries are valid. I found some that AREN'T.
- The "expected" list is taken directly from the "returnList" test cases. This means that it doesn't matter what the filtering / sorting / etc is done. So again we're testing basically nothing here.
- It's pretty tedious to have to edit the tests whenever the SQL queries changes. We shouldn't care about the query, but instead we should care about the results.
- I found a few issues where some tests shoudl work but fail. I have commented out these tests.
- The expected total vs got total is just wrong. It checks the returned items, but that's not what it should be when using pagination. It's supposed to check how many TOTAL items are in the DBs regardless of how many we ask for.

# Some issues

- It seems there are some issues with sorting on labels when the label doesn't exist
- Unclear if expected at this layer, but for sorting, no sort == default sorting on namespace and name, but if sort is set then we only sort on the given sorting params. Should we always try to sort on namespace, name in addition to what is given?
- Partitions with Namespace set doesn't seem to work

fyi @ericpromislow

### DIFF
--- a/pkg/sqlcache/informer/listoption_indexer_test.go
+++ b/pkg/sqlcache/informer/listoption_indexer_test.go
@@ -11,20 +11,53 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
-	"reflect"
-	"strings"
 	"testing"
 
 	"github.com/rancher/steve/pkg/sqlcache/db"
+	"github.com/rancher/steve/pkg/sqlcache/encryption"
 	"github.com/rancher/steve/pkg/sqlcache/partition"
 	"github.com/rancher/steve/pkg/sqlcache/sqltypes"
+	"github.com/rancher/steve/pkg/sqlcache/store"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/cache"
 )
+
+func makeListOptionIndexer(ctx context.Context, fields [][]string) (*ListOptionIndexer, error) {
+	gvk := schema.GroupVersionKind{
+		Group:   "",
+		Version: "v1",
+		Kind:    "ConfigMap",
+	}
+	example := &unstructured.Unstructured{}
+	example.SetGroupVersionKind(gvk)
+	name := informerNameFromGVK(gvk)
+	m, err := encryption.NewManager()
+	if err != nil {
+		return nil, err
+	}
+
+	db, err := db.NewClient(nil, m, m)
+	if err != nil {
+		return nil, err
+	}
+
+	s, err := store.NewStore(ctx, example, cache.DeletionHandlingMetaNamespaceKeyFunc, db, false, name)
+	if err != nil {
+		return nil, err
+	}
+
+	listOptionIndexer, err := NewListOptionIndexer(ctx, fields, s, true)
+	if err != nil {
+		return nil, err
+	}
+
+	return listOptionIndexer, nil
+}
 
 func TestNewListOptionIndexer(t *testing.T) {
 	type testCase struct {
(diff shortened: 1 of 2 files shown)


pkg/sqlcache/informer/listoption_indexer_test.go
```diff
@@ -650,165 +674,133 @@ func TestListByOptions(t *testing.T) {
...
-		returnList:        []any{&unstructured.Unstructured{Object: unstrTestObjectMap}, &unstructured.Unstructured{Object: unstrTestObjectMap}},
-		expectedList:      &unstructured.UnstructuredList{Object: map[string]interface{}{"items": []map[string]interface{}{unstrTestObjectMap, unstrTestObjectMap}}, Items: []unstructured.Unstructured{{Object: unstrTestObjectMap}, {Object: unstrTestObjectMap}}},
+		partitions:        []partition.Partition{{All: true}},
+		ns:                "",
+		expectedList:      makeList(t, lodgePole, toto, baz, bar, foo),
+		expectedTotal:     5,
 		expectedContToken: "",
 		expectedErr:       nil,
 	})
+	// tests = append(tests, testCase{
```
ericpromislow, line 703: Why is this test commented out?

pkg/sqlcache/informer/listoption_indexer_test.go
```diff
@@ -264,40 +297,112 @@ func TestNewListOptionIndexer(t *testing.T) {
...
+			"name":      "obj5",
+			"namespace": "ns-b",
+			"unknown":   "hi",
+			"labels": map[string]any{
+				"guard.cattle.io": "lodgepole",
+			},
+		},
+	}
+
+	makeList := func(t *testing.T, objs ...map[string]any) *unstructured.UnstructuredList {
```
ericpromislow, line 373: I would prefer this function to be pulled out and made package-level because it doesn't appear to be a closure on any variables in the enclosing scope. I realize it's only used in this test, but it's useful and we might want to call it from other tests.

ericpromislow:
You can probably reference issue [#48263](https://github.com/rancher/rancher/issues/48263) here

ericpromislow:
> It seems there are some issues with sorting on labels when the label doesn't exist

So the reason why the tests with unused labels are failing is that the generator is incomplete.  For the first commented-out code we're generating
```
SELECT DISTINCT o.object, o.objectnonce, o.dekid FROM "_v1_ConfigMap" o
  JOIN "_v1_ConfigMap_fields" f ON o.key = f.key
  LEFT OUTER JOIN "_v1_ConfigMap_labels" lt1 ON o.key = lt1.key
  WHERE
    (lt1.label = ?)
  ORDER BY (CASE lt1.label WHEN ? THEN lt1.value ELSE NULL END) DESC NULLS FIRST
```
where both params are the unused label "flip".


In obsolete PR https://github.com/rancher/steve/pull/614 I'm generating code like this (hang on, it's a bit hairy, but if you 
parse it before and after the `UNION` bit it will make sense:

```
SELECT __ix_object, __ix_objectnonce, __ix_dekid FROM (

  SELECT DISTINCT o.object AS __ix_object, 
                o.objectnonce AS __ix_objectnonce,
                o.dekid AS __ix_dekid,
                ext2."spec.clusterName" AS __ix_ext2_spec_clusterName
                f."metadata.name" AS __ix_f_metadata_name FROM "_v1_Namespace" o
      JOIN "_v1_Namespace_fields" f ON o.key = f.key
      LEFT OUTER JOIN "_v1_Namespace_labels" lt1 ON o.key = lt1.key
      JOIN "management.cattle.io_v3_Project_fields" ext2 ON lt1.value = ext2."metadata.name"
    WHERE lt1.label = ?

UNION ALL

  SELECT DISTINCT o.object AS __ix_object, o.objectnonce AS __ix_objectnonce, o.dekid AS __ix_dekid, NULL AS __ix_ext2_spec_clusterName, f."metadata.name" AS __ix_f_metadata_name FROM
    "_v1_Namespace" o
      JOIN "_v1_Namespace_fields" f ON o.key = f.key
      LEFT OUTER JOIN "_v1_Namespace_labels" lt1 ON o.key = lt1.key
    WHERE o.key NOT IN (SELECT o1.key FROM "_v1_Namespace" o1
               JOIN "_v1_Namespace_fields" f1 ON o1.key = f1.key
               LEFT OUTER JOIN "_v1_Namespace_labels" lt1i1 ON o1.key = lt1i1.key
               WHERE lt1i1.label = ?)
)

ORDER BY __ix_ext2_spec_clusterName ASC NULLS LAST, __ix_f_metadata_name ASC`,
               expectedStmtArgs: []any{"field.cattle.io/projectId", "field.cattle.io/projectId"},
               expectedErr:      "",
       })

```

ericpromislow:
> Partitions with Namespace set doesn't seem to work

Could be -- all the work I've done so far has focused on query generating and executing, and I haven't looked at if we're handling partitions correctly.

PR #661, opened 2025-06-04
Keep track of events for SQL cache watches

tomleb:

This PR adds support for `ResourceVersion` to watch. Essentially, we're keeping track of events as they come in, and then the UI will be able to start a watch request at a specific `ResourceVersion`, which means at a specific point in time.

The UI will receive only events after the given `ResourceVersion`.

**Note:** This currently keeps all events on disk, ALWAYS. I'll add "garbage collection" in another PR.

## How to test

Here's how you can test this.

1. Run Steve with SQL cache enabled in one terminal. (Make sure `KUBECONFIG` env var points to a Kubernetes cluster)
```
go run . --sql-cache --debug
```

2. Make a list request and keep track of the resource version returned. For example, here I get back `985232` as the resource version.

```sh
$ curl -sk https://localhost:9443/v1/configmaps | jq -r '.revision'
985232
```

3. Make some configmaps modification

eg:

```sh
kubectl create configmap foo
```

4. In another terminal, you can run [websocat](https://github.com/vi/websocat). This is similar to telnet where you'll be able to enter messages (separated by newlines (by pressing enter)). **Note:** See I set the resourceVersion to the value gotten from the list request.

```
$ websocat -k wss://localhost:9443/v1/subscribe
{"resourceType":"configmaps","resourceVersion":"985232"}
```

You should receive events from the point of time where you made modifications (in step 3)

### DIFF
--- a/pkg/sqlcache/informer/listoption_indexer.go
+++ b/pkg/sqlcache/informer/listoption_indexer.go
@@ -431,8 +431,8 @@ func (l *ListOptionIndexer) notifyEvent(eventType watch.EventType, oldObj any, o
 	l.watchersLock.RUnlock()
 
 	l.latestRVLock.Lock()
-	l.latestRV = latestRV
 	defer l.latestRVLock.Unlock()
+	l.latestRV = latestRV
 	return nil
 }
 
--- a/pkg/sqlcache/informer/listoption_indexer.go
+++ b/pkg/sqlcache/informer/listoption_indexer.go
@@ -43,7 +43,7 @@ type ListOptionIndexer struct {
 
 	upsertEventsQuery      string
 	findEventsRowByRVQuery string
-	listEventsQuery        string
+	listEventsAfterQuery   string
 	addFieldsQuery         string
 	deleteFieldsByKeyQuery string
 	deleteFieldsQuery      string
@@ -53,7 +53,7 @@ type ListOptionIndexer struct {
 
 	upsertEventsStmt      *sql.Stmt
 	findEventsRowByRVStmt *sql.Stmt
-	listEventsStmt        *sql.Stmt
+	listEventsAfterStmt   *sql.Stmt
 	addFieldsStmt         *sql.Stmt
 	deleteFieldsByKeyStmt *sql.Stmt
 	deleteFieldsStmt      *sql.Stmt
@@ -82,7 +82,7 @@ const (
                        event BLOB NOT NULL,
                        PRIMARY KEY (type, rv)
           )`
-	listEventsAfterRVFmt = `SELECT type, rv, event
+	listEventsAfterFmt = `SELECT type, rv, event
 	       FROM "%s_events"
 	       WHERE rowid > ?
        `
@@ -219,8 +219,8 @@ func NewListOptionIndexer(ctx context.Context, fields [][]string, s Store, names
 	)
 	l.upsertEventsStmt = l.Prepare(l.upsertEventsQuery)
 
-	l.listEventsQuery = fmt.Sprintf(listEventsAfterRVFmt, dbName)
-	l.listEventsStmt = l.Prepare(l.listEventsQuery)
+	l.listEventsAfterQuery = fmt.Sprintf(listEventsAfterFmt, dbName)
+	l.listEventsAfterStmt = l.Prepare(l.listEventsAfterQuery)
 
 	l.findEventsRowByRVQuery = fmt.Sprintf(findEventsRowByRVFmt, dbName)
 	l.findEventsRowByRVStmt = l.Prepare(l.findEventsRowByRVQuery)
@@ -266,7 +266,7 @@ func (l *ListOptionIndexer) Watch(ctx context.Context, opts WatchOptions, events
 	err := l.WithTransaction(ctx, true, func(tx transaction.Client) error {
 		rowIDRows, err := tx.Stmt(l.findEventsRowByRVStmt).QueryContext(ctx, targetRV)
 		if err != nil {
-			return &db.QueryError{QueryString: l.listEventsQuery, Err: err}
+			return &db.QueryError{QueryString: l.listEventsAfterQuery, Err: err}
 		}
 		if !rowIDRows.Next() && targetRV != latestRV {
 			return fmt.Errorf("resourceversion too old")
@@ -276,9 +276,9 @@ func (l *ListOptionIndexer) Watch(ctx context.Context, opts WatchOptions, events
 		rowIDRows.Scan(&rowID)
 
 		// Backfilling previous events from resourceVersion
-		rows, err := tx.Stmt(l.listEventsStmt).QueryContext(ctx, rowID)
+		rows, err := tx.Stmt(l.listEventsAfterStmt).QueryContext(ctx, rowID)
 		if err != nil {
-			return &db.QueryError{QueryString: l.listEventsQuery, Err: err}
+			return &db.QueryError{QueryString: l.listEventsAfterQuery, Err: err}
 		}
 
 		for rows.Next() {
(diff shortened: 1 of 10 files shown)


pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -63,7 +74,23 @@ const (
 	matchFmt                 = `%%%s%%`
 	strictMatchFmt           = `%s`
 	escapeBackslashDirective = ` ESCAPE '\'` // The leading space is crucial for unit tests only '
-	createFieldsTableFmt     = `CREATE TABLE "%s_fields" (
+
+	createEventsTableFmt = `CREATE TABLE "%s_events" (
```
ericpromislow, line 78: For people like me who jump right into the sql statements, a quick comment that "rv" stands for "resource version" would be helpful. Don't need much more than that.  Even something like

```
// rv: resourceVersion
… CREATE TABLE "%s_events" (
                       rv TEXT NOT NULL,
…
```
would be fine. The codebase sometimes uses `rv` to indicate `return value`, sometimes `reflect.ValueOf(...)`, so just the one comment will help readers like me quickly disambiguate

pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -179,6 +212,18 @@ func NewListOptionIndexer(ctx context.Context, fields [][]string, s Store, names
 		return nil, err
 	}
 
+	l.upsertEventsQuery = fmt.Sprintf(
+		`REPLACE INTO "%s_events"(rv, type, event) VALUES (?, ?, ?)`,
+		dbName,
+	)
+	l.upsertEventsStmt = l.Prepare(l.upsertEventsQuery)
+
+	l.listEventsQuery = fmt.Sprintf(listEventsAfterRVFmt, dbName)
```
ericpromislow, line 221: These would be more explanatory if called
`l.listEventsAfterQuery` and
`l.listEventsAfterStmt`

gehrkefc (approved):

pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -204,10 +249,94 @@ func NewListOptionIndexer(ctx context.Context, fields [][]string, s Store, names
...
+			if err != nil {
+				return fmt.Errorf("decoding event object: %w", err)
+			}
+
+			obj, ok := val.Elem().Interface().(runtime.Object)
+			if !ok {
+				continue
+			}
+
+			if !matchFilter(opts.Filter.ID, opts.Filter.Namespace, opts.Filter.Selector, obj) {
```
ericpromislow, line 302: Probably can hoist `filter := opts.Filter` before the `for rows.Next()` stmt and change this to

`if !matchFilter(filter.ID, filter.Namespace, filter.Selector, obj) {`

The compiler is probably doing this anyway -- this is for human readability, so we don't have to repeatedly parse `opts.Filter`

pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -204,10 +249,94 @@ func NewListOptionIndexer(ctx context.Context, fields [][]string, s Store, names
...
+		key = l.addWatcher(eventsCh, opts.Filter)
+		return nil
+	})
 	<-ctx.Done()
 	l.removeWatcher(key)
-	return nil
+	return err
+}
+
+func toBytes(obj any) []byte {
```
ericpromislow, line 324: We should put `toBytes` and `fromBytes` in a shared util package like `pkg/sqlcache/util` to avoid the duplication

pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -204,10 +249,94 @@ func NewListOptionIndexer(ctx context.Context, fields [][]string, s Store, names
...
+	l.latestRVLock.RUnlock()
+
+	targetRV := opts.ResourceVersion
+	if opts.ResourceVersion == "" {
+		targetRV = latestRV
+	}
+
+	var events []watch.Event
+	var key *watchKey
+	// Even though we're not writing in this transaction, we prevent other write to SQL
```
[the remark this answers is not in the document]
tomleb: Fixed.

pkg/sqlcache/informer/listoption_indexer.go
tomleb: Good idea, I added a comment. Resolving, feel free to re-open.

pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -204,10 +249,94 @@ func NewListOptionIndexer(ctx context.Context, fields [][]string, s Store, names
 }
 
 func (l *ListOptionIndexer) Watch(ctx context.Context, opts WatchOptions, eventsCh chan<- watch.Event) error {
```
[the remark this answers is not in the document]
tomleb: It's being used but through an interface: https://github.com/rancher/steve/blob/aa9ced6e3068a19c8874a2e957e4d9df73e58460/pkg/sqlcache/informer/informer.go#L47.

This is where it's being called: https://github.com/rancher/steve/blob/aa9ced6e3068a19c8874a2e957e4d9df73e58460/pkg/stores/sqlproxy/proxy_store.go#L579. Please re-review with this in mind!

pkg/sqlcache/informer/listoption_indexer.go
tomleb: Don't disagree here but we'll need to change this anyway because the objects need to be encrypted. For the purpose of getting UI&QA to test it, I'd rather not change it in this PR, but will change it in another PR.

pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -280,6 +429,10 @@ func (l *ListOptionIndexer) notifyEvent(eventType watch.EventType, oldObj any, o
 		}
 	}
 	l.watchersLock.RUnlock()
+
+	l.latestRVLock.Lock()
+	l.latestRV = latestRV
+	defer l.latestRVLock.Unlock()
```
ericpromislow, line 435: Usual style is to put the `defer... .Unlock()` immediately after the `...Lock()` ?

pkg/sqlcache/informer/listoption_indexer_test.go
```diff
@@ -2088,3 +2099,157 @@ func TestWatchFilter(t *testing.T) {
 	}
 
 }
+
```
ericpromislow, line 2102: This "2102" is why I broke up spun off parts of this file into `query_generator.go` .  Once this settles down I'll do that refactoring, because that PR was on the dead branch to support the `=> [...]` syntax

pkg/sqlcache/informer/listoption_indexer.go
tomleb: Good catch, fixed!

pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -268,6 +399,24 @@ func (l *ListOptionIndexer) notifyEventDeleted(key string, obj any, tx transacti
 }
 
 func (l *ListOptionIndexer) notifyEvent(eventType watch.EventType, oldObj any, obj any, tx transaction.Client) error {
+	acc, err := meta.Accessor(obj)
+	if err != nil {
+		return err
+	}
+
+	latestRV := acc.GetResourceVersion()
+	// Append a -d suffix because the RV might be the same as the previous object
```
ericpromislow, line 408: Could you add unit tests for `-d`-tagging?

ericpromislow (changes requested):
Mostly typos, style nits, a request for a test, and a question about who uses the `Watch` method.

pkg/sqlcache/informer/listoption_indexer.go
ericpromislow: sure

pkg/sqlcache/informer/listoption_indexer.go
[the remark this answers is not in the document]
ericpromislow: My `misread` -- so much of the diff is new, but the actual `Watch` line is unchanged. But will read that...

PR #662, opened 2025-06-06
Handle transaction failure due to canceled context.Context

tomleb:

This PR has a few thing.

**Better handle canceled context.Context**

From the docs:

```go
// BeginTx starts a transaction.
//
// The provided context is used until the transaction is committed or rolled back.
// If the context is canceled, the sql package will roll back
// the transaction. [Tx.Commit] will return an error if the context provided to
// BeginTx is canceled.
//
// The provided [TxOptions] is optional and may be nil if defaults should be used.
// If a non-default isolation level is used that the driver doesn't support,
// an error will be returned.
func (db *DB) BeginTx(ctx context.Context, opts *TxOptions) (*Tx, error) {
```

The context was getting canceled and we were getting the very misleading error `transaction has already been committed or rolled back`. This was due to the transaction being rolled back (via canceled context) and then rolled back again via WithTransaction. Instead, we now try to roll back or commit and detect the issue.

**~Send events directly to the channel~**

~rancher/apiserver seem to not be able to handle fast events due to this [piece of code](https://github.com/rancher/apiserver/blob/a5f8c8f8f1f8ce27a9590fc71348c1a0b886e5e0/pkg/subscribe/watcher.go#L120-L130):~

```go
			select {
			case result <- event:
			default:
				// handle slow consumer
				go func() {
					for range c {
						// continue to drain until close
					}
				}()
				return nil
			}
```

~Using a list of events and sending all of them through the channel was too fast, so we now send each of them to the channel and to a bit of processing in between each send, giving enough time for the "slow" consumer to keep up. I'll create an issue for properly fixing this in rancher/apiserver instead (eg: I think `default:` should be changed to `case <-time.After(200*time.Millisecond)` or something like this.~ I ended up fixing it in apiserver instead in https://github.com/rancher/apiserver/pull/163.

### DIFF
--- a/pkg/sqlcache/informer/listoption_indexer_test.go
+++ b/pkg/sqlcache/informer/listoption_indexer_test.go
@@ -2199,6 +2199,7 @@ func TestWatchResourceVersion(t *testing.T) {
 	tests := []struct {
 		rv             string
 		expectedEvents []watch.Event
+		expectedErr    error
 	}{
 		{
 			rv: "",
@@ -2237,6 +2238,10 @@ func TestWatchResourceVersion(t *testing.T) {
 			rv:             rv5,
 			expectedEvents: nil,
 		},
+		{
+			rv:          "unknown",
+			expectedErr: ErrTooOld,
+		},
 	}
 
 	for _, test := range tests {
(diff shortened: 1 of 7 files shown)


joshmeranda (approved):

pkg/sqlcache/informer/listoption_indexer_test.go
```diff
@@ -2245,11 +2250,14 @@ func TestWatchResourceVersion(t *testing.T) {
...
 
-			assert.Equal(t, test.expectedEvents, gotEvents)
-
 			cancel()
 			err := waitStopWatcher(errCh)
-			assert.NoError(t, err)
+			if test.expectedErr != nil {
+				assert.ErrorIs(t, err, ErrTooOld)
+			} else {
+				assert.Equal(t, test.expectedEvents, gotEvents)
```
ericpromislow, line 2258: This block should be
```
require.NoError(t, err)
assert.Equal(t, test.expectedEvents, gotEvents)
```

Assuming if we aren't expecting an error and we get one, comparing the expected events first will give a confusing result.

ericpromislow (changes requested):
Quick style nit, rest is good

pkg/sqlcache/informer/listoption_indexer_test.go
tomleb: Good point, fixed

ericpromislow (approved):
LG

PR #663, opened 2025-06-07
Use SQL `WITH` statements to sort unbound labels.

ericpromislow:
Related to issue [#50581](https://github.com/rancher/rancher/issues/50581)

It turns out that if you sort a dataset by some label L1, and some rows don't have that label, you still want to see them in the final list. They weren't getting displayed.

Here I create a virtual table of all the rows, substituting a null value for each label that isn't associated on the row, and then sort on that.


ericpromislow:
Fixed the comment

PR #664, opened 2025-06-07
Add garbage collection for events table

tomleb:

This builds on top of https://github.com/rancher/steve/pull/662, I'll rebase when it's reviewed / merged.

This PR adds basic form of garbage collection to the events table. Basically, we don't want to store inifitely all events.

The idea here is simple: every insert to events table we keep only `N` number of events. This is configurable with a default value and per-GVK. **Steve doesn't configure it by default, we'll configure it in rancher/rancher instead**

@pmatseykanets mentioned it would be better to have it be done in batch. It seemed simpler to do it this way (and easier to test). I'm not sure how many will be kept by GVK, probably 1000+. With this amount of data, will we see an impact going this route (every  add/update/delete we run the GC). We can always improve this in the future too.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -3,6 +3,7 @@ package server
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net/http"
 
 	apiserver "github.com/rancher/apiserver/pkg/server"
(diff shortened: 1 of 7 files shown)


ericpromislow (approved):
LG. Nice breakup in the separate commits. 

The DELETE sql stmt needs close reading, but those negative stmts can be a bit tricky. I looked at doing this with a `WITH` block but it didn't make things any clearer.

joshmeranda (approved):

PR #667, opened 2025-06-10
Return `resourceversion too old` error to UI instead of logging

tomleb:

While testing with UI we discovered a few bugs:
- we're not returning the `resourceversion too old` error, we're only logging it.
- the `-d` suffix causes issue and is likely NOT needed, so we're removing it (UI assumes the resourceVersion to be a number and when it's not.. well it tries to parse it anyway because it's JS)
- it's better for the list to be sorted by resourceVersion as number, again because UI makes the assumption that it's a number and also keeps track of the "highest" number.

This PR addresses all these issues. Long-term we'll want to discuss with the UI team to see what we want to do with resourceVersion as they are supposed to be treated as opaque string.

### DIFF
--- a/pkg/sqlcache/informer/informer.go
+++ b/pkg/sqlcache/informer/informer.go
@@ -6,13 +6,16 @@ package informer
 
 import (
 	"context"
+	"errors"
 	"sort"
+	"strconv"
 	"time"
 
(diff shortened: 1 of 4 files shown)


crobby (approved):
Seems to do what it claims to do.  lgtm

PR #673, opened 2025-06-12
Support indexing on array-like fields

ericpromislow:
These are fields that are defined as A.B.C, but not with explicit numeric fields like "metadata.fields[1]"

This is done by registering a custom function to find the N'th substring in a string where multiple values are separated by "|" characters.  We assume that given the kind of string values kubernetes typically uses, we're not likely to look for a string that contains "|' characters.

I found that once I started registering a custom function I could no longer use sqlite in parallel tests on the same database file. So test databases are now run in a temporary directory.  The test suite should clean up these files.


crobby (approved):
I think this looks good.  Nice tests.

pkg/sqlcache/db/client.go
```diff
@@ -21,11 +24,13 @@ import (
 
 	// needed for drivers
 	_ "modernc.org/sqlite"
+	sqlite "modernc.org/sqlite"
 )
 
 const (
 	// InformerObjectCacheDBPath is where SQLite's object database file will be stored relative to process running steve
```
joshmeranda, line 31: We should move this comment to stay with the the `const` it is describing

pkg/sqlcache/db/client.go
```diff
@@ -353,27 +358,43 @@ func closeRowsOnError(rows Rows, err error) error {
...
 	}
 
 	// Set the permissions in advance, because we can't control them if
 	// the file is created by a sql.Open call instead.
-	if err := touchFile(InformerObjectCacheDBPath, informerObjectCachePerms); err != nil {
-		return nil
+	var dbPath string
+	if useTempDir {
+		dir := os.TempDir()
+		f, err := os.CreateTemp(dir, InformerObjectCacheDBPathRoot)
```
joshmeranda, line 382: Should we be using `InformerObjectCacheDBPathRoot` if we are just going to be adding `.db` to it later?

pkg/sqlcache/db/client.go
```diff
@@ -353,27 +358,43 @@ func closeRowsOnError(rows Rows, err error) error {
...
+		if err != nil {
+			return "", err
+		}
+		path := f.Name()
+		dbPath = path + ".db"
+		f.Close()
+		os.Remove(path)
```
joshmeranda, line 389: Isn't this deleting a file that does not necessarily exist? Looks like we are deleting a file at `<dir>/<InformerObjectCacheDBPathRoot> + .db` but we created a file at `<dir>/<InformerObjectCacheDBPathRoot>`?

joshmeranda (changes requested):

pkg/sqlcache/db/client.go
ericpromislow: No, because it provides the prefix for the base temp file:

```
	InformerObjectCacheDBPathRoot := "ioc"
	dir  := os.TempDir()
	f, _ := os.CreateTemp(dir, InformerObjectCacheDBPathRoot)
	fmt.Printf("filename: %s\n", f.Name())
```

=> "/tmp/ioc3178645212" and then we need to append the ".db" extension to keep sqlite happy

ericpromislow: Also sqlite creates 2 auxiliary files, `informer_object_cache.db-shm` and `informer_object_cache.db-wal` so it looks like filename extensions matter.

joshmeranda: Sick, works for me :+1:

pkg/sqlcache/db/client.go
joshmeranda: I'm referring to the `InformerObjectCacheDBPath` constant. We moved it below `InformerObjectCacheDBPathRoot` so in the docs we'll have a description for `InformerObjectCacheDBPath` but on the wrong const `InformerObjectCacheDBPathRoot`

joshmeranda (approved):

pkg/sqlcache/db/client.go
ericpromislow: Related change: the `cleanTempFiles` function in `pkg/sqlcache/informer/listoption_indexer_test.go` to remove the 3 sqlite files after the test finishes.

PR #682, opened 2025-06-17
vai: Rework events garbage collection delete query

pmatseykanets:
Rework the delete query introduced in #664 to avoid `NOT IN` that requires to keep the intermediate result set.


ericpromislow (approved):
LGTM - this SQL is more straightforward than working through the negative test.

PR #684, opened 2025-06-18
#50687 - Added Timestamp Parsing Support

gehrkefc:
# Issue: https://github.com/rancher/rancher/issues/50687

The date fields parser added at https://github.com/rancher/rancher/issues/48673 was only expecting k8s HumanDuration timestamps but as we could verify it can be dangerous, so now it also supports timestamps represented in RFC3339 and if it can't parse, it returns the value 'as is'.

### DIFF
--- a/pkg/resources/common/duration_test.go
+++ b/pkg/resources/common/duration_test.go
@@ -7,7 +7,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestParseHumanDuration(t *testing.T) {
+func TestParseHumanReadableDuration(t *testing.T) {
 	testCases := []struct {
 		name        string
 		input       string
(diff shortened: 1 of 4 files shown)


pkg/resources/common/duration_test.go
```diff
@@ -63,12 +63,16 @@ func TestParseHumanDuration(t *testing.T) {
...
+			output, err := ParseTimestampOrHumanReadableDuration(tc.input)
 			if tc.expectedErr {
 				assert.Error(t, err)
 			} else {
 				assert.Equal(t, tc.expected, output)
 			}
 		})
 	}
+
+	// Testing timestamp parsing separately
```
crobby, line 75: Not really a problem, I guess, but why is this one done separately from the others?

crobby (approved):
Just one question

PR #700, opened 2025-07-01
Fix parsing of metadata.fields when age is already a duration

aruiz14:
Refers to rancher/rancher#50882

It seems under certain conditions, this value can be already transformed into a duration, making this value "polymorphic", so this patch adds support for that.

While looking into this, I noticed we are incorrectly mocking `time.Now`, and only using its value, which can lead to incorrect values.

### DIFF
--- a/pkg/resources/common/formatter.go
+++ b/pkg/resources/common/formatter.go
@@ -291,7 +291,7 @@ func convertMetadataTimestampFields(request *types.APIRequest, gvk schema2.Group
 }
 
 func isDuration(value string) (time.Duration, bool) {
-	d, err := time.ParseDuration(value)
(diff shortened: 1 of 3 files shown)


pkg/resources/common/formatter.go
```diff
@@ -287,6 +290,11 @@ func convertMetadataTimestampFields(request *types.APIRequest, gvk schema2.Group
 	}
 }
 
+func isDuration(value string) (time.Duration, bool) {
+	d, err := time.ParseDuration(value)
```
aruiz14, line 294: Should I use `ParseTimestampOrHumanReadableDuration` here instead?
```suggestion
	d, err := ParseTimestampOrHumanReadableDuration(value)
```

gehrkefc (approved):

PR #717, opened 2025-07-10
Better gc

tomleb:
# Context

(Need to create the issue for this)

We do garbage collection on the *_events table to prevent infinite growing of that table (and thus disk usage). The current way has been found to be taking a lot of CPU when a lot of events are coming in.

This PR introduce a time-based approach to GC: every X interval we run the GC query, which keeps only a configured N events. In a future Rancher/rancher PR, this will be configured via two settings.

The PR includes some additional error wrapping for sQL code (which were found missing). 

That NewInformer function has way too many parameters but I hope to change the way we create informers and cache factories in near future so I kept it this way for now..

I have also moved the log message "CacheFor blablabla" to be an Info message after some feedback on slack.

### DIFF
--- a/pkg/server/cli/clicontext.go
+++ b/pkg/server/cli/clicontext.go
@@ -2,10 +2,12 @@ package cli
 
 import (
 	"context"
+	"time"
 
 	steveauth "github.com/rancher/steve/pkg/auth"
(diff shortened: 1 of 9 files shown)


pmatseykanets (approved):
Looks good.

crobby (approved):

PR #724, opened 2025-07-11
VAI: Replace namespace+name default sort with id 

ericpromislow:
Related to [#50999](https://github.com/rancher/rancher/issues/50999)

The problem this PR solves is that while there are indices on the `namespace` and
`name` fields for each table, there isn't a combined index for the two fields.
It turns out that the `id` field is always set to either `<namespace>/<name>` or `<name>`
for non-namespaced resources, and it's indexed, so we can sort on that.

<strike>
The second commit adds benchmarks. Here are the results:
```
Size    Namespace+Name (ns) time/size (psec)    Id (ns)   time/size (psec)  Indexed NS+N per/size
1000    0.0307               0.0307              0.0165    0.0165            0.01607       0.0161
5000    0.0959               0.0192              0.0821    0.0164            0.08147       0.0163
10000   0.1765               0.0177              0.1845    0.0185            0.1743        0.0174
50000   0.8637               0.0173              0.9515    0.0190            0.8982        0.0180
100000  1.8877               0.0189              2.1145    0.0211            1.8402        0.0184


```

The main thing this tells us is that for smaller datasets sorting by `id` is faster,
but it starts to become slower than the unindexed `namespace`+`name`. I think
this is because the database can very quickly sort items by `namespace` first (as
it's indexed), and since there are smaller sets of `names` within each namespace,
assuming an `O(n log n)` algorithm, sorting each set of names is very fast. With
the combined `id` sort, we get no advantage of first partitioning the data.

The right most pair of columns show the results for adding a combined index. These results
are better for both the unindexed ns+name and id in 9 out of 10 cases. I say we go with that instead.

So another commit on its way…
</strike>

Silvio posted some benchmarks on a networked system that shows that sorting on `id` is much faster than `namespace`+`name` -- this is because the second sort ends up doing a linear search on the `name` field. By `id` everything is indexed.

### DIFF
--- a/pkg/sqlcache/informer/listoption_indexer.go
+++ b/pkg/sqlcache/informer/listoption_indexer.go
@@ -794,7 +794,8 @@ func (l *ListOptionIndexer) constructQuery(lo *sqltypes.ListOptions, partitions 
 	} else {
 		// make sure one default order is always picked
 		if l.namespaced {
-			query += "\n  ORDER BY f.\"metadata.namespace\" ASC, f.\"metadata.name\" ASC "
+			// ID == metadata.namespace + "/" + metaqata.name
+			query += "\n  ORDER BY f.\"id\" ASC "
 		} else {
 			query += "\n  ORDER BY f.\"metadata.name\" ASC "
 		}
--- a/pkg/sqlcache/informer/listoption_indexer_test.go
+++ b/pkg/sqlcache/informer/listoption_indexer_test.go
@@ -11,6 +11,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"math"
 	"os"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


pkg/sqlcache/informer/listoption_indexer_test.go
```diff
@@ -1030,6 +1040,145 @@ func TestNewListOptionIndexerEasy(t *testing.T) {
...
+		IsNamespaced: true,
+	}
+	loi, dbPath, err := makeListOptionIndexer(ctx, opts, false)
+	defer cleanTempFiles(dbPath)
+	assert.NoError(b, err)
+	for _, item := range itemList.Items {
+		err = loi.Add(&item)
+		assert.NoError(b, err)
+	}
+	b.Run(fmt.Sprintf("sort-%d with explicit namespace/name", size), func(b *testing.B) {
```
tomleb, line 1125: Make sure you use b.Loop() here or b.N. Otherwise each test runs only once and that makes benchmarks not reliable. The framework assume you'll be running it multiple times. Here's a resource: https://go.dev/blog/testing-b-loop.

pkg/sqlcache/informer/listoption_indexer_test.go
```diff
@@ -1030,6 +1040,145 @@ func TestNewListOptionIndexerEasy(t *testing.T) {
...
+		}
+		if total != size {
+			b.Errorf("expecting %d items, got %d", size, total)
+		}
+		if len(list.Items) != size {
+			b.Errorf("expecting %d items, got %d", size, len(list.Items))
+		}
+		//verifyListIsSorted(b, list, size)
+	})
+	b.Run(fmt.Sprintf("sort-%d with explicit id", size), func(b *testing.B) {
```
tomleb, line 1154: Make sure you use b.Loop() here or b.N. Otherwise each test runs only once and that makes benchmarks not reliable. The framework assume you'll be running it multiple times. Here's a resource: https://go.dev/blog/testing-b-loop.

pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -790,7 +790,8 @@ func (l *ListOptionIndexer) constructQuery(lo *sqltypes.ListOptions, partitions
 	} else {
 		// make sure one default order is always picked
 		if l.namespaced {
-			query += "\n  ORDER BY f.\"metadata.namespace\" ASC, f.\"metadata.name\" ASC "
+			// ID == metadata.namespace + "/" + metaqata.name
+			query += "\n  ORDER BY f.\"id\" ASC "
```
tomleb, line 794: I know the original issue says `id` column but I honestly thought it would be the `key` column which is the name+namespace already. Isn't that what Silvio meant originally?

Otherwise, I don't see where this id column comes from. Objects received from k8s don't have such a field.

tomleb (changes requested):

PR #725, opened 2025-07-14
Respect shouldEncrypt configuration when storing events

tomleb:
# Issue

In https://github.com/rancher/rancher/issues/40773 we introduced a new `_events` table to store Events to be able to support watches and resource version. In that table we store the resource version, the type of event and the object.

By default, the SQL cache encrypts Tokens and Secrets object. It's also possible to encrypt all objects by setting the env var `CATTLE_ENCRYPT_CACHE_ALL=true`.

When trying to allow multiple encoding options for objects in the DB (as it's currently using `gob`), I noticed that the objects in the events table wasn't encrypted. As such, this adds this support.

Note: The code is essentially the same as https://github.com/rancher/steve/blob/2bba9dfd5c468e594a193a2691d99d7d23a47839/pkg/sqlcache/db/client.go#L354-L363 and https://github.com/rancher/steve/blob/2bba9dfd5c468e594a193a2691d99d7d23a47839/pkg/sqlcache/db/client.go#L335-L349.

---

Since we're so late in the release cycle, I refrained from making big changes to the codebase. But if we go forward with the different encoding approach then I'll probably end up pulling out the encryptor outside of the DB code, so the DB would only store a blob and not "know" about the encryption. (eg: removing the nonce / kid columns).

### DIFF
--- a/pkg/sqlcache/db/client.go
+++ b/pkg/sqlcache/db/client.go
@@ -49,6 +49,8 @@ type Client interface {
 	Upsert(tx transaction.Client, stmt *sql.Stmt, key string, obj any, shouldEncrypt bool) error
 	CloseStmt(closable Closable) error
 	NewConnection(isTemp bool) (string, error)
+	Encryptor() Encryptor
+	Decryptor() Decryptor
 }
 
 // WithTransaction runs f within a transaction.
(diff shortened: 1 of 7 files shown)


ericpromislow (approved):
Looks good

pjbgf (approved):

PR #729, opened 2025-07-16
Stop resetting the VAI DB on any schema or api-service change.

ericpromislow:
### Until v2.12 has shipped -t his is for 2.12.1

Related to [#50978](https://github.com/rancher/rancher/issues/50978)

These are the only two paths into

`pkg/controllers/schema.handler#refreshAll()`, from `OnChangeCRD` and `OnChangeAPIService` .

A change in either of these doesn't need a change in the DB schema, and we still update VAI tables when an object is modified or deleted, so there's no need to reset the DB regarding the data either.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -237,9 +237,6 @@ func setup(ctx context.Context, server *Server) error {
 			if err := ccache.OnSchemas(schemas); err != nil {
 				return err
 			}
-			if err := s.Reset(); err != nil {
-				return err
-			}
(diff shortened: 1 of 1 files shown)


tomleb:
Here's why I don't think we should do this: https://github.com/rancher/rancher/issues/50978#issuecomment-3080388944.

ericpromislow:
No, this isn't enough, because some of the fields we index get picked up from the schema.

PR #735, opened 2025-07-21
Stop rebuilding all watchers: part 2

ericpromislow:
Related to [#50978](https://github.com/rancher/rancher/issues/50978)

Depends on steve PR #733


ericpromislow:
This turned out to be a dead end. See PR <TBD>

PR #745, opened 2025-07-24
Reset db more selectively: approach #2

ericpromislow:
Related to [#509780(https://github.com/rancher/rancher/issues/50978)

This is the first step to solving the underlying problem that we reset the VAI database too often.

This PR ensures the database and informers/watchers are initialized, and then resets them only
when there's been a change that requires doing that. These changes are limited to:

1. A change in a CRD that caused a change in the schema fields that we need to index in the database
2. A CRD is deleted
3. A new CRD is added

### DIFF
--- a/pkg/controllers/schema/schemas.go
+++ b/pkg/controllers/schema/schemas.go
@@ -34,8 +34,8 @@ var (
 
 type SchemasHandlerFunc func(schemas *schema2.Collection, changedSchemas map[string]*types.APISchema, deletedSomething bool) error
 
-func (s SchemasHandlerFunc) OnSchemas(schemas *schema2.Collection, changedSchemas map[string]*types.APISchema, deletedSomething bool) error {
-	return s(schemas, changedSchemas, deletedSomething)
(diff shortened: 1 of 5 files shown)


pkg/controllers/schema/schemas.go
```diff
@@ -88,7 +132,28 @@ func (h *handler) queueRefresh() {
...
+			deletedCRDs = h.deletedCRDs
+			h.deletedCRDs = make(map[k8sapimachineryschema.GroupVersionKind]bool)
+		}
+		if len(h.changedIDs) > 0 {
+			changedIDs = h.changedIDs
+			h.changedIDs = make(map[k8sapimachineryschema.GroupVersionKind]bool)
+		}
+		h.Unlock()
+		if len(changedIDs) > 0 || len(deletedCRDs) > 0 || len(createdCRDs) > 0 {
+			err = h.refreshAll(h.ctx, changedIDs, len(deletedCRDs) > 0 || len(createdCRDs) > 0)
```
crobby, line 154: I think it works, but it does feel/look awkward with the compound conditional used as the parameter.  Maybe split that out for clarity?

ericpromislow: Fixed

gehrkefc (approved):

PR #760, opened 2025-07-31
Sort by ipaddr

ericpromislow:
This lets users sort indexed fields that represent IPv4 and IPv6 addresses correctly, on a byte-by-byte basis. Otherwise an IP like '123.150.2.3' would sort before '123.9.0.0' because the string "1...." after "123." sorts before the "9".

The solution here is to implement mysql's `inet_aton` function as a custom function (using our own code), and then invoking that only when sorting on a field with the `ip(X)` directive.

The function supports both IPv4 and IPv6 addresses. The result of sorting a mix of the two is consistently the same, but I'm not going to attempt to explain how to look at two strings and determine which is the lesser. The function converts the strings into 64-bit integers and does that comparison internally.

In a later issue we can decide if we want to expose this function, and in fact other builtin sqlite functions, more generally, as in:
```
$ curl .../?filter=ip(status.lables.smallerIP)<ip(123.9.0.0)...
```

This PR just supports correct sorting of IP addresses using the `sort=...ip(FIELD)` directive.

The "ip" part is case-sensitive right now, as is everything else in the query parser.

### DIFF
--- a/pkg/sqlcache/db/client.go
+++ b/pkg/sqlcache/db/client.go
@@ -9,9 +9,11 @@ import (
 	"context"
 	"database/sql"
 	"database/sql/driver"
+	"encoding/binary"
 	"encoding/gob"
 	"fmt"
 	"io/fs"
+	"net"
 	"os"
 	"reflect"
 	"strconv"
@@ -21,6 +23,7 @@ import (
 	"errors"
 
 	"github.com/rancher/steve/pkg/sqlcache/db/transaction"
+	"github.com/sirupsen/logrus"
 
 	// needed for drivers
 	_ "modernc.org/sqlite"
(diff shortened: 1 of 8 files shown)


richard-cox:
would this work as follows?

For `/v1/services?sort=ip(spec.clusterIP),metadata.name` they would be sorted by ip and then name. This would be the order?

1. spec.clusterIP: 1.1.1.1
2. spec.clusterIP: 10.1.10.10
4. spec.clusterIP: 10.9.10.10
5. spec.clusterIP: 10.10.10.10
6. spec.clusterIP: 10.10.10.999
7. spec.clusterIP: 10.10.999.10
9. spec.clusterIP: 10.999.10.10

ericpromislow:
That would be the exact order for both sets:
```
# UNSORTED
$ curl -sgkL https://localhost:5111/v1/services'?filter=id~cattle' | jq '.data[] | [.id, .spec.clusterIP]' 
[
  "cattle-fleet-system/gitjob",
  "10.43.252.153"
]
[
  "cattle-fleet-system/monitoring-fleet-controller",
  "10.43.57.166"
]
[
  "cattle-fleet-system/monitoring-gitjob",
  "10.43.206.132"
]
[
  "cattle-provisioning-capi-system/capi-webhook-service",
  "10.43.176.58"
]
[
  "cattle-system/imperative-api-extension",
  "10.43.33.243"
]
[
  "cattle-system/rancher",
  "10.43.16.170"
]
[
  "cattle-system/rancher-webhook",
  "10.43.182.36"
]

# SORTED:
$ curl -sgkL https://localhost:5111/v1/services'?filter=id~cattle&sort=ip(spec.clusterIP)' | jq '.data[] | [.id, .spec.clusterIP]'
[
  "cattle-system/rancher",
  "10.43.16.170"
]
[
  "cattle-system/imperative-api-extension",
  "10.43.33.243"
]
[
  "cattle-fleet-system/monitoring-fleet-controller",
  "10.43.57.166"
]
[
  "cattle-provisioning-capi-system/capi-webhook-service",
  "10.43.176.58"
]
[
  "cattle-system/rancher-webhook",
  "10.43.182.36"
]
[
  "cattle-fleet-system/monitoring-gitjob",
  "10.43.206.132"
]
[
  "cattle-fleet-system/gitjob",
  "10.43.252.153"
]

crobby (approved):

gehrkefc (approved):

PR #761, opened 2025-08-01
Tracing example

tomleb:
Just an example

Run jaeger with:

```
docker run -d --name jaeger   -e COLLECTOR_OTLP_ENABLED=true   -p 16686:16686   -p 4317:4317   -p 4318:4318   jaegertracing/all-in-one:latest
```

Run steve with:

```
OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318 go run . --sql-cache --debug
```

Go to web browser at http://localhost:16686.


pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -611,6 +612,9 @@ func (l *ListOptionIndexer) deleteLabels(tx transaction.Client) error {
 //   - a continue token, if there are more pages after the returned one
 //   - an error instead of all of the above if anything went wrong
 func (l *ListOptionIndexer) ListByOptions(ctx context.Context, lo *sqltypes.ListOptions, partitions []partition.Partition, namespace string) (*unstructured.UnstructuredList, int, string, error) {
+	ctx, span := steveotel.Tracer.Start(ctx, "ListOptionIndexer.ListByOptions")
+	defer span.End()
+
 	queryInfo, err := l.constructQuery(lo, partitions, namespace, db.Sanitize(l.GetName()))
```
gehrkefc, line 618: I suggest adding attributes to the span like the query and the params for example.

```go
// setting attributes at creation...
ctx, span = tracer.Start(ctx, "attributesAtCreation", trace.WithAttributes(attribute.String("hello", "world")))
// ... and after creation
span.SetAttributes(attribute.Bool("isTrue", true), attribute.String("stringAttr", "hi!"))
```
These are the valid types for attrs:
https://pkg.go.dev/go.opentelemetry.io/otel/attribute#KeyValue

PR #762, opened 2025-08-06
Fix slow labels queries

tomleb:
# Issue https://github.com/rancher/rancher/issues/51333

This PR does two things:
1. Configure SQLite to store temporary tables to memory
2. Use `DISTINCT` only when necessary

[This comment](https://github.com/rancher/rancher/issues/51333#issuecomment-3158096876) shows that using in-memory temporary tables speed up queries using `DISTINCT` (due to slow disks).

`DISTINCT` is only necessary when filtering by labels. (TODO: Verify if it's only necessary when making queries filtering by labels with `OR` filters, so that the general case is fast)

### DIFF
--- a/pkg/sqlcache/db/client.go
+++ b/pkg/sqlcache/db/client.go
@@ -458,6 +458,9 @@ func (c *client) NewConnection(useTempDir bool) (string, error) {
 		// if two transactions want to write at the same time, allow 2 minutes for the first to complete
 		// before baling out
 		"_pragma=busy_timeout=120000&"+
+		// store temporary tables to memory, to speed up queries making use
+		// of temporary tables (eg: when using DISTINCT)
+		"_pragma=temp_store=2&"+
 		// default to IMMEDIATE mode for transactions. Setting this parameter is the only current way
 		// to be able to switch between DEFERRED and IMMEDIATE modes in modernc.org/sqlite's implementation
 		// of BeginTx
--- a/pkg/sqlcache/informer/listoption_indexer.go
+++ b/pkg/sqlcache/informer/listoption_indexer.go
@@ -653,7 +653,11 @@ func (l *ListOptionIndexer) constructQuery(lo *sqltypes.ListOptions, partitions 
 		params = withParams
 		joinPartsToUse = joinParts
 	}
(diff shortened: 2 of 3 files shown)


gehrkefc (approved):

ericpromislow (changes requested):
It looks good, although I assume the `logLongQuery` part is temporary. If so there should be a `TODO:` comment specifying it should be removed by version 2.12.x (for x ~ 3 or 4)?

Also I'd like to see a test like this to verify that we handle multiple unbound sort parameters that don't hit every object:
```
diff --git a/pkg/sqlcache/informer/listoption_indexer_test.go b/pkg/sqlcache/informer/listoption_indexer_test.go
index 772eeda..453dd1e 100644
--- a/pkg/sqlcache/informer/listoption_indexer_test.go
+++ b/pkg/sqlcache/informer/listoption_indexer_test.go
@@ -1033,6 +1033,25 @@ func TestNewListOptionIndexerEasy(t *testing.T) {
 		expectedContToken: "",
 		expectedErr:       nil,
 	})
+	tests = append(tests, testCase{
+		description: "ListByOptions sorting on two existing labels, with no label filters, should sort correctly",
+		listOptions: sqltypes.ListOptions{
+			SortList: sqltypes.SortList{
+				SortDirectives: []sqltypes.Sort{
+					{
+						Fields: []string{"metadata", "labels", "horses"},
+					},
+					{
+						Fields: []string{"metadata", "labels", "cows"},
+					},
+				},
+			},
+		},
+		partitions: []partition.Partition{{All: true}},
+		expectedList: makeList(t, obj02a_beef_saddles, obj02_milk_saddles, obj03_saddles,
+			obj02b_milk_shoes, obj03a_shoes, obj04_milk, obj01_no_labels, obj05__guard_lodgepole),
+		expectedTotal: len(allObjects),
+	})
 	//tests = append(tests, testCase{
 	//	description: "ListByOptions with a Namespace Partition should select only items where metadata.namespace is equal to Namespace and all other conditions are met",
 	//	partitions: []partition.Partition{
```

ericpromislow (approved):
Thx for adding the test

PR #765, opened 2025-08-07
Dropping debug-level default to 1 to reduce noise from reflector

crobby:
Issue:  https://github.com/rancher/rancher/issues/44908


ericpromislow (approved):
This is the ticket

PR #770, opened 2025-08-09
[v0.6] Fix new projectsornamespaces negative tests

gehrkefc:
# Original PR: https://github.com/rancher/steve/pull/769


ericpromislow (changes requested):
We know this is wrong.  Don't do anything with this branch until main is working again.

PR #772, opened 2025-08-11
Further fixes to the projectsornamespaces param, with new tests.

ericpromislow:
Related to [#51411](https://github.com/rancher/rancher/51411)


crobby (approved):
I think this looks good.  I believe Eric mentioned that the r/r steveapi tests are also fixed by this, so that is also excellent.

PR #779, opened 2025-08-14
[main] Use normal client for resource creation

joshmeranda:
https://github.com/rancher/rancher/issues/51200

When making create requests for some resources we construct a `TableClient` which set the `Accept` header to `application/json;as=Table;v=v1;g=meta.k8s.io,application/json;as=Table;v=v1beta1;g=meta.k8s.io` which while useful and appropriate for `GET` requests, is not supported for `POST` requests.

To resolve, we create normal clients rather than table clients when trying to create resources via steve.

### DIFF
--- a/pkg/stores/proxy/proxy_store_test.go
+++ b/pkg/stores/proxy/proxy_store_test.go
@@ -49,6 +49,10 @@ func (t *testFactory) TableClient(ctx *types.APIRequest, schema *types.APISchema
 	return t.fakeClient.Resource(schema2.GroupVersionResource{}).Namespace(namespace), nil
 }
 
+func (t *testFactory) Client(ctx *types.APIRequest, schema *types.APISchema, namespace string, warningHandler rest.WarningHandler) (dynamic.ResourceInterface, error) {
+	return t.fakeClient.Resource(schema2.GroupVersionResource{}).Namespace(namespace), nil
+}
+
 func TestWatchNamesErrReceive(t *testing.T) {
 	testClientFactory, err := client.NewFactory(&rest.Config{}, false)
 	assert.Nil(t, err)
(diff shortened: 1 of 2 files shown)


pkg/stores/proxy/proxy_store.go
```diff
@@ -447,7 +447,8 @@ func (s *Store) Create(apiOp *types.APIRequest, schema *types.APISchema, params
 	input["apiVersion"], input["kind"] = gvk.ToAPIVersionAndKind()
 
 	buffer := WarningBuffer{}
-	k8sClient, err := metricsStore.Wrap(s.clientGetter.TableClient(apiOp, schema, namespace, &buffer))
+	// k8sClient, err := metricsStore.Wrap(s.clientGetter.TableClient(apiOp, schema, namespace, &buffer))
```
tomleb, line 450: Can be removed since git will keep history in case we need it back

pkg/stores/sqlproxy/proxy_store.go
```diff
@@ -665,7 +665,8 @@ func (s *Store) Create(apiOp *types.APIRequest, schema *types.APISchema, params
 	input["apiVersion"], input["kind"] = gvk.ToAPIVersionAndKind()
 
 	buffer := WarningBuffer{}
-	k8sClient, err := metricsStore.Wrap(s.clientGetter.TableClient(apiOp, schema, namespace, &buffer))
+	// k8sClient, err := metricsStore.Wrap(s.clientGetter.TableClient(apiOp, schema, namespace, &buffer))
```
tomleb, line 668: Can be removed since git will keep history in case we need it back

tomleb (commented):
Are we planning to add integration tests in Rancher to ensure that this works correctly?

Also, do we need to do the same for update/delete?

pkg/stores/proxy/proxy_store.go
joshmeranda: Oops didn't realize I hadn't remove those :expressionless:

crobby (commented):
Is a similar tweak needed for update requests?

tomleb (approved):
LGTM as there will be tests added in rancher/rancher

PR #804, opened 2025-09-03
Fix using old database while doing a DB reset

tomleb:
# Issue https://github.com/rancher/rancher/issues/51583

There are two problems here:
1. It was possible for the garbage collection to run SQL queries on the old DB with the right timing. We were cancelling the `context.Context` but not waiting on it to finish.
2. We were not fully preventing list requests from hitting an old DB

More details below

## Garbage collection

See first commit. We're now overriding the k8s interface to run the Garbage collection within a wait group. This means [Run() will be called](https://github.com/rancher/steve/blob/0abbdd2940805e4e5ba55936d029299b9a7854f5/pkg/sqlcache/informer/factory/informer_factory.go#L172) which will call the SharedIndexInformer's Run and our GC Run. And then we [wait for both of these to be done](https://github.com/rancher/steve/blob/0abbdd2940805e4e5ba55936d029299b9a7854f5/pkg/sqlcache/informer/factory/informer_factory.go#L206) before doing the DB reset.

## Preventing list requests

We already had a mutex that would prevent calls for CacheFor to get an informer. However, once the informer is acquired, nothing prevents it from accessing the DB (the rest of the List request). So, with the right timing, it's possible that the following happens:
1. CacheFor to get the informer (mutex lock + mutex unlock)
2. Reset is called
3. Rest of list request accessing old DB

To prevent this, I went with the following solution: CacheFor only calls lock on success. It's up to the caller to unlock the mutex with DoneWithCache() once done with the request.

This allows multiple concurrent requests because it's a RWLock and CacheFor does a RLock() call. As soon as Reset is called, no new CacheFor call will go through, it'll wait for all list requests to be finished (shouldn't be too long) and then the DB can be reset. Finally, once the Reset is done, pending CacheFor calls will be able to continue.

## Future work

The next step will be working on https://github.com/rancher/rancher/issues/51569 so DoneWithCache will _likely_ be per-GVK instead of global. We'll see.. It might be more complex to deal with GVKs that depends on other GVKs.

### DIFF
--- a/pkg/sqlcache/informer/factory/informer_factory.go
+++ b/pkg/sqlcache/informer/factory/informer_factory.go
@@ -108,11 +108,20 @@ func NewCacheFactory(opts CacheFactoryOptions) (*CacheFactory, error) {
 
 // CacheFor returns an informer for given GVK, using sql store indexed with fields, using the specified client. For virtual fields, they must be added by the transform function
 // and specified by fields to be used for later fields.
-func (f *CacheFactory) CacheFor(ctx context.Context, fields [][]string, externalUpdateInfo *sqltypes.ExternalGVKUpdates, selfUpdateInfo *sqltypes.ExternalGVKUpdates, transform cache.TransformFunc, client dynamic.ResourceInterface, gvk schema.GroupVersionKind, namespaced bool, watchable bool) (Cache, error) {
+//
+// Don't forget to call DoneWithCache with the given informer once done with it.
+func (f *CacheFactory) CacheFor(ctx context.Context, fields [][]string, externalUpdateInfo *sqltypes.ExternalGVKUpdates, selfUpdateInfo *sqltypes.ExternalGVKUpdates, transform cache.TransformFunc, client dynamic.ResourceInterface, gvk schema.GroupVersionKind, namespaced bool, watchable bool) (*Cache, error) {
 	// First of all block Reset() until we are done
 	f.mutex.RLock()
-	defer f.mutex.RUnlock()
+	cache, err := f.cacheFor(ctx, fields, externalUpdateInfo, selfUpdateInfo, transform, client, gvk, namespaced, watchable)
+	if err != nil {
+		f.mutex.RUnlock()
+		return nil, err
+	}
+	return cache, nil
+}
 
+func (f *CacheFactory) cacheFor(ctx context.Context, fields [][]string, externalUpdateInfo *sqltypes.ExternalGVKUpdates, selfUpdateInfo *sqltypes.ExternalGVKUpdates, transform cache.TransformFunc, client dynamic.ResourceInterface, gvk schema.GroupVersionKind, namespaced bool, watchable bool) (*Cache, error) {
 	// Second, check if the informer and its accompanying informer-specific mutex exist already in the informers cache
 	// If not, start by creating such informer-specific mutex. That is used later to ensure no two goroutines create
 	// informers for the same GVK at the same type
@@ -146,7 +155,7 @@ func (f *CacheFactory) CacheFor(ctx context.Context, fields [][]string, external
 		shouldEncrypt := f.encryptAll || encryptResourceAlways
 		i, err := f.newInformer(f.ctx, client, fields, externalUpdateInfo, selfUpdateInfo, transform, gvk, f.dbClient, shouldEncrypt, namespaced, watchable, f.gcInterval, f.gcKeepCount)
 		if err != nil {
-			return Cache{}, err
+			return nil, err
 		}
 
 		err = i.SetWatchErrorHandler(func(r *cache.Reflector, err error) {
@@ -157,7 +166,7 @@ func (f *CacheFactory) CacheFor(ctx context.Context, fields [][]string, external
 			cache.DefaultWatchErrorHandler(ctx, r, err)
 		})
 		if err != nil {
-			return Cache{}, err
+			return nil, err
 		}
 
 		f.wg.StartWithChannel(f.ctx.Done(), i.Run)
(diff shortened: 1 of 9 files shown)


pkg/sqlcache/informer/factory/informer_factory.go
```diff
@@ -108,11 +108,20 @@ func NewCacheFactory(opts CacheFactoryOptions) (*CacheFactory, error) {
...
 // CacheFor returns an informer for given GVK, using sql store indexed with fields, using the specified client. For virtual fields, they must be added by the transform function
 // and specified by fields to be used for later fields.
-func (f *CacheFactory) CacheFor(ctx context.Context, fields [][]string, externalUpdateInfo *sqltypes.ExternalGVKUpdates, selfUpdateInfo *sqltypes.ExternalGVKUpdates, transform cache.TransformFunc, client dynamic.ResourceInterface, gvk schema.GroupVersionKind, namespaced bool, watchable bool) (Cache, error) {
+//
+// Don't forget to call DoneWithCache with the given informer once done with it.
+func (f *CacheFactory) CacheFor(ctx context.Context, fields [][]string, externalUpdateInfo *sqltypes.ExternalGVKUpdates, selfUpdateInfo *sqltypes.ExternalGVKUpdates, transform cache.TransformFunc, client dynamic.ResourceInterface, gvk schema.GroupVersionKind, namespaced bool, watchable bool) (*Cache, error) {
 	// First of all block Reset() until we are done
 	f.mutex.RLock()
-	defer f.mutex.RUnlock()
+	cache, err := f.cacheFor(ctx, fields, externalUpdateInfo, selfUpdateInfo, transform, client, gvk, namespaced, watchable)
```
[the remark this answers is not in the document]
tomleb: How about this.. In future PRs though..

CacheFor has a LOT of parameters and for some reason we allow doing this:

1. Call CacheFor in List request with fields set A, initializes the cache and returns it
2. Call CacheFor in Watch request with a different set of fields, returns the already initialized cache

Instead, I'd like to do the following:

Have the cachefactory already have all of the fields / transform / etc. And then CacheForGVK would just be: `CacheForGVK(gvk)` and nothing else.

Anyway, I changed cacheFor to cacheForLocked for now. Hope taht's good enough until I rework a bit the cachefor stuff.

pkg/stores/sqlproxy/proxy_store.go
```diff
@@ -803,6 +811,7 @@ func (s *Store) ListByPartitions(apiOp *types.APIRequest, apiSchema *types.APISc
 	if err != nil {
 		return nil, 0, "", fmt.Errorf("cachefor %v: %w", gvk, err)
 	}
+	defer s.cacheFactory.DoneWithCache(inf)
```
ericpromislow, line 814: I assume this is in a `defer` to handle the early-return conditions if an error is detected before getting to `inf.ListByOptions()`, right? Because we can free the lock immediately after that call, right? I guess once we're there we aren't far from returning from the function, so might as well stick with the `defer(...)`.

tomleb: Right, as soon as `CacheFor` succeeds, we need to ensure that `DoneWithCache` is called even if the rest of this function returns an error.

pkg/sqlcache/informer/factory/informer_factory.go
```diff
@@ -166,11 +175,18 @@ func (f *CacheFactory) CacheFor(ctx context.Context, fields [][]string, external
...
 
 	// At this point the informer is ready, return it
-	return Cache{ByOptionsLister: gi.informer}, nil
+	return &Cache{ByOptionsLister: gi.informer}, nil
+}
+
+// DoneWithCache must be called for every CacheFor call.
+//
+// This ensures that there aren't any inflight list requests while we are resetting the database.
+func (f *CacheFactory) DoneWithCache(_ *Cache) {
```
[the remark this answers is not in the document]
ericpromislow: No, please keep it, and add the comment that you're planning to do per-GVK locking.

ericpromislow:
Review note: You'll probably want to add a note to `pkg/sqlcache/Readme.md` around line 81 that there's a semi-global lock that will need to be manually released via `DoneWithCache()`

pkg/sqlcache/informer/factory/informer_factory.go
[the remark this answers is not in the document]
tomleb: added a comment

pkg/sqlcache/informer/factory/informer_factory.go
```diff
@@ -129,24 +129,52 @@ func (f *CacheFactory) CacheFor(ctx context.Context, fields [][]string, external
...
+		}
+	}()
+	if err != nil {
+		f.mutex.RUnlock()
+		return nil, err
+	}
+
+	if !cache.WaitForCacheSync(f.ctx.Done(), gi.informer.HasSynced) {
+		return nil, fmt.Errorf("failed to sync SQLite Informer cache for GVK %v", gvk)
+	}
```
tomleb, line 145: WaitForCacheSync has been moved outside to facilitate the migration to per-GVK in the next PR. Essentially, we're going to be avoiding a deadlock by not having it behind the gi.mutex lock.

tomleb: Pulled this out as it's only needed for the next PR.

ericpromislow (changes requested):
I've got a couple of questions about the code that could possibly be answered with a couple of comments, assuming the code is correct.

ericpromislow (approved):
This looks like a good solution to this first part of the general problem

crobby (approved):
I think this looks good.  Good comments, tests updated.

PR #807, opened 2025-09-08
Drop tables instead of removing the database completely

tomleb:
# Issue https://github.com/rancher/rancher/issues/51569

In this first part, we drop the tables instead of nuking the database.

The next part will doing a more granular reset (dropping only part of the tables).


pkg/sqlcache/informer/factory/informer_factory.go
```diff
@@ -199,25 +199,36 @@ func (f *CacheFactory) Stop() error {
...
 	f.informersMutex.Lock()
 	defer f.informersMutex.Unlock()
-	f.informers = make(map[schema.GroupVersionKind]*guardedInformer)
 
-	// finally, reset the DB connection
-	_, err := f.dbClient.NewConnection(false)
-	if err != nil {
-		return err
+	for gvk, informer := range f.informers {
+		// DropAll needs its own context because the context from the
```
ericpromislow, line 223: Part of the comment is missing: `context from the ???`

pkg/sqlcache/store/store.go
```diff
@@ -524,6 +530,34 @@ func (s *Store) RegisterAfterDeleteAll(f func(txC transaction.Client) error) {
...
 
+func (s *Store) RegisterBeforeDropAll(f func(txC transaction.Client) error) {
+	s.beforeDropAll = append(s.beforeDropAll, f)
+}
+
+// DropAll effectively removes the store from the database. The store must be
+// recreated with NewStore.
+//
+// The store shouldn't be used once DropAll is called.
+func (s *Store) DropAll(ctx context.Context) error {
```
ericpromislow, line 541: This should really be called `DropBase`, and all the other variants of `dropall` should also move to `dropbase`.  "dropall" suggests code that would drop the tables `Foo`, `FooFields`, `FooLabels`, `FooIndices` and any other `Foo*` tables we create in the future.

tomleb: I'd be okay with just `Drop`, `DropBase` feels weird because we'd be calling `DropBase` but then more things would  be dropped in that call due to all registered functions.

crobby (approved):
lgtm.  I'm good with whichever way the Drop naming shakes out.

pkg/sqlcache/informer/factory/informer_factory.go
tomleb: Fixed

PR #812, opened 2025-09-09
Stop single caches instead of all of them

tomleb:
# Issue https://github.com/rancher/rancher/issues/51569

There are 3 areas of focus for this PR:
1. Reverting schema refresh code
2. A new SchemaTracker for the cache factory
3. Only resetting a specific GVK instead of all GVKs

# Reverting schema refresh code

The recent commits around this area introduced some regression which have been fixed, but there are still problems:
- there seems to remain a race with [needToSync](https://github.com/rancher/steve/blob/ca3fa10db55d800a4607ba3319819fad44780a61/pkg/controllers/schema/schemas.go#L319) (I can't remember the details, will try to remember and come back to write it here)
- ccache.OnSchemas is called only when a CRD is added / deleted or fields changes. It used to be called ALWAYS before and I'm nervous this could be a regression.

So instead, I'm reverting it to how it was before, meaning [h.handler.OnSchemas](https://github.com/rancher/steve/blob/ca3fa10db55d800a4607ba3319819fad44780a61/pkg/controllers/schema/schemas.go#L273) only gives the "newest" schema state. Whatever callbacks we have here is responsible to keep track of changes schemas on its own.

This leads to.. 

# SchemaTracker

We need to keep track of the schemas that have been added/ removed and their columns changed. That's the only thing that requires a reset of a cache for a given GVK.

Added tests for this.

# Only resetting a specific GVK

Now that we keep track of which GVK needs to be reset.. We need to DropAll on its cache. The complexity here is dealing with the many locks we have. We end up with 3 locks:
- [informersMutex](https://github.com/rancher/steve/blob/e0c7ae2ce8d0230084c7b41bb20e90bbd9e2cdec/pkg/sqlcache/informer/factory/informer_factory.go#L45): this really is just a lock to guard write/reads in the [informers](https://github.com/rancher/steve/blob/e0c7ae2ce8d0230084c7b41bb20e90bbd9e2cdec/pkg/sqlcache/informer/factory/informer_factory.go#L44) table.
- [informerMutex](https://github.com/rancher/steve/blob/e0c7ae2ce8d0230084c7b41bb20e90bbd9e2cdec/pkg/sqlcache/informer/factory/informer_factory.go#L52): we initialize a new [guardedInformer.informer](https://github.com/rancher/steve/blob/e0c7ae2ce8d0230084c7b41bb20e90bbd9e2cdec/pkg/sqlcache/informer/factory/informer_factory.go#L49) only once. So everytime we want to read/write to it, we use a mutex.
- [stopMutex](https://github.com/rancher/steve/blob/e0c7ae2ce8d0230084c7b41bb20e90bbd9e2cdec/pkg/sqlcache/informer/factory/informer_factory.go#L59): This one replaces the previous global lock for `CacheFor` and `DoneWithCache`. It prevents calling DropAll on a cache that is still being used, which would very likely cause trouble. 

**Note:** [Stop](https://github.com/rancher/steve/blob/e0c7ae2ce8d0230084c7b41bb20e90bbd9e2cdec/pkg/sqlcache/informer/factory/informer_factory.go#L244) takes the informersMutex lock during the whole DropAll bit, meaning it will block new caches from a different GVK being created until the DB tables have been dropped. That seems okay to me since dropAll is a pretty rare operation.

### DIFF
--- a/pkg/controllers/schema/schemas.go
+++ b/pkg/controllers/schema/schemas.go
@@ -40,14 +40,14 @@ func (s SchemasHandlerFunc) OnSchemas(schemas *schema2.Collection) error {
 type handler struct {
 	sync.Mutex
 
-	ctx     context.Context
-	toSync  int32
-	schemas *schema2.Collection
-	client  discovery.DiscoveryInterface
-	cols    *common.DynamicColumns
-	crd     apiextcontrollerv1.CustomResourceDefinitionClient
-	ssar    authorizationv1client.SelfSubjectAccessReviewInterface
-	handler SchemasHandlerFunc
+	ctx       context.Context
+	toSync    int32
+	schemas   *schema2.Collection
+	client    discovery.DiscoveryInterface
+	cols      *common.DynamicColumns
+	crdClient apiextcontrollerv1.CustomResourceDefinitionClient
+	ssar      authorizationv1client.SelfSubjectAccessReviewInterface
+	handler   SchemasHandlerFunc
 }
 
 func Register(ctx context.Context,
@@ -60,13 +60,13 @@ func Register(ctx context.Context,
 	schemas *schema2.Collection) {
 
 	h := &handler{
-		ctx:     ctx,
-		cols:    cols,
-		client:  discovery,
-		schemas: schemas,
-		handler: schemasHandler,
-		crd:     crd,
-		ssar:    ssar,
+		ctx:       ctx,
+		cols:      cols,
+		client:    discovery,
+		schemas:   schemas,
+		handler:   schemasHandler,
+		crdClient: crd,
+		ssar:      ssar,
 	}
 
 	apiService.OnChange(ctx, "schema", h.OnChangeAPIService)
@@ -156,7 +156,7 @@ func (h *handler) refreshAll(ctx context.Context) error {
 		return nil
 	}
 
-	schemas, err := converter.ToSchemas(h.crd, h.client)
+	schemas, err := converter.ToSchemas(h.crdClient, h.client)
 	if err != nil {
 		return err
 	}
--- a/pkg/controllers/schema/schemas.go
+++ b/pkg/controllers/schema/schemas.go
@@ -19,7 +19,6 @@ import (
 	authorizationv1 "k8s.io/api/authorization/v1"
 	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	k8sapimachineryschema "k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/discovery"
 	authorizationv1client "k8s.io/client-go/kubernetes/typed/authorization/v1"
 	apiv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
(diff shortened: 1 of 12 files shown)


pkg/controllers/schema/schemas.go
```diff
@@ -32,31 +31,23 @@ var (
...
-	createdCRDs       map[k8sapimachineryschema.GroupVersionKind]bool
-	deletedCRDs       map[k8sapimachineryschema.GroupVersionKind]bool
-	apiServiceChanged bool
-	gvksFromKeys      map[string][]k8sapimachineryschema.GroupVersionKind
+	ctx     context.Context
+	toSync  int32
+	schemas *schema2.Collection
+	client  discovery.DiscoveryInterface
+	cols    *common.DynamicColumns
+	crd     apiextcontrollerv1.CustomResourceDefinitionClient
```
ericpromislow, line 48: I prefer the old name `crdClient` for an object that's a client and not an instance of a CRD

tomleb: Sure, fixed :+1:

tomleb:
Had to rebase due to conflicts

pkg/sqlcache/informer/factory/informer_factory.go
```diff
@@ -49,13 +47,31 @@ type CacheFactory struct {
...
 	informer *informer.Informer
-	mutex    *sync.Mutex
+	// informerMutex ensures informer is only set by one goroutine even if
+	// multiple concurrent calls to CacheFor are made
+	informerMutex *sync.Mutex
+
+	// stopMutex ensures no CacheFor call can be made for a given GVK when
+	// a Stop call is on-going.
+	//
+	// CacheFactory.informersMutex is not enough because because part of the
```
ericpromislow, line 57: typo: one "because" is fine

pkg/sqlcache/schema.go
```diff
@@ -0,0 +1,78 @@
+package sql
```
ericpromislow, line 1: I would prefer to see this in a sub-package called something like `schematracker` as in `pkg/sqlcache/schematracker`. It feels wrong to put it at the top-level `pkg/sqlcache` when we have no other code at that level, just `integration_test.go`

pkg/sqlcache/schema_test.go
```diff
@@ -0,0 +1,231 @@
+package sql
```
ericpromislow, line 1: Same as above - I'd like to see this in `pkg/sqlcache/schematracke`

ericpromislow (changes requested):
Just typos and packaging issues. Code looks good. Nice moving the schema-tracking stuff into a separate package.

pkg/sqlcache/schema_test.go
tomleb: Good point done.

pkg/sqlcache/informer/factory/informer_factory.go
tomleb: Done.

pkg/sqlcache/informer/factory/informer_factory.go
```diff
@@ -49,13 +47,31 @@ type CacheFactory struct {
 
 type guardedInformer struct {
 	informer *informer.Informer
-	mutex    *sync.Mutex
+	// informerMutex ensures informer is only set by one goroutine even if
+	// multiple concurrent calls to CacheFor are made
+	informerMutex *sync.Mutex
+
+	// stopMutex ensures no CacheFor call can be made for a given GVK when
+	// a Stop call is on-going.
```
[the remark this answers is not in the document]
tomleb: Fixed.

ericpromislow (approved):
Looks good with the various changes.

gehrkefc (approved):

PR #834, opened 2025-09-22
Add debug logging for queries

aruiz14:
## Issue: https://github.com/rancher/rancher/issues/52161
Depends on #833

Setting the `CATTLE_DEBUG_QUERY_LOG` env. variable to a certain value will make Vai print SQL queries and their duration to the requested path.
Query params will be omitted by default, but can also be enabled by setting another env. variable `CATTLE_DEBUG_QUERY_INCLUDE_PARAMS=true`


pkg/sqlcache/db/client.go
```diff
@@ -34,6 +36,8 @@ const (
 	InformerObjectCacheDBPath     = InformerObjectCacheDBPathRoot + ".db"
 
 	informerObjectCachePerms fs.FileMode = 0o600
+
+	debugQueryLogPathEnvVar = "CATTLE_DEBUG_QUERY_LOG"
```
aruiz14, line 40: Probably a terrible new, open to suggestions... 😅

PR #837, opened 2025-09-24
Capture type info so numeric values sort correctly.

ericpromislow:
Related to [#51914](https://github.com/rancher/rancher/issues/51914)

QA template at https://github.com/rancher/rancher/issues/51914#issuecomment-3326025048

### DIFF
--- a/pkg/sqlcache/informer/listoption_indexer_test.go
+++ b/pkg/sqlcache/informer/listoption_indexer_test.go
@@ -1311,11 +1311,40 @@ func TestNewListOptionIndexerTypeGuidance(t *testing.T) {
 			"favoriteFruit":    "9lime",
 		},
 	}
+	// obj17: favoriteFruit is entered as a string
+	// obj18: favoriteFruit is entered as an integer
+	obj17 := map[string]any{
+		"metadata": map[string]any{
+			"name":             "obj17",
+			"namespace":        "ns-a",
+			"someNumericValue": "17",
+			"favoriteFruit":    "17",
+		},
+	}
+	obj18 := map[string]any{
+		"metadata": map[string]any{
+			"name":             "obj18",
+			"namespace":        "ns-a",
+			"someNumericValue": "18",
+			"favoriteFruit":    18,
+		},
+	}
+	obj100 := map[string]any{
+		"metadata": map[string]any{
+			"name":             "obj100",
+			"namespace":        "ns-a",
+			"someNumericValue": "100",
+			"favoriteFruit":    "guava",
+		},
+	}
 	// construct the source list so it isn't sorted either ASC or DESC
 	allObjects := []map[string]any{
+		obj18,
 		obj01,
 		obj11,
 		obj05,
+		obj17,
+		obj100,
 	}
 	ns_a := map[string]any{
 		"metadata": map[string]any{
@@ -1340,7 +1369,7 @@ func TestNewListOptionIndexerTypeGuidance(t *testing.T) {
 	var tests []testCase
 	tests = append(tests,
 		testCase{
-			description: "TestNewListOptionIndexerTypeGuidance() with type-guidance sorts numerically",
+			description: "TestNewListOptionIndexerTypeGuidance() with type-guidance INT on non-ints sorts as string",
 			opts: ListOptionIndexerOptions{
 				Fields:       fields,
 				IsNamespaced: true,
@@ -1349,8 +1378,8 @@ func TestNewListOptionIndexerTypeGuidance(t *testing.T) {
 				},
 			},
 			sortFields:           []string{"metadata", "someNumericValue"},
-			expectedListAscObjs:  []map[string]any{obj01, obj05, obj11},
-			expectedListDescObjs: []map[string]any{obj11, obj05, obj01},
+			expectedListAscObjs:  []map[string]any{obj01, obj05, obj11, obj17, obj18, obj100},
+			expectedListDescObjs: []map[string]any{obj100, obj18, obj17, obj11, obj05, obj01},
 		})
 	tests = append(tests,
 		testCase{description: "TestNewListOptionIndexerTypeGuidance() without type-guidance sorts as strings",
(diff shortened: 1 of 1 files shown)


crobby (approved):
A couple non-blocking questions.
Also...will this wind-up requiring steveapi_test changes when we try to bump steve in r/r?

pkg/stores/sqlproxy/proxy_store.go
```diff
@@ -789,6 +799,33 @@ func (s *Store) Delete(apiOp *types.APIRequest, schema *types.APISchema, id stri
 	return obj, buffer, nil
 }
 
+func getTypeGuidance(cols []common.ColumnDefinition) map[string]string {
+	guidance := make(map[string]string)
+	ptn := regexp.MustCompile(`(?i)\bnumber of\b`)
```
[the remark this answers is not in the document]
tomleb: > Do CRDs have associated schemas?

Yes. CRDs have schema (openAPI just like built-ins) as well as column definitions. (Though remember that [we're handling that not super well right now](https://github.com/rancher/rancher/issues/51934))

In any case, as long as the user defines though additionalPrinterColumn with the correct type, then we don't need special handling for this. Built-in don't define these type correctly, which is unfortunate but probably can't be fixed for backwards compatibility.

pkg/stores/sqlproxy/proxy_store.go
```diff
@@ -789,6 +799,33 @@ func (s *Store) Delete(apiOp *types.APIRequest, schema *types.APISchema, id stri
...
+	for _, col := range cols {
+		td := col.TableColumnDefinition
+		// These come from k8s.io/kubernetes/pkg/printers/internalversion
+		// Some 'number of' fields are declared to be string, but we want to
+		// sort those numbers numerically (like the POD # of a pod)
+		colType := td.Type
+		if colType == "integer" || colType == "boolean" || ptn.MatchString(td.Description) {
+			colType = "INT"
+		}
+		if colType != "string" {
```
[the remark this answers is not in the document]
ericpromislow: yes

ericpromislow:
> Also...will this wind-up requiring steveapi_test changes when we try to bump steve in r/r?

I ran the existing tests before pushing this PR and they passed. Will look to see if we can add this.

ericpromislow:
> I see what you're doing here but I wonder if that's the approach we should go with

I figured there was no point constructing a table given that we can't handle CRDs. 

As for the builtin types, my analysis of printer.go looked to be comprehensive, but I see some of the
code relies on things like `{Name: "Type", Type: "string", Description: apiv1.Secret{}.SwaggerDoc()["type"]},`

Trying to follow the code for `Description` led me to https://kubernetes.io/docs/concepts/configuration/secret/#secret-types , and that wasn't helpful.

ericpromislow:
Proposed integration tests for secrets are at https://github.com/rancher/rancher/pull/52090

.github/workflows/ci.yaml
```diff
@@ -22,9 +22,9 @@ jobs:
     - name: Install mockgen
       run: go install go.uber.org/mock/mockgen@v0.5.2
     - name: Install golangci-lint
-      uses: golangci/golangci-lint-action@55c2c1448f86e01eaae002a5a3a9624417608d84 # v6.5.2
+        uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 # v8.0.0
       with:
-        version: v1.64.8
+        version: v2.4.0
```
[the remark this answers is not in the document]
ericpromislow: I was wrong about that, and backed it out.

crobby (approved):
Seems to be in-line with what was discussed.

pkg/stores/sqlproxy/proxy_store.go
```diff
@@ -817,6 +827,46 @@ func (s *Store) Delete(apiOp *types.APIRequest, schema *types.APISchema, id stri
...
+		"metadata.fields[1]": true, // name: Secrets
+	},
+	schema.GroupVersionKind{Group: "", Version: "v1", Kind: "ConfigMap"}: {
+		"metadata.fields[1]": true, // name: Data
+	},
+}
+
+func getTypeGuidance(cols []common.ColumnDefinition, gvk schema.GroupVersionKind) map[string]string {
+	guidance := make(map[string]string)
+	ptn := regexp.MustCompile(`(?i)\bnumber of\b`)
```
tomleb, line 844: dumb question but what does ptn stands for here?

pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -200,7 +202,12 @@ func NewListOptionIndexer(ctx context.Context, s Store, opts ListOptionIndexerOp
 	l.RegisterBeforeDropAll(l.dropFields)
 	columnDefs := make([]string, len(indexedFields))
 	for index, field := range indexedFields {
-		column := fmt.Sprintf(`"%s" TEXT`, field)
+		typeName := "TEXT"
```
tomleb, line 205: Can we add some tests in listoption_indexer_test.go? Sounds like a `TestTypeGuidance` test would be useful here to see that part work (eg: perhaps with sorting as well, like sorting/filtering on a number?) Does that make sense?

tomleb (commented):
Seems fine but I think we'd benefit from a few more test in listoption_indexer

pkg/stores/sqlproxy/proxy_store.go
ericpromislow: `pattern`

pkg/sqlcache/informer/listoption_indexer.go
ericpromislow: I just added one test with a non-empty TypeGuidance field to verify that the field type is "INT" and not "TEXT". But I don't think we have a unit-test framework that can do a full test that verifies sorting/filtering on a number behaves as expected. If I'm reading the code right, `listoption_indexer_test.go` isn't aware of which kind of database table it's working with; it just passes the fields. So there's no processing `TypeGuidance` there. And we don't have any other unit tests that call `loi.ListByOptions()`.

Any thoughts?

PR #840, opened 2025-09-25
Avoid Watches using write transactions

aruiz14:
## Issue: https://github.com/rancher/rancher/issues/52163
Using `ResourceVersion` for `Watch` requests works as a "resume from previous watch", making the server replay events after the specified point.
In order to not lose any events that could be observed while still replaying those events, we were using a write transaction to prevent any writes, but this affects every connection to the database, not only the event ingestion for this type.
This PR implements an alternative way to prevent losing those events that does not involve blocking the ingestion:
The informer will not be provided with the watcher channel directly, but a intermediate one which will keep all the writes in a buffer until the backfilling is completed. Then, the buffer is written to the original consumer channel. Finally, both channels are piped so it works as normal.

Note: I've merged `latestRV` and `watchers` locks into a single one, as they will now be used in the same place.

### DIFF
--- a/pkg/sqlcache/informer/listoption_indexer.go
+++ b/pkg/sqlcache/informer/listoption_indexer.go
@@ -92,7 +92,7 @@ const (
                        event BLOB NOT NULL,
                        eventnonce BLOB,
 	               dekid BLOB,
-                       PRIMARY KEY (type, rv)
+                       PRIMARY KEY (rv, type)
           )`
 	listEventsAfterFmt = `SELECT type, rv, event, eventnonce, dekid
 	       FROM "%s_events"


pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -323,28 +321,90 @@ func NewListOptionIndexer(ctx context.Context, s Store, opts ListOptionIndexerOp
...
+
+	// We can keep receiving events while replaying older events for the watcher.
+	// By early registering this watcher, this channel will buffer any new events while we are still backfilling old events.
+	// When we finish, calling backfillDone will write all events in the buffer, then listen to new events as normal.
+	watcherChannel, backfillDone := watcherWithBackfill(ctx, eventsCh)
+
+	l.lock.Lock()
 	latestRV := l.latestRV
-	l.latestRVLock.RUnlock()
+	key := l.addWatcherLocked(watcherChannel, opts.Filter)
```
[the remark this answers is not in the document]
aruiz14: Aaah you are completely right 🤦 let me try to test and fix it

pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -323,28 +321,90 @@ func NewListOptionIndexer(ctx context.Context, s Store, opts ListOptionIndexerOp
...
-	l.latestRVLock.RUnlock()
+	l.lock.RUnlock()
 
 	return latestRV
 }
 
+func watcherWithBackfill[T any](ctx context.Context, eventsChan chan<- T) (chan<- T, func()) {
+	backfill, backfillDone := context.WithCancel(ctx)
+	writeChan := make(chan T)
+	buffer := make(chan T, 100)
```
tomleb, line 334: nit: Should buffer be `buffer := make(chan []T, 1)` instead? That way we just fill up a list and after backfilling you pop from that list. (instead of involving channels).

It does mean that we lose on the "free" (but not quite) max length.

pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -323,28 +321,90 @@ func NewListOptionIndexer(ctx context.Context, s Store, opts ListOptionIndexerOp
...
+	go func() {
+		defer close(buffer)
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-backfill.Done():
+				return
+			case event := <-writeChan:
+				buffer <- event
```
tomleb, line 348: Just an observation and if you could confirm that'd be great too.

After 100 events, this write will block until we've read from the DB (backfilling done). This will in turn block the send to watchers in `notifyEvent` which will also block ingestion.

I think it's unlikely that we'll receive 100 events before we finish reading from the DB.. What do you think?

tomleb (commented):
Mostly nits and need for comment but otherwise this looks good.

The watcherWithBackfill I find a bit mind-bending but I suppose that's due to what it's doing.. I'm not sure I'd be able to write it in a more readable way.

pkg/sqlcache/informer/listoption_indexer.go
aruiz14: Correct, and that's the reason why I used a buffered channel and not just a slice, so it could block if the buffer became full, and prevent a possible memory peak (e.g. event storm creating a big buffer slice).

pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -323,28 +321,90 @@ func NewListOptionIndexer(ctx context.Context, s Store, opts ListOptionIndexerOp
...
-	l.latestRVLock.RLock()
+	l.lock.RLock()
 	latestRV = []string{l.latestRV}
-	l.latestRVLock.RUnlock()
+	l.lock.RUnlock()
 
 	return latestRV
 }
 
+func watcherWithBackfill[T any](ctx context.Context, eventsChan chan<- T) (chan<- T, func()) {
```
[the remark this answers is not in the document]
aruiz14: Sure!

pkg/sqlcache/informer/listoption_indexer.go
aruiz14: Please see my comment below, explaining why using a channel instead of a slice, so we can keep the buffer size under control and temporarily block ingestion if that happens.

aruiz14 (commented):
> The watcherWithBackfill I find a bit mind-bending but I suppose that's due to what it's doing.. I'm not sure I'd be able to write it in a more readable way.

Hehe I agree, and trust me I had to rewrite it a few times because I wasn't too happy with it, but this is the best I could do for now.

tomleb (approved):
LGTM.

I would've liked to see more integration tests between watcherWithBackfill and Watch / ingestion. eg: Ensuring we can't receive duplicate events (we know we can't now, but that'd be more to prevent regressions). Those might be hard to write though so I'm not blocking this PR on this. Feel free to add some if you think it's not too difficult.

pkg/sqlcache/informer/listoption_indexer.go
```diff
@@ -393,6 +389,84 @@ func (l *ListOptionIndexer) decryptScanEvent(rows db.Rows, into runtime.Object)
 	return watch.EventType(typ), nil
 }
 
+// watcherWithBackfill creates a proxy channel that buffers events during a "backfill" phase
+// and then seamlessly transitions to live event processing.
+func watcherWithBackfill[T any](ctx context.Context, eventsChan chan<- T, maxBufferSize int) (chan T, func(), func()) {
```
tomleb: I found this function pretty complex to review ngl, there's a lot of variables and different cases, etc. It _probably_ works.. But I felt like this could be simplified.

At its core, it seems like we essentially have a few steps:
1. accumulating events from the ListWatcher into a bounded FIFO queue (size=100)
2. Once we reach this size, we effectively block the ListWatcher ingestion (in your case, reachChan is nil and outputChan is nil so it block on context + backfill done)
3. Once backfill is complete, then we can simply pop from the FIFO and then start poping from the watcherChan channel (basically unblocking the ListWatcher)

I ended up re-writing it to see if I could make this simpler, what do you think about this? https://github.com/tomleb/rancher-steve/blob/pr840-suggestion/pkg/sqlcache/informer/listoption_indexer.go#L394-L459

Less branches and weird Go details to know about (I had to check that a `nil` chan basically "disables" that branch, it wasn't obvious, good thing you commented it though).

PR #841, opened 2025-09-26
Fix nil pointer panic when stopping a cache factory

tomleb:
# Issue https://github.com/rancher/rancher/issues/52124

Basically, if newTestInformer or SetWatchErrorHandler fails, we end up not assigning a `gi.informer`. So then if we stop the cache at that point, we try to `DropAll` from the nil `gi.informer`.

This PR fixes the panic by checking for `gi.informer` first. I added a test to verify this, without the fix, the test panics. With the fix, no longer any panic.

### DIFF
--- a/pkg/sqlcache/informer/factory/informer_factory_test.go
+++ b/pkg/sqlcache/informer/factory/informer_factory_test.go
@@ -535,11 +535,6 @@ func TestCacheFor(t *testing.T) {
 			informers:   map[schema.GroupVersionKind]*guardedInformer{},
 		}
 		f.ctx, f.cancel = context.WithCancel(context.Background())
-
-		go func() {
-			time.Sleep(10 * time.Second)
-			f.Stop(expectedGVK)
-		}()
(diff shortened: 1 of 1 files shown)


pkg/sqlcache/informer/factory/informer_factory_test.go
```diff
@@ -519,6 +520,33 @@ func TestCacheFor(t *testing.T) {
...
+			encryptAll:  true,
+			informers:   map[schema.GroupVersionKind]*guardedInformer{},
+		}
+		f.ctx, f.cancel = context.WithCancel(context.Background())
+
+		go func() {
+			time.Sleep(10 * time.Second)
+			f.Stop(expectedGVK)
+		}()
+		// CacheFor(ctx context.Context, fields [][]string, externalUpdateInfo *sqltypes.ExternalGVKUpdates, transform cache.TransformFunc, client dynamic.ResourceInterface, gvk schema.GroupVersionKind, namespaced bool, watchable bool)
```
crobby, line 544: nonbinding nit, is this commented out line necessary?

PR #847, opened 2025-10-01
Fix race in debouncer refresher

aruiz14:
Related to https://github.com/rancher/rancher/issues/48903

I found the implementation too convoluted, and missing context awareness, so I refactored it for clarity and added unit tests.


ericpromislow (changes requested):
I still see many race-condition failures when running `go test -race ./...` in `scripts/tests.sh`.  Too much output to dump here.

Also I'm getting cert failures, which I saw when trying to bump `stretchr/testify` to `v1.11.1` from `v1.10.0`. When I take out the `-race` option everything passes.

This *might* be failing due to my platform - macOS on M1

aruiz14:
@ericpromislow thanks for the review!
> I still see many race-condition failures when running `go test -race ./...` in `scripts/tests.sh`. Too much output to dump here.

Please note this PR only tackles the data race in a specific tests. There could be more, which would appear if running the entire test suite with `./...`. I checked multiple times that the code modified by this PR no longer has data races:
```
❯ go test -race ./pkg/debounce/... ./pkg/schema/definitions/...
ok  	github.com/rancher/steve/pkg/debounce	1.466s
ok  	github.com/rancher/steve/pkg/schema/definitions	1.586s
```

> Also I'm getting cert failures, which I saw when trying to bump `stretchr/testify` to `v1.11.1` from `v1.10.0`. When I take out the `-race` option everything passes.
> 
> This _might_ be failing due to my platform - macOS on M1

Without more information, I'm afraid I'm not really sure what you are referring to. Do you mean some warnings like these?:
```
authentication.go:75] "Unable to authenticate the request" err="x509: certificate signed by unknown authority (possibly because of \"x509: ECDSA verification failure\" while trying to verify candidate authority certificate \"envtest-environment\")"
```
I also see those, but those do not seem to affect the result of the tests?

ericpromislow:
> authentication.go:75] "Unable to authenticate the request"

Yes, these are exactly what I see, but only under certain conditions. I still haven't worked out why we get them then.

aruiz14:
> > authentication.go:75] "Unable to authenticate the request"
> 
> Yes, these are exactly what I see, but only under certain conditions. I still haven't worked out why we get them then.

I've added `-v` to the `go test` command used by the test script, and that makes it appear in the CI logs as well:
https://github.com/rancher/steve/actions/runs/18338949746/job/52229457781?pr=847#step:8:138
They happen during `TestAuthenticationCustom/authentication_error`, so I guess that's a test that expects those errors to actually happen.

PR #1039, opened 2026-03-04
Make virtual transformers idempotent

rohitsuse:
Issue : https://github.com/rancher/rancher/issues/53703

- In this PR, I am making the transformers function idempotent due to the following the PR description and code comment. https://github.com/kubernetes/kubernetes/pull/131799 mentions that transformers can run twice. 
- There is also code comment in the deltafifo.go codebase of client-go about making transformers idempotent.
```
// It's recommended for the TransformFunc to be idempotent.
// It MUST be idempotent if objects already present in the cache are passed to
// the Replace() to avoid re-mutating them. Default informers do not pass
// existing objects to Replace though.
```
- Make the synthetic watcher to send a bookmark event after the initial list call so that reflector doesn't complain about `no bookmark event` for our synthetic watcher.
- The other commits are about reverting the disabling of WatchList feature since they were planned for 2.15.0.

### DIFF
--- a/pkg/clustercache/controller.go
+++ b/pkg/clustercache/controller.go
@@ -150,17 +150,7 @@ func (h *clusterCache) OnSchemas(schemas *schema.Collection) error {
 		opts := &client.Options{
 			Schema: schema.Schema,
 		}
-		kubeconfigGVK := schema2.GroupVersionKind{Group: "ext.cattle.io", Version: "v1", Kind: "Kubeconfig"}
-		tokenGVK := schema2.GroupVersionKind{Group: "ext.cattle.io", Version: "v1", Kind: "Token"}
-		client := h.summaryClient
-		// Due to a bug in Rancher's extension apiserver for the token and kubeconfig APIs, we
-		// must disable the WatchList features for those APIs.
-		if gvk == kubeconfigGVK || gvk == tokenGVK {
-			client = &noWatchListClient{
-				ExtendedInterface: h.summaryClient,
-			}
-		}
-		summaryInformer := informer.NewFilteredSummaryInformerWithOptions(client, gvr, opts, metav1.NamespaceAll, 2*time.Hour,
+		summaryInformer := informer.NewFilteredSummaryInformerWithOptions(h.summaryClient, gvr, opts, metav1.NamespaceAll, 2*time.Hour,
 			cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}, nil)
 		ctx, cancel := context.WithCancel(h.ctx)
 		w := &watcher{
@@ -309,13 +299,4 @@ func callAll(handlers []interface{}, gvr schema2.GroupVersionKind, key string, o
 	}
 
 	return obj, merr.NewErrors(errs...)
-}
-
-// noWatchListListWatch disables WatchList feature
-type noWatchListClient struct {
-	client.ExtendedInterface
-}
-
-func (n *noWatchListClient) IsWatchListSemanticsUnSupported() bool {
-	return true
 }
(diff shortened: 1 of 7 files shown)


crobby (approved):
I think this should do the trick.  I don't have a better solution for isUnixMilli.

pkg/resources/virtual/dates/dates.go
```diff
@@ -129,3 +133,15 @@ func getValueFromJSONPath(obj map[string]any, jp *jsonpath.JSONPath) (string, bo
 	}
 	return "", false
 }
+
+func isUnixMilli(s string) bool {
+	if len(s) != 13 {
```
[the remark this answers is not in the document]
ericpromislow: Looking backwards, we'll never need to represent dates between 1970-01-01 and 2001-09-09 right? I suppose if we would, we'd be responsible for all those other dates before the Unix epoch, and we aren't.

ericpromislow (approved):
I would change the subject to something like "Work with new k8s 1.35 ListWatcher" because that's the key part of this PR; making the transformers idempotent is a secondary requirement.

ericpromislow (approved):
LG

gehrkefc (approved):

pkg/sqlcache/informer/synthetic_watcher.go
```diff
@@ -51,9 +54,14 @@ func (rw *SyntheticWatcher) receive(client dynamic.ResourceInterface, options me
 		previousState := make(map[string]objectHolder)
 		ticker := time.NewTicker(interval)
 
+		initialSyncSent := false
+
 		for {
 			select {
 			case <-ticker.C:
+				options.SendInitialEvents = nil
```
tomleb, line 64: nit: A comment explaining why would be great here

pkg/clustercache/controller.go
```diff
@@ -150,17 +150,7 @@ func (h *clusterCache) OnSchemas(schemas *schema.Collection) error {
...
-		client := h.summaryClient
-		// Due to a bug in Rancher's extension apiserver for the token and kubeconfig APIs, we
-		// must disable the WatchList features for those APIs.
-		if gvk == kubeconfigGVK || gvk == tokenGVK {
-			client = &noWatchListClient{
-				ExtendedInterface: h.summaryClient,
-			}
-		}
-		summaryInformer := informer.NewFilteredSummaryInformerWithOptions(client, gvr, opts, metav1.NamespaceAll, 2*time.Hour,
+		summaryInformer := informer.NewFilteredSummaryInformerWithOptions(h.summaryClient, gvr, opts, metav1.NamespaceAll, 2*time.Hour,
```
tomleb, line 153: Ok, the issue was fixed in v2.14.0 so assuming custom upgrade from v2.14.x to v2.15, this would work. 👍

pkg/resources/virtual/dates/dates.go
```diff
@@ -87,6 +87,10 @@ func (d *Converter) Transform(obj *unstructured.Unstructured) (*unstructured.Uns
 			}
 		}
 
+		if isUnixMilli(value) {
```
tomleb, line 90: Any chance we can somehow test this?

pkg/sqlcache/informer/synthetic_watcher.go
rohitsuse: Added

pkg/resources/virtual/dates/dates.go
rohitsuse: Yes I added a unit test for the isUnixMilli function.

tomleb: @rohitsakala I realize my question was super vague. I meant more like can we test that running twice the dates transformer will succeed? It's a bit unfortunate that we don't really have a way to test for "success" vs "success but skipped due to error" vs "error" :thinking: 

I think it would be worth it having test for each transformers, running twice, and ensuring they "work". Maybe we can leave that to another tech-debt issue, wdyt?

rohitsuse: Done. Added a unit test for the dates transformer.

tomleb (approved):

PR #1091, opened 2026-04-09
Add zizmor workflow

tomleb:


tomleb:
Not mandatory yet (if ever), will investigate another time

PR #1115, opened 2026-04-15
[v2.12] Close prepared SQL-statements after use

aruiz14:
## Issue: rancher/rancher#54598

Backports: #1081 (issue: rancher/rancher#54478)

### DIFF
--- a/pkg/sqlcache/store/store.go
+++ b/pkg/sqlcache/store/store.go
@@ -181,6 +181,7 @@ func (s *Store) updateExternalInfo(tx transaction.Client, key string, externalUp
 			labelDep.TargetFinalFieldName,
 		)
 		getStmt := s.Prepare(rawGetStmt)
+		defer s.CloseStmt(getStmt)
 		rows, err := s.QueryForRows(s.ctx, getStmt, labelDep.SourceLabelName)
 		if err != nil {
 			if !isDBError(err) {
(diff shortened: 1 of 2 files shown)


crobby (approved):
Looks like a pretty reasonable approximation of the original

pkg/sqlcache/store/store.go
```diff
@@ -181,6 +181,7 @@ func (s *Store) updateExternalInfo(tx transaction.Client, key string, externalUp
 			labelDep.TargetFinalFieldName,
 		)
 		getStmt := s.Prepare(rawGetStmt)
+		defer s.CloseStmt(getStmt)
```
ericpromislow, line 184: In the main branch I avoided using `defer` mainly because I'm not a fan of stacking up `defer`s when inside a loop, and assumed that the next line's call to `s.QueryForRows` is very unlikely to panic as it's mostly running database code -- if that panics we have bigger problems.

If you really want to defer the call I would use an anonymous function:

```
rows, err := func() {
  getStmt := s.Prepare(rawGetStmt)
  defer s.CloseStmt(getStmt)
  return s.QueryForRows(s.ctx, getStmt)
}()

ericpromislow (changes requested):
I'd prefer not to have `defer`s inside loops.

pkg/sqlcache/store/store.go
crobby: good catch....indeed, defer in a loop doesn't do what we'd (or at least I) would like to believe it does.

ericpromislow:
See PR https://github.com/rancher/steve/pull/1153 instead -- puts `defer`s in closures

PR #1116, opened 2026-04-15
[v0.7] Fix off-by-one column

tomleb:
# Issue https://github.com/rancher/rancher/issues/52780 and https://github.com/rancher/rancher/issues/53240

Backport of https://github.com/rancher/steve/pull/980

For CRDs (and only for CRDs) we start sending the timestamp as the column value instead of the duration.

This has the benefit of fixing sorting / filtering in the UI (eg: NodeDriver is broken), moves us closer to a solution we want (columns fields is `metadata.fields[X]` format) and obviously fixes the off-by-one issue we had with CRDs (where the default Name column wasn't taken into account). We also no longer set the Columns attribute in two different places, instead we enhance the attributes with a JSONPath for CRDs.

The PR is probably better reviewed commit by commit:
- The first commit gets rid of the duplicated columns struct (finally)
- The second commit is the meat of the PR, we return timestamp instead of the duration. We do this the same way that the API server does it, by using client-go's jsonpath package. I added a test to ensure it works with CRDs that use selectors (eg: `jsonPath: .conditions[?(@.type=="Selected")].lastUpdateTime`)
- The rest adds integration tests


crobby (approved):
Looks good, seems to work.

ericpromislow (commented):
LG, close enough to the PR on `main`

PR #1130, opened 2026-04-18
Sort POD timestamps according to a specific order

ericpromislow:
Ref issue: [54531](https://github.com/rancher/rancher/issues/54531)

Ascending: When sorting by age, in an ascending sort we want to show the youngest pods first,
and push the NULL-values to the end.

In a descending sort, first show the pods with NULL ages, then the *oldest*, with the youngest at the end.

Since this field, `pod: metadata.fields[3]_1`, is stored as a timestamp, with 0 => null, we need to subtract the
non-0 values from the current timestamp.

This breaks for any pods that were created at midnight Jan 1, 1970 UTC. Fortunately there aren't any.

### DIFF
--- a/tests/integration/testdata/podrestarts/podrestarts.test.yaml
+++ b/tests/integration/testdata/podrestarts/podrestarts.test.yaml
@@ -33,14 +33,17 @@ tests:
     expect:
       - name: restarting-pod
 
-  - description: sort-by-restart-timestamp-ascending-stable-first
+# Timestamps are reported as <number><units>
+# so a timestamp like "5m" (5 minutes ago)
+# "feels" like it's less than an older object with a time like "1h" (1 hour ago)
+  - description: sort-by-restart-timestamp-ascending-restarting-first
     namespace: test-restarts
     query: sort=metadata.fields[3][1]
     expect:
       - name: restarting-pod
       - name: stable-pod
 
-  - description: sort-by-restart-timestamp-descending-restarting-first
+  - description: sort-by-restart-timestamp-descending-stable-first
     namespace: test-restarts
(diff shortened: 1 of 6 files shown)


tests/integration/testdata/podrestarts/podrestarts.test.yaml
```diff
@@ -37,15 +37,15 @@ tests:
     namespace: test-restarts
     query: sort=metadata.fields[3][1]
     expect:
-      - name: stable-pod
       - name: restarting-pod
+      - name: stable-pod
```
crobby, line 44: Seems like the description of this test should change to be stable-last?

tests/integration/testdata/podrestarts/podrestarts.test.yaml
```diff
@@ -37,15 +37,15 @@ tests:
...
 
   - description: sort-by-restart-timestamp-descending-restarting-first
     namespace: test-restarts
     query: sort=-metadata.fields[3][1]
     expect:
-      - name: restarting-pod
       - name: stable-pod
+      - name: restarting-pod
```
crobby, line 51: Similar description change...looks like restarting-last is the case now

crobby (approved):
Seems reasonable.  Other than the description on the tests, I don't have any issues.  I took it for a test drive and it seem to work well.

tests/integration/testdata/podrestarts/podrestarts.test.yaml
ericpromislow: Good catch - I missed those

crobby (approved):

rohitsuse (approved):

PR #1159, opened 2026-04-27
Replace push:tags release trigger with validated workflow_dispatch

tomleb:
# Issue rancher/rancher#54790

Replace the `on: push: tags: v*` trigger with two `workflow_dispatch` workflows that validate the version before creating the tag and GitHub release.

## Cut release (`.github/workflows/cut-release.yaml`)

Entry point. The `validate` job:
- Checks version format (`v<major>.<minor>.<patch>[-prerelease]`)
- Confirms the version prefix matches the branch's allowed minor (read from `VERSION.md` on the default branch)
- Enforces sequential patch ordering within the minor line
- Confirms the tag does not already exist

The `tag` job (runs after `validate`):
- Creates and pushes the annotated tag
- Dispatches the On release workflow at the new tag ref via `gh workflow run release.yaml --ref <tag>`

## On release (`.github/workflows/release.yaml`)

`workflow_dispatch`. A `guard` job rejects any dispatch on a non-tag ref. The `release` job creates the GitHub release using `github.ref_name` as the version.

## Why two workflows

For consistency with rancher/webhook.

A `push: tags` trigger does not work here: GitHub blocks workflow runs triggered by `GITHUB_TOKEN` from starting other runs via push events, so the tag pushed by Cut release would not fire On release. `workflow_dispatch` is exempt from that restriction.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -27,7 +27,7 @@ jobs:
         env:
           VERSION: ${{ inputs.version }}
         run: |
-          if ! [[ "$VERSION" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(?:-[a-zA-Z0-9.]+)?$ ]]; then
+          if ! [[ "$VERSION" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9.]+)?$ ]]; then
             echo "::error::Version '$VERSION' must match v<major>.<minor>.<patch>[-prerelease]"
             exit 1
           fi
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -27,7 +27,7 @@ jobs:
         env:
           VERSION: ${{ inputs.version }}
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -1,21 +1,118 @@
...
+            exit 1
+          fi
+
+      - name: Validate version is next-sequential
+        env:
+          VERSION: ${{ inputs.version }}
+          ALLOWED: ${{ steps.minor.outputs.allowed }}
+        run: |
+          v_core="${VERSION%%-*}"
+          latest=$(git tag -l "${ALLOWED}.*" | grep -v -- '-' | sort -V | tail -n1 || true)
```
ericpromislow: I would prefer `sort --version-sort` -- single-letter options are better for the command-line, especially when using a relatively recently added option.

Same for preferring `git tag --list` over `git tag -l` but no need to do the same for `grep -v` or `tail -n` -- these are commonly used idioms.

ericpromislow (changes requested):
Request to use a couple of longer-form options for less well-known options.

ericpromislow (approved):
Would have preferred `git --list` to `git -l` but this si fine

ericpromislow (approved):
Is fine -- I'd prefer using long options instead of shortcut options for all but the most common options, butlet's get this in....

PR #1164, opened 2026-04-28
[v2.13] Add multiple label and field pair support in ExternalLabelDepe…

joshmeranda:
https://github.com/rancher/rancher/issues/54674

> https://github.com/rancher/rancher/issues/52810
>
> When two Projects are manually created with the same metadata.name on different clusters, the sql queries in pkg/sqlcache/store/store.go checking for projects matching a secret's management.cattle.io/project-scoped-secret label will return both secrets. In order to uniquely identify the Project to link to a project scoped secret, we need to check both the metadata.name AND spec.clusterName.

### DIFF
--- a/pkg/sqlcache/store/store.go
+++ b/pkg/sqlcache/store/store.go
@@ -186,8 +186,8 @@ func (s *Store) updateExternalInfo(tx db.TxClient, key string, externalUpdateInf
 				labelDep.SourceGVK, labelDep.TargetFinalFieldName)
 			err = func() error {
 				preparedStmt := s.Prepare(rawStmt)
+				defer preparedStmt.Close()
 				_, err := tx.Stmt(preparedStmt).Exec(finalTargetValue, sourceKey)
-				preparedStmt.Close()
 				return err
 			}()
 			if err != nil {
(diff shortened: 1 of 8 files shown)


pkg/sqlcache/store/store.go
```diff
@@ -196,8 +186,8 @@ func (s *Store) updateExternalInfo(tx db.TxClient, key string, externalUpdateInf
 				labelDep.SourceGVK, labelDep.TargetFinalFieldName)
 			err = func() error {
 				preparedStmt := s.Prepare(rawStmt)
-				defer preparedStmt.Close()
 				_, err := tx.Stmt(preparedStmt).Exec(finalTargetValue, sourceKey)
+				preparedStmt.Close()
```
ericpromislow: Better to keep the `defer preparedStmt.Close()`

pkg/sqlcache/store/store.go
```diff
@@ -249,8 +239,8 @@ func (s *Store) updateExternalInfo(tx db.TxClient, key string, externalUpdateInf
 				nonLabelDep.SourceGVK, nonLabelDep.TargetFinalFieldName)
 			err = func() error {
 				preparedStmt := s.Prepare(rawStmt)
-				defer preparedStmt.Close()
 				_, err := tx.Stmt(preparedStmt).Exec(finalTargetValue, sourceKey)
+				preparedStmt.Close()
```
ericpromislow: Same as above: keep the `defer preparedStmt.Close()` in case the call to `tx.Stmt(...).Exec` panics

ericpromislow (changes requested):
Looks fine, but I think we can use the `defer` calls.

ericpromislow (approved):
Looks fine - straightforward backport

PR #1165, opened 2026-04-28
Remove Table GVK on converted UnstructuredList

crobby:
client-go for k8s v1.36 reflectors reject lists whose top-level GVK is meta.k8s.io/v1 Kind=Table. After unwrapping the rows into Items, overwrite the list's kind/apiVersion so the reflector accepts the result.


crobby:
Closing as this has been folded into the k8s 1.36 bump for steve.

PR #1174, opened 2026-05-06
Bump k8v1.36

Teja78906:
https://github.com/rancher/rancher/issues/54303

add support for kubernetes v1.36

Bump Kubernetes v1.35 → v1.36: 
1. Dependency bumps (go.mod / go.sum): All k8s.io/* packages (api, apimachinery, apiserver, client-go, etc.) moved from v0.35.4 to v0.36.0, along with downstream deps like etcd, OpenTelemetry, gRPC, and various golang.org/x/* libraries that track k8s releases. Go toolchain was also bumped from 1.24/1.25 → 1.26.0.
2. Interface compliance fixes: k8s v1.36 added new methods to cache.Store and cache.SharedIndexInformer. The PR adds stub implementations (LastStoreSyncResourceVersion, Bookmark, HasSyncedChecker) to keep the project's types compliant with the updated interfaces.
3. Table → List GVK rewrite (tablelistconvert/client.go): client-go v0.36's reflector now rejects lists whose top-level GVK is Table. The fix rewrites the object's kind/apiVersion to match the actual resource list type (or drops them entirely when there are no rows).
4. Deprecation cleanups: Replaced deprecated io/ioutil calls (ioutil.TempFile, ioutil.WriteFile, ioutil.ReadAll) with their modern os/io equivalents —> these were flagged by the updated linter (golangci-lint v2.12.1).
5. CI / envtest: Updated envtest binaries to v1.36.0 with new checksums, and made the checksum verification script portable across Linux and macOS.


crobby (approved):

pkg/sqlcache/store/store.go
```diff
@@ -70,13 +70,23 @@ type Store struct {
 	beforeDropAll  []func(tx db.TxClient) error
 }
 
+// LastStoreSyncResourceVersion implements [cache.Store].
+func (s *Store) LastStoreSyncResourceVersion() string {
+	panic("unimplemented")
```
tomleb: Since we don't really control when that is called and can change under us when we bump k8s patches, we shouldn't panic here.

I'm trying to see what a minimal implementation could look like, let me come back real quick on this.

---

Note:

The proxy store in steve already keeps track of a RV that is updated when we receive Added/Modified/Deleted events. One would think we could use those new interfaces to keep track of the RV instead, however some code relies on the RV being an event that changes the event log (eg: the UI sends an RV for a watch, we look up the event log). Since those bookmark events don't end up in that event log, the UI calls could fail.

So either we unify those two RV tracking mechanism (so we'd add bookmark events in the event log perhaps) or we keep them separate.

tomleb (changes requested):
Let's not panic, but implement the interface correctly.

pkg/sqlcache/store/store.go
Abhishek-Valaboju: Thanks for the context. I removed the panic and added a minimal RV tracker for Bookmark() and LastStoreSyncResourceVersion().

pkg/sqlcache/store/store.go
```diff
@@ -68,6 +69,26 @@ type Store struct {
...
+
+// LastStoreSyncResourceVersion implements [cache.Store].
+func (s *Store) LastStoreSyncResourceVersion() string {
+	s.lastSyncRVMu.RLock()
+	defer s.lastSyncRVMu.RUnlock()
+	return s.lastSyncRV
+}
+
+// Bookmark implements the cache.Store interface.
+func (s *Store) Bookmark(resourceVersion string) {
```
tomleb, line 86: We also need to do this for the [Add](https://github.com/rancher/steve/blob/da57bfe56daa061e431b91311c0f14ea713f7240/pkg/sqlcache/store/store.go#L357) method, the [Update](https://github.com/rancher/steve/blob/da57bfe56daa061e431b91311c0f14ea713f7240/pkg/sqlcache/store/store.go#L382) method and the [Delete](https://github.com/rancher/steve/blob/da57bfe56daa061e431b91311c0f14ea713f7240/pkg/sqlcache/store/store.go#L407) method.

This is what upstream's Store implementation does:
- [Add](https://github.com/kubernetes/client-go/blob/1d95f02c1723330f751c9eb255aefb0dcd7ca3d6/tools/cache/thread_safe_store.go#L301): Just calls Update which updates the RV
- [Update](https://github.com/kubernetes/client-go/blob/1d95f02c1723330f751c9eb255aefb0dcd7ca3d6/tools/cache/thread_safe_store.go#L309): Updates the RV
- [Delete](https://github.com/kubernetes/client-go/blob/1d95f02c1723330f751c9eb255aefb0dcd7ca3d6/tools/cache/thread_safe_store.go#L329): Only updates the RV if obj is not nil and is an actual object (deletes calls don't always contain the object)

tomleb (approved):

PR #1202, opened 2026-05-19
[main] add rancher settings formatters

joshmeranda:
Add formattesr currently done by [norman](https://github.com/rancher/rancher/blob/main/pkg/api/norman/customization/setting/setting.go#L20-L33) to steve

### DIFF
--- a/pkg/resources/formatters/formatter.go
+++ b/pkg/resources/formatters/formatter.go
@@ -5,7 +5,9 @@ import (
 	"compress/gzip"
 	"encoding/base64"
 	"encoding/json"
+	"fmt"
 	"io"
+	"slices"
 
 	// helm v2 is long since deprecated
 	// Unlike helm v3, it uses Protobuf encoding, so we can't use generic decoding without the message descriptors.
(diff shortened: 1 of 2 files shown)


pkg/resources/formatters/formatter.go
```diff
@@ -68,6 +70,27 @@ func Pod(_ *types.APIRequest, resource *types.RawResource) {
 	}
 }
 
+var ReadOnlySettings = []string{
+	"cacerts",
+}
+
+// Setting is the default formatter for management.cattle.io.setting resources.
+// It intentionally performs no mutation and serves as an explicit extension point for setting-specific formatting.
```
crobby: It looks like this function does affect both value and links, so I'm not sure the "intentionally performs no mutation" comment is correct

crobby (changes requested):
One question around the "no mutation" comment and should we have a small test for this?

pkg/resources/formatters/formatter.go
joshmeranda: Oh good find, I must've missed this comment.

crobby (approved):

gehrkefc (approved):

PR #1206, opened 2026-05-25
[release/v0.7] Close rows before Err in ListOptionIndexer.Watch backfill

tomleb:
# Issue https://github.com/rancher/rancher/issues/55165

## Summary
- `decryptScanEvent` scans into `sql.RawBytes`, so `database/sql` keeps `Rows.closemu` RLocked across the `Scan` return (Go's `closemuScanHold`, [Go issue 60304](https://github.com/golang/go/issues/60304)).
- If the `Watch` backfill loop exits via `latestRevisionReached` (no further `Next()` call), that RLock is still outstanding when `rows.Err()` runs.
- A racing ctx cancel can wedge `rows.Err()` on a recursive RLock behind a writer queued by `Rows.awaitDone`, which in turn pins `Tx.closemu` so `Tx.rollback` wedges. The SQLite read transaction never closes, the WAL cannot be reset, and writes append unboundedly (observed at 34 GB in production, SURE-11644).
- This PR calls `rows.Close()` explicitly before `rows.Err()` on the happy-path return. `Close` releases the scan-held RLock first and then takes the writer lock cleanly, so the following `rows.Err()` acquires its RLock against an unheld mutex.
- The deferred `rows.Close()` stays as a safety net for the early-return paths inside the loop; it's a no-op after the explicit call (`Close` is idempotent per the stdlib docs).

## Test plan
- [x] `go build ./...`
- [x] `go test -count=1 ./pkg/sqlcache/informer/...`
- [x] Standalone reproducer (Scan into `*sql.RawBytes`, exit loop early, call `Err` before `Close` with racing ctx cancel) deadlocks pre-fix and unwinds cleanly post-fix.
- [x] Standalone reproducer also confirms that with the deadlock active, `PRAGMA wal_checkpoint(RESTART|TRUNCATE)` reports `busy=1` and the WAL file grows unbounded under continued writes from a second connection — matching the production symptom.

## Notes
A deterministic unit test for the wedge condition isn't feasible: the race window is microseconds wide and lives inside a closure that `Watch` doesn't expose hooks into. The `select { case eventsCh <- ev: case <-ctx.Done(): }` in the loop body intercepts any cancellation that arrives during iteration, so the only path that reaches the vulnerable `return rows.Err()` is the `latestRevisionReached` exit. Production hits it via sheer volume of watch establishments × cancellations over long-running deployments, with a permanent consequence per hit (no recovery without restart).

### DIFF
--- a/pkg/sqlcache/db/client.go
+++ b/pkg/sqlcache/db/client.go
@@ -239,13 +239,7 @@ func (c *client) ReadObjects(rows Rows, typ reflect.Type) ([]any, error) {
 		}
 		result = append(result, dest)
 	}
-	err := rows.Err()
-	if err != nil {
-		return nil, closeRowsOnError(rows, err)
-	}
-
-	err = rows.Close()
-	if err != nil {
+	if err := rows.Close(); err != nil {
 		return nil, err
 	}
 
@@ -267,13 +261,7 @@ func (c *client) ReadStrings(rows Rows) ([]string, error) {
 
 		result = append(result, key)
 	}
-	err := rows.Err()
-	if err != nil {
-		return nil, closeRowsOnError(rows, err)
-	}
-
-	err = rows.Close()
-	if err != nil {
+	if err := rows.Close(); err != nil {
 		return nil, err
 	}
 
@@ -295,13 +283,7 @@ func (c *client) ReadStrings2(rows Rows) ([][]string, error) {
 
 		result = append(result, []string{key1, key2})
 	}
-	err := rows.Err()
-	if err != nil {
-		return nil, closeRowsOnError(rows, err)
-	}
-
-	err = rows.Close()
-	if err != nil {
+	if err := rows.Close(); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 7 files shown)


aruiz14 (approved):

tomleb:
Alright so we have a choice to make between:
1. fixing it in the interface (Merging Close and Err together, always calling Close before Err), and forward porting that.
2. just fixing it locally here in 2.13.5, calling `rows.Close()` before the `rows.Err()`

Note that this issue is fixed upstream and will be released in 1.27 eventually (unsure about backports though). This means we should be able to remove the fix in the future. Given 2.14 and 2.15 doesn't have this code at all, it would be safe to fix it locally here only, and :crossed_fingers: we wouldn't see that issue again. Personally I'm fine with either solution, but forwarding porting the fix ensures we can sleep at night without worrying.

cc: @gehrkefc

gehrkefc:
@tomleb I'd go with the interface fix and forward port it. That'd make sure we always call both functions and in the right order.

gehrkefc (approved):

PR #1209, opened 2026-05-27
[release/v0.8] sqlcache: tolerate DeletedFinalStateUnknown in Store.Delete

tomleb:
# Issue https://github.com/rancher/rancher/issues/56398

## Symptom

With the default `ui-sql-cache=true`, on a moderately busy downstream cluster the `cattle-cluster-agent` (`cluster-register` container) logs hundreds of identical errors per watch resync window:

```
level=error msg="Error in Store.Delete for type _v1_Event: transaction: object does not implement the Object interfaces"
```

Reported volume from one agent restart in the issue: 238 errors in one minute right after `_v1_Secret` informer init, 514 in a later resync minute. The error fires per resource instance, so anything with high churn (Events especially) amplifies the noise. And it's not just noise — see "Functional impact" below.

## Trigger

When the reflector loses its watch (network blip, apiserver-side watch timeout, `410 Gone` from a stale resourceVersion, etc.) and falls back to a fresh `List`, client-go's FIFO performs deletion detection: for every key it knew about that is no longer in the new list, it synthesizes a `Deleted` event wrapping the last-known object in `cache.DeletedFinalStateUnknown{Key, Obj}`. `processDeltas` then calls `clientState.Delete(tombstone)` — and `clientState` is steve's SQLite store.

```mermaid
sequenceDiagram
    autonumber
    participant API as kube-apiserver
    participant R as Reflector
    participant F as DeltaFIFO
    participant P as processDeltas
    participant S as steve Store.Delete<br/>(SQLite)
    participant H as notifyEventDeleted<br/>(after-delete hook)

    R->>API: List
    API-->>R: items [A, B, C]
    R->>F: Replace([A, B, C])
    F->>P: Sync deltas
    P->>S: Add(A), Add(B), Add(C)

    R->>API: Watch (RV)
    Note over R,API: watch dies (network/timeout/410)

    R->>API: List (re-list)
    API-->>R: items [A, C]  (B was deleted while watch was down)
    R->>F: Replace([A, C])
    Note over F: B missing → synthesize<br/>Deleted{DeletedFinalStateUnknown{key=B, Obj=B_lastKnown}}
    F->>P: Deleted delta (tombstone)
    P->>S: Delete(tombstone)

    rect rgb(255, 230, 230)
        S->>H: runAfterDelete(key, tombstone, tx)
        H->>H: meta.Accessor(tombstone) ❌<br/>"object does not implement the Object interfaces"
        H-->>S: error
        Note over S: WithTransaction rolls back<br/>→ SQL DELETE never commits<br/>→ B stays in SQLite cache
    end
```

The `clientState.Delete` call lands in `pkg/sqlcache/store/store.go`. The after-delete hook `notifyEventDeleted` in `pkg/sqlcache/informer/listoption_indexer.go` eventually calls `meta.Accessor` on the raw arg. `cache.DeletedFinalStateUnknown` does not implement `metav1.Object`, so the hook errors. Because the after-delete hook runs inside the same SQL transaction as the `DELETE FROM …`, the error rolls the whole transaction back.

## Functional impact

Not just log spam. Per `db.WithTransaction` in `pkg/sqlcache/db/client.go`, the after-delete hook's error rolls back the surrounding transaction — so the SQLite row that was about to be deleted **survives**. The cache then holds an entry for an object that no longer exists upstream, until the next full resync rebuilds the table. Any reads served from the SQLite cache during that window (the whole point of `ui-sql-cache=true`) will see the deleted object.

## Fix

Two layered changes:

1. **`Store.Delete` unwraps the tombstone.** `cache.Store` consumers are expected to tolerate `DeletedFinalStateUnknown` (that's why `DeletionHandlingMetaNamespaceKeyFunc` exists). After computing the key, unwrap the wrapper so all after-delete hooks see a typed object (or `nil`). If the wrapped `Obj` is `nil`, short-circuit: the FIFO's `knownObjects` is this same Store, so if its lookup failed we have nothing meaningful to do either. Logged at `warn` (not `error`) to keep the rare case visible without re-introducing the log flood.

2. **`notifyEventDeleted` collapses to a one-liner.** With the unwrap upstream, the hook no longer needs the `GetByKey` round-trip that fetched `oldObj`. That also removes its pre-existing `!exists` error path — which would have rolled the DELETE back on a missing cache entry, exactly the same failure shape this PR fixes.

Net effect:
- Bug fixed.
- One fewer SQLite read per delete in the steady-state path.
- The rare nil-payload tombstone degrades to a single warn log and a no-op, instead of an error stream and a stale row.

## Tests

New end-to-end test `TestStoreDeleteTombstone` in `tests/integration/tombstone_test.go` reproduces the bug through the real reflector pipeline:

- Real apiserver-backed `List` (via `dynamic.ResourceInterface`).
- Test-controlled `Watch` (a `watch.FakeWatcher` we can `.Error(...)` on demand) — required because a live watch would deliver the Delete event normally and never trigger the tombstone path.
- Delete the Banana via the real apiserver, then close the fake watch with a 410 Gone-equivalent error. `handleWatch` returns the error, `r.watch` returns, `ListAndWatch` returns, the outer `Run` loop re-enters `ListAndWatch`, the fresh `List` hits the real apiserver and comes back empty, `DeltaFIFO.Replace` synthesizes the tombstone, and `processDeltas` hands it to steve's `Store.Delete`.
- Asserts the SQLite row is actually gone after the re-list.

Before the fix the test fails with the exact error from the issue:
```
level=error msg="Error in Store.Delete for type fruits.cattle.io_v1_Banana: transaction: object does not implement the Object interfaces"
```
After the fix it passes in ~1.5s.

Also verified: existing unit tests in `pkg/sqlcache/store` and `pkg/sqlcache/informer` still pass, full `tests/integration` suite still passes.

### DIFF
--- a/pkg/sqlcache/informer/listoption_indexer.go
+++ b/pkg/sqlcache/informer/listoption_indexer.go
@@ -361,16 +361,8 @@ func (l *ListOptionIndexer) notifyEventModified(key string, obj any, _ db.TxClie
 	return l.notifyEvent(watch.Modified, oldObj, obj)
 }
 
-func (l *ListOptionIndexer) notifyEventDeleted(key string, obj any, _ db.TxClient) error {
-	oldObj, exists, err := l.GetByKey(key)
-	if err != nil {
-		return fmt.Errorf("error getting old object: %w", err)
-	}
-
-	if !exists {
-		return fmt.Errorf("old object %q should be in store but was not", key)
-	}
-	return l.notifyEvent(watch.Deleted, oldObj, obj)
+func (l *ListOptionIndexer) notifyEventDeleted(_ string, obj any, _ db.TxClient) error {
+	return l.notifyEvent(watch.Deleted, obj, obj)
 }
 
 func (l *ListOptionIndexer) notifyEvent(eventType watch.EventType, old any, current any) error {
--- a/pkg/sqlcache/informer/relist_test.go
+++ b/pkg/sqlcache/informer/relist_test.go
@@ -1,0 +1,173 @@
+package informer
+
+import (
+	"context"
+	"strconv"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/rancher/steve/pkg/sqlcache/db"
+	"github.com/rancher/steve/pkg/sqlcache/encryption"
+	"github.com/rancher/steve/pkg/sqlcache/store"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/cache"
+)
+
+// TestRelistAfterWatchError exercises the path the user actually hits in
+// rancher/rancher#55228: the reflector's watch dies, the reflector re-lists
+// from a source whose contents have changed during the gap, and steve's
+// SQLite cache must end up reflecting the new state.
+//
+// This is a behavioral test, deliberately not pinned to any particular
+// client-go FIFO. On client-go v0.34/v0.35 (DeltaFIFO) the deletion travels
+// through processDeltas → Store.Delete(DeletedFinalStateUnknown). On v0.36+
+// with AtomicFIFO it travels through processReplacedAllInfo → Store.Replace.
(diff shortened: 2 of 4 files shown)


vvlisn:
Could you confirm whether this fix is planned for the next Rancher 2.13 patch release?

crobby (approved):

prachidamle:
@tomleb do we need this fix in main branch - I see it is for steven 0.8 and 0.7?

gehrkefc (approved):

gehrkefc:
> @tomleb do we need this fix in main branch - I see it is for steven 0.8 and 0.7?

I think we need a forward port to main.

PR #1214, opened 2026-05-29
refactor(sqlcache): replace imperative SQL generation with recursive expression tree

gehrkefc:
# Nobody asked for this, it's just me testing an idea that came to mind sometime ago.

## Summary

Replaces the imperative string-concatenation SQL generation in `pkg/sqlcache/informer/sqlgenerator.go` with a fully functional, recursive expression tree + pipeline-based compiler.

### Layer 1: Expression Tree (`pkg/sqlcache/sqlexpr`)
- Core `Expr` interface: `Resolve() (string, []any)`
- Leaf nodes: `Raw`, `Col`, `Param`
- Logic combinators: `And`, `Or`, `Not`, `FlatAnd`, `FlatOr`
- Query nodes: `Select`, `Join`, `OrderBy`, `WithClause`, `CountWrap`
- Each node resolves itself recursively — no mutable accumulators

### Layer 2: Functional Compiler (`pkg/sqlcache/sqlgen`)
- **`QueryState`**: immutable accumulator with copy-on-write methods (clones slices/maps on every mutation)
- **`Transform`**: `func(QueryState) → (QueryState, error)` — pure function
- **`Pipeline`**: composes transforms left-to-right
- **Op registries**: `FieldOps`, `LabelOps`, `ProjectOps` — map-based dispatch, no switch statements

### Layer 3: Informer Bridge (`pkg/sqlcache/informer`)
- `constructQuery` and `ListSummaryFields` call the new compiler
- Old internal functions retained only for test compatibility

## How it works

```go
pipeline := Pipeline(
    WithUnboundSortLabels(lo),
    WithFilters(lo, registry, false),
    WithProjects(lo.ProjectsOrNamespaces, registry),
    WithNamespace(namespace),
    WithPartitions(namespace, partitions),
    WithSort(lo.SortList, registry, namespaced),
    WithPagination(lo.Pagination),
)
final, err := pipeline(initial)
```

## Adding a new operator

One function + one map entry. Example:

```go
func fieldBetween(col sqlexpr.Expr, f sqltypes.Filter) (sqlexpr.Expr, error) {
    return sqlexpr.Raw{
        SQL:    "? BETWEEN ? AND ?",
        Params: []any{col, f.Matches[0], f.Matches[1]},
    }, nil
}

// Register:
var FieldOps = map[sqltypes.Op]FieldOpHandler{
    // ...existing ops...
    sqltypes.Between: fieldBetween,
}
```

No switch statement to find. No 1000-line file to navigate.

## Key design decisions

- **Immutable state**: `QueryState` copy-on-write prevents shared backing array bugs
- **Two concerns separated**: label join allocation (in `EnsureLabelJoin`) vs expression building (in op handlers)
- **Pure op handlers**: field handlers take `(col, filter) → Expr`, label handlers take `(filter, ctx) → Expr`
- **Deterministic ordering**: label aliases allocated in filter scan order, maintained via ordered slice

## Test Results

- ✅ All 42 informer unit tests pass
- ✅ All 11 sqlgen unit tests pass  
- ✅ All 28 sqlexpr unit tests pass
- ✅ Integration tests pass (k3d cluster)
- ✅ Factory tests pass

## File structure

```
pkg/sqlcache/sqlgen/
├── pipeline.go          — QueryState + Transform + Pipeline (core abstractions)
├── ops_field.go         — FieldOps registry (8 handlers, no switch)
├── ops_label.go         — LabelOps registry (10 handlers, no switch)
├── ops_project.go       — ProjectOps registry (4 handlers)
├── transforms.go        — WithFilters, WithSort, WithProjects, etc.
├── compile.go           — CompileListQuery/CompileSummaryListQuery (pipeline orchestration)
├── compile_summary.go   — Summary field query builder
├── compile_partition.go — Partition RBAC compilation
├── field_registry.go    — Field path → column resolution
├── helpers.go           — Shared utilities
└── errors.go
```

## Notes

Old internal functions (`compileQuery`, `constructSummaryQueryForField`, etc.) are retained in `sqlgenerator.go` because existing tests call them directly. The production code paths are fully migrated to the new functional compiler.


pkg/sqlcache/sqlexpr/expr_test.go
```diff
@@ -0,0 +1,337 @@
+package sqlexpr
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestRaw(t *testing.T) {
```
ericpromislow, line 8: I would put all the tests in alphabetical order just to impose some kind of order -- conceptual gets too confusing here

ericpromislow (commented):
There's too much here to do a detailed review. But the whole SQL backend of VAI has been ad-hoc from the beginning so it's worthwhile replacing it with formal layers based on a structured architecture.

Probably draw up a diagram for how the SQL code is now generated and let's move forward with this

PR #1225, opened 2026-06-08
Fall back to apiOp.Namespace in Store.Update for PATCH

rohitsuse:
Issue : https://github.com/rancher/rancher/issues/55391

Added namespace assignment in `update` function like it was in the `create` function above. This fixes the problem @richard-cox was facing.


ericpromislow (approved):
LG

crobby (approved):

PR #1227, opened 2026-06-09
Add tests for cacheForwithdeps

rohitsuse:
Issue : https://github.com/rancher/rancher/issues/55180

### DIFF
--- a/pkg/stores/sqlproxy/proxy_store_test.go
+++ b/pkg/stores/sqlproxy/proxy_store_test.go
@@ -277,6 +277,91 @@ func TestNewProxyStore(t *testing.T) {
 	}
 }
 
+func TestCacheForWithDepsSecretGatesProjectsByPresence(t *testing.T) {
+	secretSchema := &types.APISchema{
+		Schema: &schemas.Schema{
+			ID: "secret",
+			Attributes: map[string]interface{}{
+				"group":      "",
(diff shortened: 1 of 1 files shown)


pkg/stores/sqlproxy/proxy_store_test.go
```diff
@@ -277,6 +277,91 @@ func TestNewProxyStore(t *testing.T) {
...
+		cf.EXPECT().CacheFor(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), secretGVK, gomock.Any(), gomock.Any()).Return(secretCache, nil)
+
+		// both released on cleanup, in reverse order
+		cf.EXPECT().DoneWithCache(secretCache)
+		cf.EXPECT().DoneWithCache(projectCache)
+
+		inf, done, err := s.cacheForWithDeps(context.Background(), apiOp, secretSchema)
+		assert.NoError(t, err)
+		assert.Equal(t, secretCache, inf)
+		assert.NotNil(t, done)
```
ericpromislow, line 339: `assert` should be `require` becaue if `done` is nil the next line will panic. Same at line 360

ericpromislow (changes requested):
Just do a `require` to avoid unnecessary panics

PR #1229, opened 2026-06-18
Detect aggregated APIs and add watch-list policy

rohitsuse:
## Problem

- client-go ≥ v0.35 (k8s 1.35) enables `WatchListClient` by default: informers open a watch with `sendInitialEvents=true` and finish initial sync only after a `BOOKMARK` annotated `k8s.io/initial-events-end=true`.
- Built-in resources / CRDs emit it; **aggregated (extension) APIs may not** — e.g. Antrea's `controlplane.antrea.io` → the reflector hangs, `HasSynced` never becomes true.
- The cluster cache's `OnSchemas` holds the write lock across a 15-min `WaitForCacheSync`, so **one stuck informer freezes every reader** (`/v1/counts`, `fleetworkspaces`). (rancher/rancher#55606)

## Fix

1. **Detect** — tag schemas whose group/version is served by an `APIService` with an external `.spec.service` (built-ins/CRDs untagged).
2. **Policy** (`pkg/watchlist`) — disable watch-list for aggregated, non-whitelisted APIs; built-ins keep it. Fails safe (LIST+WATCH always completes); a `Whitelist` lets a verified aggregated API opt back in.
3. **Apply** in both informer paths — clustercache (`noWatchListClient`) and sqlcache (`disableWatchList`); the reflector falls back to LIST+WATCH for that informer.
4. **Harden the lock** — `OnSchemas` releases the write lock before `WaitForCacheSync`, so a never-syncing informer leaves only its own cache empty instead of blocking all reads (the actual #55606 outage fix).

### DIFF
--- a/pkg/sqlcache/informer/factory/informer_factory.go
+++ b/pkg/sqlcache/informer/factory/informer_factory.go
@@ -57,7 +57,7 @@ type guardedInformer struct {
 	wg sync.WaitGroup
 }
 
-type newInformer func(ctx context.Context, client dynamic.ResourceInterface, fields map[string]informer.IndexedField, externalUpdateInfo *sqltypes.ExternalGVKUpdates, selfUpdateInfo *sqltypes.ExternalGVKUpdates, transform cache.TransformFunc, gvk schema.GroupVersionKind, db db.Client, shouldEncrypt bool, namespace bool, watchable bool, gcKeepCount int) (*informer.Informer, error)
+type newInformer func(ctx context.Context, client dynamic.ResourceInterface, fields map[string]informer.IndexedField, externalUpdateInfo *sqltypes.ExternalGVKUpdates, selfUpdateInfo *sqltypes.ExternalGVKUpdates, transform cache.TransformFunc, gvk schema.GroupVersionKind, db db.Client, shouldEncrypt bool, namespace bool, watchable bool, disableWatchList bool, gcKeepCount int) (*informer.Informer, error)
 
 type Cache struct {
 	informer.ByOptionsLister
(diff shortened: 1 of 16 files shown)


pkg/clustercache/controller_test.go
```diff
@@ -0,0 +1,81 @@
...
+	coll := schema.NewCollection(ctx, types.EmptyAPISchemas(), nil)
+	coll.Reset(map[string]*types.APISchema{s.ID: s})
+
+	// OnSchemas builds the never-syncing informer and blocks in WaitForCacheSync.
+	go h.OnSchemas(coll)
+	time.Sleep(500 * time.Millisecond) // let OnSchemas reach the sync wait
+
+	done := make(chan struct{})
+	go func() {
+		h.List(k8sschema.GroupVersionKind{Version: "v1", Kind: "ConfigMap"})
```
tomleb: I think we should test if the List calls happen to be on the resource that is no syncing too. What is the expected behavior?

pkg/sqlcache/informer/factory/informer_factory.go
```diff
@@ -57,7 +57,7 @@ type guardedInformer struct {
 	wg sync.WaitGroup
 }
 
-type newInformer func(ctx context.Context, client dynamic.ResourceInterface, fields map[string]informer.IndexedField, externalUpdateInfo *sqltypes.ExternalGVKUpdates, selfUpdateInfo *sqltypes.ExternalGVKUpdates, transform cache.TransformFunc, gvk schema.GroupVersionKind, db db.Client, shouldEncrypt bool, namespace bool, watchable bool, gcKeepCount int) (*informer.Informer, error)
+type newInformer func(ctx context.Context, client dynamic.ResourceInterface, fields map[string]informer.IndexedField, externalUpdateInfo *sqltypes.ExternalGVKUpdates, selfUpdateInfo *sqltypes.ExternalGVKUpdates, transform cache.TransformFunc, gvk schema.GroupVersionKind, db db.Client, shouldEncrypt bool, namespace bool, watchable bool, disableWatchList bool, gcKeepCount int) (*informer.Informer, error)
```
tomleb, line 60: One day I'll have Claude move these to a option struct...

tomleb (commented):
LGTM only one test missing that I think should be there (in the comments).

Would be great to have an integration test with a custom apiserver but I don't think we have anything to build that quickly yet, so we can pass for now.

pkg/clustercache/controller_test.go
rohitsuse: I think it should return the list that is added in the store. Added a test for it.

tomleb (approved):

PR #1231, opened 2026-06-23
Non-sql mode is no longer an option, so remove all the non-sql code.

ericpromislow:
# DNM

## This is for 2.16

Ref [#53996](https://github.com/rancher/rancher/issues/53996)

The integration tests that used `nondeterministictoken` and `nondeterministicint` now just use
`pagesize` and `page` because we expect clients will be tracking page numbers of requests.

The `--sql-cache` option is still accepted but now triggers a warning message that it is no longer needed.
I'll leave it for Product to decide when it should no longer be allowed.

Note that I'm also working on applying this change to rancher/rancher as well -- it affects the way r/r invokes the steve server, as well as the legacy integration tests.

### DIFF
--- a/tests/integration/summary_test.go
+++ b/tests/integration/summary_test.go
@@ -69,16 +69,12 @@ func (i *IntegrationSuite) TestSummary() {
 	// Cleanup common manifests after all tests complete
 	defer i.doManifestReversed(ctx, commonManifestsFile, i.doDelete)
 
-	// Run SQL mode first, then non-SQL mode sequentially
-	i.Run("SQL", func() {
-		i.runSummaryTest(ctx, true, gvrs)
-	})
-	i.Run("NonSQL", func() {
-		i.runSummaryTest(ctx, false, gvrs)
+	i.Run("Summary Tests", func() {
+		i.runSummaryTest(ctx, gvrs)
 	})
 }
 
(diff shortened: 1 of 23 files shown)


tests/integration/summary_test.go
```diff
@@ -168,30 +154,23 @@ func (i *IntegrationSuite) readSummaryTestConfig(testFile string) SummaryTestCon
...
-			}
-			if strings.Contains(query, "nondeterministicint") {
-				query = strings.Replace(query, "nondeterministicint", lastRevision, 1)
-			}
+			//if strings.Contains(query, "nondeterministictoken") {
+			//	query = strings.Replace(query, "nondeterministictoken", lastContinueToken, 1)
+			//}
+			//if strings.Contains(query, "nondeterministicint") {
+			//	query = strings.Replace(query, "nondeterministicint", lastRevision, 1)
+			//}
```
gehrkefc, line 172: Any reason to comment instead of remove?

gehrkefc (approved):

tests/integration/summary_test.go
ericpromislow: Nope - missed that

tomleb:
@ericpromislow @gehrkefc just reminder this is for 2.16 and main currently is for 2.15 (I know it's in draft for now, just making sure :) )

crobby (commented):
Seems pretty thorough, tests and docs updated.

PR #1235, opened 2026-07-10
Fix panic when sorting non-SQL lists by a boolean field (rancher/rancher#55593)

### LINKED ISSUE #55593: [BUG] Steve panics with "interface conversion: interface {} is bool, not string" when sorting a list by a boolean field (metadata.state)
**Rancher Server Setup**
- Rancher version: 2.14.2
- Installation option (Docker install/Helm Chart):
   - Helm install k3s version v1.34.8+k3s1 (fb21251e)
- Proxy/Cert Details:
   No proxy. TLS terminated at an external LB with a Let's Encrypt cert
**Information about the Cluster**
- Kubernetes version: v1.34.7+rke2r1
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster?  Imported (RKE2 cluster) 
   - Infrastructure: VMware
 
**User Information**
- What is the role of the user logged in? Cluster Owner
  - If custom, define the set of permissions: N/A

**Describe the bug**

Opening the Cluster Explorer for certain downstream clusters renders a blank page.
The cluster itself is healthy — kubectl through the Rancher proxy works, the
cattle-cluster-agent is connected, and nodes are Ready. Only the Steve-backed
dashboard views fail.

The rancher pod logs show every affected list request panicking:

  http: panic serving 10.x.x.x: interface conversion: interface {} is bool, not string
    github.com/rancher/steve/pkg/stores/partition/listprocessor.SortList(...)
          steve@v0.8.24/pkg/stores/partition/listprocessor/processor.go:383
    github.com/rancher/steve/pkg/stores/partition.(*Store).List(...)
          steve@v0.8.24/pkg/stores/partition/store.go:209
    github.com/rancher/steve/pkg/stores/proxy.(*unformatterStore).List(...)
    github.com/rancher/steve/pkg/stores/metrics.(*Store).List(...)

The panic kills the request handler, so the HTTP call never returns. The dashboard
waits on /v1/<type> and /v1/counts, those hang, and the page stays blank.

**To Reproduce**

1. Rancher v2.14.2 (steve v0.8.24) with at least one imported downstream cluster.
2. In the Cluster Explorer for that cluster, open a resource list (e.g. Pods or
   Services) and click the "State" column header to sort by it.

   Or hit the Steve API directly with a sort on a boolean field:

     GET /k8s/clusters/<cluster-id>/v1/pods?sort=metadata.state.error

in my testing, whether a given list panics depended on the actual objects returned (it reproduced reliably on some clusters' Pods/Services lists but not others)

**Result**
Blank page

**Expected Result**
Cluster Dashboard

rohitsuse:
## Problem

Sorting a non-SQL list by a boolean field panics and kills the request handler, so Cluster Explorer renders blank:

```
http: panic serving ...: interface conversion: interface {} is bool, not string
    listprocessor.SortList(...) processor.go:383
```

## Cause

`SortList` did an unchecked `val.(string)` on values from `data.GetValueFromAny`, but sortable fields aren't always strings — e.g. `metadata.state.error` (a bool, behind the dashboard's **State** column). Regressed in #618, which dropped the earlier `convert.ToString(...)`. Non-SQL path only; the SQL cache sorts via `ORDER BY`.

## Fix

Restore `convert.ToString(val)`; add a `TestSortList` regression case for boolean/missing fields.

Fixes rancher/rancher#55593

### DIFF
--- a/pkg/stores/partition/listprocessor/processor.go
+++ b/pkg/stores/partition/listprocessor/processor.go
@@ -380,7 +380,7 @@ func SortList(list []unstructured.Unstructured, s Sort) []unstructured.Unstructu
 		for j, fields := range s.Fields {
 			val, ok := data.GetValueFromAny(list[i].Object, fields...)
 			if ok {
-				td.stringValues[j] = val.(string)
+				td.stringValues[j] = convert.ToString(val)
 			} else {
 				logrus.Debugf("Failed to walk list item %d with fields %s", i, fields)
 				td.stringValues[j] = ""
--- a/pkg/stores/partition/listprocessor/processor_test.go
+++ b/pkg/stores/partition/listprocessor/processor_test.go
@@ -2665,6 +2665,61 @@ func TestSortList(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "sort by boolean field metadata.state.error",
+			objects: []unstructured.Unstructured{
+				{
+					Object: map[string]interface{}{
+						"kind": "pod",
+						"metadata": map[string]interface{}{
+							"name": "alpha",
+							"state": map[string]interface{}{
+								"error": true,
+							},
+						},
+					},
+				},
+				{
+					Object: map[string]interface{}{
+						"kind": "pod",
+						"metadata": map[string]interface{}{
+							"name": "bravo",
+							"state": map[string]interface{}{
+								"error": false,
+							},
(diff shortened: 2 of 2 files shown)


crobby (approved):
Looks good

tomleb (approved):

PR #1237, opened 2026-07-23
Derive generateName prefix from Kind instead of schema ID

gehrkefc:
## Problem

Steve's generic `Create` path defaults `metadata.generateName` to the first character of the schema ID when no name is provided:

```go
input.SetNested(schema.ID[0:1]+"-", "metadata", "generateName")
```

Steve schema IDs are group-qualified (`management.cattle.io.user`), so the first character is `m`, producing names like `m-rfsrk` for new users instead of the expected `u-` prefix. Norman derives the prefix from the bare kind (`user` → `u-`).

See rancher/rancher#56214.

## Change

- Add `converter.GenerateNamePrefix(kind)`, mirroring norman's initials convention (`User` → `u-`, `GlobalRole` → `gr-`).
- Both the `proxy` and `sqlproxy` stores now derive the prefix from `attributes.GVK(schema).Kind`, falling back to the previous `schema.ID[0:1]` behavior for schemas without a Kind.
- Update the two existing `Create` tests and add `GenerateNamePrefix` unit tests.

## Testing

```
go test ./pkg/schema/converter/ ./pkg/stores/proxy/ ./pkg/stores/sqlproxy/
```

All pass.

### DIFF
--- a/pkg/schema/converter/k8stonorman.go
+++ b/pkg/schema/converter/k8stonorman.go
@@ -4,6 +4,7 @@ package converter
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/rancher/apiserver/pkg/types"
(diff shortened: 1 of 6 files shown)


crobby (commented):
The notes mention that existing incorrectly prefixed objects would need either a migration or a consumer change.  Are either of those planned?  Do we need to include a migration of some sort for this to fix existing problematic resources?

gehrkefc:
> The notes mention that existing incorrectly prefixed objects would need either a migration or a consumer change. Are either of those planned? Do we need to include a migration of some sort for this to fix existing problematic resources?

I don't think we can change the existing resources. But I can check and maybe have a follow-up PR only for that. WDYT?

crobby:
> > The notes mention that existing incorrectly prefixed objects would need either a migration or a consumer change. Are either of those planned? Do we need to include a migration of some sort for this to fix existing problematic resources?
> 
> I don't think we can change the existing resources. But I can check and maybe have a follow-up PR only for that. WDYT?

I was just making sure that we're not breaking anything that would be fixed by a migration.  Sounds like there's no breakage, so I'm ok with this.

crobby (approved):

tomleb (approved):

PR #1252, opened 2026-07-30
use CollectT instead of t inside EventuallyWithT closures

crobby:
Calling FailNow on the outer *testing.T from there permanently fails the subtest even if a later retry succeeds.

This fix should allow us to successfully bump testify in https://github.com/rancher/steve/pull/794


tomleb (approved):
Classic mistake

ericpromislow (approved):
LGTM

PR #1253, opened 2026-07-30
Remove non-SQL cache code

rohitsuse:


ericpromislow:
As of the first commit, this PR is deleting some code that `rancher/rancher` still depends on - see https://github.com/rancher/rancher/issues/53996#issuecomment-5136597802 for a list

I didn't find any instances in the other dependencies, but then `steve` sits near the top of the dependency food-chain, just below `rancher/rancher`

PR #1256, opened 2026-08-03
sqlcache: tolerate DeletedFinalStateUnknown in Store.Delete

tomleb:
# Issue https://github.com/rancher/rancher/issues/56397

Forward-port of #1209 (release/v0.8) to `main`.

## Symptom / Fix

Same as #1209: when the reflector loses its watch and falls back to a fresh List, client-go's FIFO deletion-detection can hand `Store.Delete` a `cache.DeletedFinalStateUnknown` tombstone. The after-delete hook's `meta.Accessor` call errors on the wrapper (it doesn't implement `metav1.Object`), rolling back the SQL DELETE and leaving stale rows in the cache.

`Store.Delete` now unwraps the tombstone before running after-delete hooks, and `notifyEventDeleted` drops the now-unnecessary `GetByKey` round-trip (which had its own `!exists` error path that would trigger the same rollback failure mode on a missing cache entry).

## Note on relevance to `main`

`main` has since bumped `k8s.io/client-go` to v0.36.0 (#1174), where the `AtomicFIFO` feature gate defaults to `true`. In that mode, informer relists go through `processReplacedAllInfo` → `Store.Replace` (bulk delete+insert) rather than emitting a per-object `Deleted` delta through `Store.Delete`, so this specific bug is not currently reachable on `main` by default. This PR is a defensive/consistency forward-port in case that feature gate is ever disabled, or for any other codepath that reaches `Store.Delete` with a tombstone.

## Tests

Same coverage as #1209: `TestRelistAfterWatchError` (pkg/sqlcache/informer/relist_test.go) reproduces the tombstone-through-relist path end-to-end, plus new `TestDeleteTombstone` unit tests in pkg/sqlcache/store/store_test.go.

Verified locally: `go build ./...` and `go test ./pkg/sqlcache/...` (including the envtest-backed integration suite) pass.

### DIFF
--- a/pkg/sqlcache/informer/listoption_indexer.go
+++ b/pkg/sqlcache/informer/listoption_indexer.go
@@ -361,16 +361,8 @@ func (l *ListOptionIndexer) notifyEventModified(key string, obj any, _ db.TxClie
 	return l.notifyEvent(watch.Modified, oldObj, obj)
 }
 
-func (l *ListOptionIndexer) notifyEventDeleted(key string, obj any, _ db.TxClient) error {
-	oldObj, exists, err := l.GetByKey(key)
-	if err != nil {
-		return fmt.Errorf("error getting old object: %w", err)
-	}
-
-	if !exists {
-		return fmt.Errorf("old object %q should be in store but was not", key)
-	}
-	return l.notifyEvent(watch.Deleted, oldObj, obj)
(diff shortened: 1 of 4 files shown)


crobby (approved):
looks like a clean port

ericpromislow (approved):
LGTM

I also approve of renaming the type `noWatchListListWatch` to
the more golike `noWatchListListWatcher`, but this happened previously to this PR,
but is the source of the only diff between the first PR and this forward-port.

PR #1257, opened 2026-08-03
[release/v0.9] sqlcache: tolerate DeletedFinalStateUnknown in Store.Delete

tomleb:
# Issue https://github.com/rancher/rancher/issues/55228

Forward-port of #1209 (release/v0.8) to `release/v0.9`.

## Symptom / Fix

Same as #1209: when the reflector loses its watch and falls back to a fresh List, client-go's FIFO deletion-detection can hand `Store.Delete` a `cache.DeletedFinalStateUnknown` tombstone. The after-delete hook's `meta.Accessor` call errors on the wrapper (it doesn't implement `metav1.Object`), rolling back the SQL DELETE and leaving stale rows in the cache.

`Store.Delete` now unwraps the tombstone before running after-delete hooks, and `notifyEventDeleted` drops the now-unnecessary `GetByKey` round-trip (which had its own `!exists` error path that would trigger the same rollback failure mode on a missing cache entry).

## Note on relevance to `release/v0.9`

`release/v0.9` is already on `k8s.io/client-go` v0.36.0, same as `main`, where the `AtomicFIFO` feature gate defaults to `true` and relists go through `Store.Replace` rather than per-object `Store.Delete`. This bug is not currently reachable by default on this branch either. This PR is a defensive/consistency forward-port in case that feature gate is ever disabled, or for any other codepath that reaches `Store.Delete` with a tombstone.

## Tests

Same coverage as #1209: `TestRelistAfterWatchError` (pkg/sqlcache/informer/relist_test.go) reproduces the tombstone-through-relist path end-to-end, plus new `TestDeleteTombstone` unit tests in pkg/sqlcache/store/store_test.go.

Verified locally: `go build ./...` and `go test ./pkg/sqlcache/...` (including the envtest-backed integration suite) pass.

### DIFF
--- a/pkg/sqlcache/informer/listoption_indexer.go
+++ b/pkg/sqlcache/informer/listoption_indexer.go
@@ -361,16 +361,8 @@ func (l *ListOptionIndexer) notifyEventModified(key string, obj any, _ db.TxClie
 	return l.notifyEvent(watch.Modified, oldObj, obj)
 }
 
-func (l *ListOptionIndexer) notifyEventDeleted(key string, obj any, _ db.TxClient) error {
-	oldObj, exists, err := l.GetByKey(key)
-	if err != nil {
-		return fmt.Errorf("error getting old object: %w", err)
-	}
-
-	if !exists {
-		return fmt.Errorf("old object %q should be in store but was not", key)
-	}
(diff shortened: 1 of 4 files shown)


crobby (approved):
port looks clean, lgtm

gehrkefc (approved):

PR #1259, opened 2026-08-04
[v0.8] Derive generateName prefix from Kind instead of schema ID

gehrkefc:
Backport of #1237 to `release/v0.8` (Rancher v2.14).

Steve's generic `Create` path defaulted `metadata.generateName` to the first character of the group-qualified schema ID (e.g. `management.cattle.io.user`), producing prefixes like `m-` for new users instead of the expected `u-`. This derives the prefix from the resource Kind, matching norman's convention.

See rancher/rancher#56214.

### DIFF
--- a/pkg/schema/converter/k8stonorman.go
+++ b/pkg/schema/converter/k8stonorman.go
@@ -4,6 +4,7 @@ package converter
 
 import (
 	"fmt"
+	"regexp"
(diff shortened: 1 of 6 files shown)


ericpromislow (approved):
Straightforward

crobby (approved):

PR #1262, opened 2026-08-06
Remove non-SQL cache code

rohitsuse:
Issue : https://github.com/rancher/rancher/issues/53997

### Summary 
- Removes all non SQL codebase from steve. 
- Organised the PR according to commits. So browser by commits to understand the PR faster.


crobby (approved):
I think this looks good.  Even the docs and diagrams are updated.  Mostly pretty straightforward.

ericpromislow (approved):
Reviewing the diff with `git diff -w --color-moved main` shows how this PR is mostly deleting most code and moving a bit around for the `client-getter`

PR #1264, opened 2026-08-06
Use unique temp SQLite path per integration test cache factory

crobby:
Refs rancher/rancher#56459

Each integration test's `server.New()` builds its own `CacheFactory`, backed by the fixed `InformerObjectCacheDBPath`. Opening that path deletes and recreates the file every time, so one test's setup could delete the SQLite file still in use by a previous test's not-yet-stopped informer goroutines - causing sporadic `no such table` / `SQL logic error` failures across unrelated tests in CI.

Fix: give each test's cache factory a unique temp-dir SQLite path (`UseTempDir: true`) instead of sharing the fixed one.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -78,6 +78,15 @@ type Server struct {
 	SQLCache                   bool
 }
 
+// SQLCacheDBPath returns the path to the SQLite database file backing the SQL cache,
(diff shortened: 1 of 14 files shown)


tomleb (approved):
LGTM

PR #1265, opened 2026-08-06
Retry CacheFor when a concurrent schema Stop() resets the cache

crobby:
Refs rancher/rancher#56459

`Store.Reset(gvk)` cancels a GVK's informer context to interrupt any in-flight `CacheFor` wait when a CRD/column definition change requires rebuilding the cache. A concurrent read for the same GVK sees this as a canceled-context error and surfaces a 500, even though the cache is immediately usable again once `Stop()` returns.

Fix: export `factory.ErrCacheReset` so callers can identify this transient, safe-to-retry race, and retry a few times at the `sqlproxy` call site instead of failing the request. `CacheFor`'s own error contract is unchanged (an existing unit test relies on it erroring on this race), so the retry lives one layer up.

### DIFF
--- a/pkg/sqlcache/informer/factory/informer_factory.go
+++ b/pkg/sqlcache/informer/factory/informer_factory.go
@@ -5,6 +5,7 @@ package factory
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"sync"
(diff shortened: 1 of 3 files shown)


tomleb (approved):
Lgtm

This doesn't differentiate between cache factory shutting down and cache reset, not sure we need that though, I think we only shutdown the factory in tests

PR #1283, opened 2026-08-25
[release/v0.5] Move to Go 1.25

crobby:
Issue: https://github.com/rancher/rancher/issues/56875
Supersedes #1277.

Rancher `release/v2.11` is on `go 1.25.0` / `toolchain go1.25.13` and consumes `github.com/rancher/steve v0.5.15`, so `release/v0.5` should be on the same Go level.

#1277 (cherry-picked here as the first commit) only bumps the `toolchain` directive. That is not sufficient on its own:

- The `go` directive stays at `1.23.0`, so the language version is unchanged.
- CI pins golangci-lint `v1.63.4`, which cannot type-check Go 1.25 sources. The v1 line ended at `v1.64.8` (Go 1.24 max), so this requires moving to golangci-lint v2.

Changes on top of #1277:

- `go.mod`: `go 1.23.0` -> `go 1.25.0`.
- `.golangci.json`: migrated to the v2 schema via `golangci-lint migrate`. `linters.disable-all` -> `linters.default: none`, `linters-settings` -> `linters.settings`, `issues.exclude-dirs`/`exclude-rules` -> `linters.exclusions.*`, and `gofmt`/`goimports` moved into the new `formatters` block. The `typecheck` exclusion rule is dropped (not permitted in v2), and `run.timeout` is dropped (v2 disables the timeout by default). The result matches the config already in use on `main`.
- `.github/workflows/ci.yaml`: `golangci/golangci-lint-action` `v6.5.2` -> `v9.3.0` (v6 rejects v2 binaries), pinned to golangci-lint `v2.12.1` with `install-only: true`, since `scripts/validate.sh` invokes `golangci-lint run` itself.

Verified locally with the Go 1.25.14 toolchain:

- `go build ./...` passes.
- `make test` passes.
- `go mod tidy` produces no changes; `go.sum` is untouched.
- `make validate` reports `0 issues`, `all modules verified`, and a clean tree.


tomleb:
@crobby that'd be for after the release right? That's a lot of changes for a Go bump

crobby:
> @crobby that'd be for after the release right? That's a lot of changes for a Go bump

For sure.  No intents of merging this until post-release

ericpromislow (approved):
looks fine

tomleb (approved):

PR #1284, opened 2026-08-25
[release/v0.6] Move to Go 1.25

crobby:
Issue:  https://github.com/rancher/rancher/issues/56875
Supersedes #1276. Companion to #1283, which does the same for `release/v0.5`.

#1276 (cherry-picked here as the first commit) only bumps the `toolchain` directive. That is not sufficient on its own:

- The `go` directive stays at `1.24.0`, so the language version is unchanged.
- CI pins golangci-lint `v1.64.8`, which cannot type-check Go 1.25 sources. The v1 line ended there (Go 1.24 max), so this requires moving to golangci-lint v2.
- CI pins `mockgen@v0.5.2`, which no longer builds under Go 1.25.

Changes on top of #1276:

- `go.mod`: `go 1.24.0` -> `go 1.25.0`.
- `.golangci.json`: migrated to the v2 schema via `golangci-lint migrate`. `linters.disable-all` -> `linters.default: none`, `linters-settings` -> `linters.settings`, `issues.exclude-dirs`/`exclude-rules` -> `linters.exclusions.*`, and `gofmt`/`goimports` moved into the new `formatters` block. The `typecheck` exclusion rule is dropped (not permitted in v2), and `run.timeout` is dropped (v2 disables the timeout by default). The result matches the config already in use on `main`.
- `.github/workflows/ci.yaml`: `golangci/golangci-lint-action` `v6.5.2` -> `v9.3.0` (v6 rejects v2 binaries), pinned to golangci-lint `v2.12.1` with `install-only: true`, since `scripts/validate.sh` invokes `golangci-lint run` itself. `mockgen` `v0.5.2` -> `v0.6.0`.
- Cherry-picked ccca45a1 "Fix race in debouncer refresher (#847)", already on `main`, `release/v0.9` and `release/v0.8`. It applies to `release/v0.6` with no conflicts.
- Replaced the remaining `io/ioutil` calls with their `os` / `io` equivalents.
- Fixed an ordering race in the debounce test helper that arrived with #847.

### On the mockgen bump

`go install go.uber.org/mock/mockgen@v0.5.2` fails under Go 1.25:

```
# golang.org/x/tools/internal/tokeninternal
golang.org/x/tools@v0.22.0/internal/tokeninternal/tokeninternal.go:64:9: invalid array length -delta * delta (constant -256 of type int64)
```

`v0.6.0` builds cleanly, and `go generate ./...` with it produces no diff against the committed mocks.

### On the debounce test helper fix

`mockRefresher.Refresh` in `pkg/debounce/refresher_test.go` signals on `refreshCh` before incrementing `refreshCount`, so a test woken by `waitForRefresh` can read the counter before the increment lands. Every assertion of the form `waitForRefresh(...)` followed immediately by `require.Equal(t, 1, mock.getCount())` is exposed to it.

Under `-race` this reproduces on this branch: `TestContext_Cancellation` failed with `expected: 1, actual: 0` twice in 300 runs, and `TestSchedule_Debounce` failed the same way at `refresher_test.go:79` in a later 300-run batch. With the increment moved ahead of the send, 300 runs are clean.

### Verification

Local runs with the Go 1.25.14 toolchain:

- `go build ./...` passes.
- `make test` passes.
- `make validate` reports `0 issues`, `all modules verified`, and a clean tree.
- `go generate ./...` with mockgen v0.6.0 produces no diff.
- `pkg/debounce` passes 300 runs under `-race`.


ericpromislow (approved):
looks fine

tomleb (approved):

PR #1290, opened 2026-08-28
Add ownership filter

tomleb:
# Issue https://github.com/rancher/rancher/issues/56849

Steve does a single watch via admin privileges per resource and stores that into the SQLite cache. Resources like ext.cattle.io Kubeconfigs and Tokens have special behavior where a user is only able to see its own tokens / kubeconfigs (which is global). To make this additional restriction with Steve, we were adding a SQL filter. We forgot we also need to filter the result of the `/v1/count` endpoint.

# Summary of changes

Abstract away a bit this additional "ownership filter" (see pkg/resources/ownership). Introduces a new interface with two methods: one to add a SQLite filter, another to do basic filtering on the object. (Note: the count endpoint doesn't use the SQLite database for its data)

Introduces a UserIDLabelFilter type that implements this interface. It is only used in tests in steve. 

The SQLite filter is removed from steve and moved to rancher/rancher closer to where the stores are set. That seems like the more sensible thing to do. (I'm happy to leave it in steve though if anyone wants it here).

# Tests

I added an integration test using the UserIDLabelFilter to test for regression. We don't have infrastructure code yet to deploy an actual extension apiserver, so I'm testing with a CRD instead of a custom made "Token" API.

### DIFF
--- a/pkg/resources/ownership/ownership.go
+++ b/pkg/resources/ownership/ownership.go
@@ -1,0 +1,72 @@
+// Package ownership provides a single place to describe "who owns what" for
+// resources whose visibility is scoped per-user by something other than
+// Kubernetes RBAC (eg: ext.cattle.io Tokens and Kubeconfigs, which are
+// served by rancher's extension apiserver and filtered by a
+// cattle.io/user-id label rather than by SubjectAccessReview).
+//
+// Steve itself has no hardcoded knowledge of which GVKs need this -- that's
+// rancher-specific domain knowledge that rancher/rancher's pkg/ext owns (the
+// actual GVKs and label constants live in pkg/ext/stores/tokens and
+// pkg/ext/stores/kubeconfig). rancher/rancher registers a Filter for each
+// such GVK during extension-apiserver startup, in
+// pkg/ext/stores/install.go's InstallStores, right where those stores are
+// otherwise wired up.
+package ownership
+
+import (
+	"sync"
+
+	"github.com/rancher/steve/pkg/sqlcache/sqltypes"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apiserver/pkg/authentication/user"
+)
+
+// Filter scopes visibility of a resource to its owning user, for resources
+// where visibility isn't determined by RBAC alone.
+type Filter interface {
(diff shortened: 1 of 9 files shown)


crobby (approved):
I think moving from steve to r/r is probably fine in cases like this.  We just have to remember (or be sure to use the new pattern and repeat) to apply the filter if we have similar cases in the future.

pkg/resources/ownership/ownership.go
```diff
@@ -0,0 +1,72 @@
+// Package ownership provides a single place to describe "who owns what" for
```
ericpromislow, line 1: Just a nit -- great comment, should probably go after the `import` statement just to be consistent with typical style

ericpromislow (approved):
Looks fine. Maybe just move that comment block as requested

pkg/resources/ownership/ownership.go
tomleb: The comment describes the ownership package and in Go the usual location for a package-level comment is above the `package <package>` line (see https://go.dev/doc/comment#package)

ericpromislow: ok

pkg/stores/sqlproxy/proxy_store.go
```diff
@@ -48,9 +48,11 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	rtschema "k8s.io/apimachinery/pkg/runtime/schema"
```
crobby: Do we need this import alias here?  Isn't it a dupe of the one above it?

crobby (commented):
lgtm, just the one question about imports

pkg/stores/sqlproxy/proxy_store.go
tomleb: Well spotted, I have removed both and changed rtschema to `k8sschema` which is already used in our codebase. The issue here is that `schema` is used as a name for variables all over so sometimes we need a package name for the k8s schema.

crobby (approved):

PR #1294, opened 2026-09-03
fix per-event allocation hot paths driving agent memory growth

susesamu:
issue: https://github.com/rancher/rancher/issues/57134

### DIFF
--- a/pkg/resources/counts/counts.go
+++ b/pkg/resources/counts/counts.go
@@ -115,19 +115,12 @@ func (s *Store) List(apiOp *types.APIRequest, schema *types.APISchema) (types.AP
 // Watch creates a watch for the Counts schema. This returns only the counts which have changed since the watch was established
 func (s *Store) Watch(apiOp *types.APIRequest, schema *types.APISchema, w types.WatchRequest) (chan types.APIEvent, error) {
 	var (
-		result      = make(chan Count, 100)
 		counts      map[string]ItemCount
 		gvkToSchema = map[schema2.GroupVersionKind]*types.APISchema{}
 		countLock   sync.Mutex
+		changed     = map[string]struct{}{}
+		wake        = make(chan struct{}, 1)
 	)
-
-	go func() {
-		<-apiOp.Context().Done()
-		countLock.Lock()
-		close(result)
-		result = nil
-		countLock.Unlock()
-	}()
 
 	counts = s.getCount(apiOp).Counts
 	for id := range counts {
@@ -142,10 +135,6 @@ func (s *Store) Watch(apiOp *types.APIRequest, schema *types.APISchema, w types.
 	onChange := func(add bool, gvk schema2.GroupVersionKind, _ string, obj, oldObj runtime.Object) error {
 		countLock.Lock()
 		defer countLock.Unlock()
-
-		if result == nil {
-			return nil
-		}
 
 		schema := gvkToSchema[gvk]
 		if schema == nil {
@@ -181,16 +170,38 @@ func (s *Store) Watch(apiOp *types.APIRequest, schema *types.APISchema, w types.
 		}
 
 		counts[schema.ID] = itemCount
-		changedCount := map[string]ItemCount{
-			schema.ID: *itemCount.DeepCopy(),
-		}
+		changed[schema.ID] = struct{}{}
 
-		result <- Count{
-			ID:     "count",
-			Counts: changedCount,
+		select {
+		case wake <- struct{}{}:
+		default:
 		}
 
 		return nil
+	}
+
+	snapshot := func() (Count, bool) {
+		countLock.Lock()
+		defer countLock.Unlock()
+
+		if len(changed) == 0 {
+			return Count{}, false
+		}
+
+		changedCounts := make(map[string]ItemCount, len(changed))
+		for id := range changed {
+			itemCount, ok := counts[id]
+			if !ok {
+				continue
+			}
+			changedCounts[id] = *itemCount.DeepCopy()
+		}
+		clear(changed)
+
+		return Count{
+			ID:     "count",
+			Counts: changedCounts,
+		}, true
 	}
 
 	s.ccache.OnAdd(apiOp.Context(), func(gvk schema2.GroupVersionKind, key string, obj runtime.Object) error {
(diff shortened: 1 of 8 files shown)


pkg/resources/counts/counts.go
```diff
@@ -181,18 +170,40 @@ func (s *Store) Watch(apiOp *types.APIRequest, schema *types.APISchema, w types.
...
-			Counts: changedCount,
+		select {
+		case wake <- struct{}{}:
+		default:
 		}
 
 		return nil
 	}
 
+	snapshot := func() (Count, bool) {
```
brandond: Consider changing this to return `*Count`? Otherwise you're still allocating a `Count{}` even if there are no changes.

pkg/resources/counts/counts.go
```diff
@@ -181,18 +170,40 @@ func (s *Store) Watch(apiOp *types.APIRequest, schema *types.APISchema, w types.
...
+			return Count{}, false
+		}
+
+		changedCounts := make(map[string]ItemCount, len(changed))
+		for id := range changed {
+			itemCount, ok := counts[id]
+			if !ok {
+				continue
+			}
+			changedCounts[id] = *itemCount.DeepCopy()
```
brandond: Is ItemCount mutated by consumers? Otherwise I'm confused why it's being deepcopied. If the intent is to reduce allocations, consider moving the deepcopy into callers that want to modify the shared object, instead of preemptively copying it here.

pkg/resources/counts/counts.go
susesamu: oh, I thought both are 0 allocs, since `Count{}` is a zero value struct returned in registers, not a heap allocation.

I can change to `*Count` to keep nil as the idiom, but wouldn't it have 1 alloc worse? the struct escapes, then still gets boxed into `APIObject.Object`, and it only breaks even if `toAPIObject` is also converted to take a pointer

pkg/resources/counts/counts.go
susesamu: hmm consumers don't mutate, the producer does. `addCounts`/`removeCounts` write into `itemCount.Namespaces[ns]` and `Summary.States` in place on the maps held in counts, so those maps keep changing as further events arrive. handing them out directly would let the next cluster event mutate a Count that subscribe is still serializing onto the websocket?

 obs: the modifier is upstream in onChange, not downstream.  also Emit time is the latest the copy can be taken, one copy per 5s window instead of one per event. I added a comment for this.

pkg/resources/counts/counts.go
brandond: I honestly don't know off the top of my head which is better at preventing allocation and heap escape. My usual instinct (and idiomatic go code style) is to allow returning nil instead of an empty struct on error or not-ok path.

I guess building this package with `go build -gcflags="-m"` would show for sure?

pkg/resources/counts/counts.go
brandond: > handing them out directly would let the next cluster event mutate a Count that subscribe is still serializing onto the websocket?

That sounds like an x-y problem. If the result is going to be serialized, and we want to freeze the current point-in-time view of the resources we are returning, would it be better to do that serialization now while we hold a lock, instead of deepcopying all these objects just so that some other goroutine isn't changing them while they are marshalled?

pkg/resources/counts/counts.go
susesamu:
```
$ go build -gcflags='-m' ./pkg/resources/counts/
pkg/resources/counts/counts.go:192:14: func literal escapes to heap
pkg/resources/counts/counts.go:200:24: make(map[string]ItemCount, len(changed)) escapes to heap
```
Line 197 is return `Count{}, false` and it doesn't appear, nothing escapes.

confirming the converse with `testing.AllocsPerRun`, end to end through `toAPIObject`:

`(Count{}, bool)` current: 2 allocs/op
`*Count`, deref at the call site: 3 allocs/op
`*Count + toAPIObject` also takes a pointer: 2 allocs/op

we can change it to `*Count + toAPIObject` be more idiomatic, the result will be the same.

brandond: Yeah that would be my recommendation, if the change is not too involved.

Internally this is referred to as a comma-ok expression, and when writing one I don't like to see a zero-value struct returned when `!ok` or `err != nil`

https://go.dev/doc/effective_go#maps
> this is called the “comma ok” idiom

https://github.com/golang/go/blob/go1.26.8/src/go/types/expr.go#L1289-L1292
> // If allowCommaOk is set and e is a map index, comma-ok, or comma-err
> // expression, the result is a two-element list containing the value
> // of e, and an untyped bool value or an error value, respectively.

susesamu: done!

pkg/resources/counts/counts.go
susesamu: Not an X-Y problem exactly, but you're right that serializing under the lock is a real alternative,  we could hand down a pre-marshalled `json.RawMessage` and skip the copy entirely.

I benchmarked the two against a random ItemCount:
```
 namespaces                 time   bytes   allocs 
 500         DeepCopy      70 µs   124 KB  120    
 500         json.Marshal  152 µs  34 KB   611    
 2000        DeepCopy      262 µs  507 KB  434    
 2000        json.Marshal  598 µs  171 KB  2413   
```

there's a the 3× byte reduction, but 2x slower and 5x more allocations, WDYT? I think the number of allocs matters more than the bytes, perhaps.

brandond: It takes longer and allocates more when you isolate the benchmark to just this function, right? How does it compare to the complete call stack where the marshal happens later?

I suspect that you're just measuring the additional cost of moving the marshal here, you're not measuring the reduced work that is NOT done elsewhere when the marshal is skipped?

To test this correctly you need to measure the cost of calling the function AND marshalling the return.

susesamu: oh you're right, it's way better. I fixed it!

brandond (approved):

ericpromislow (commented):
Please resolve the conflicts

brandond:
You need to rebase this onto current main, not pull main into it.

pkg/sqlcache/informer/sqlgenerator.go
```diff
@@ -1331,13 +1329,35 @@ func smartJoin(s []string) string {
...
+	if s == "" {
+		return false
+	}
+	for i := 0; i < len(s); i++ {
+		c := s[i]
+		switch {
+		case c == '_':
+		case c >= 'a' && c <= 'z':
+		case c >= 'A' && c <= 'Z':
+		case c >= '0' && c <= '9' && i > 0:
```
ericpromislow: If this is on a critical path, might as well further unroll this so the test for the first character is done before going into the loop

ericpromislow (approved):
It's fine. I would further unroll that loop, but it's up to you.

pkg/sqlcache/informer/sqlgenerator.go
susesamu: done!

brandond (approved):

PR #1353, opened 2026-09-11
Fix: COUNT websocket doesn't notify clients when new CRDs are installed

mallardduck:
## Problem

When users install new CRDs (e.g., Rancher Backup), the COUNT websocket doesn't broadcast updates about the new resource types. This causes false-positive errors in the UI like "You must define a ResourceSet" even though the resources exist. Users must hard refresh to see new resource types.

## Root Cause
COUNT's Watch() built a static `gvkToSchema` map at watch creation. When new CRDs were installed, the `onChange` callback silently  ignored events for unknown GVKs, so websocket clients never learned about new resource types.

## Solution
- Subscribe to schema factory changes and rebuild gvkToSchema when new schemas appear
- Extract countForSchema() helper to ensure consistent initialization and access control logic
- Send full COUNT update to websocket clients when new resource types are detected
- Add defensive nil-map initialization in onChange to handle race conditions

---

Related To:
- https://github.com/rancher/rancher/issues/57250
- https://github.com/rancher/backup-restore-operator/issues/977

### DIFF
--- a/pkg/resources/counts/counts.go
+++ b/pkg/resources/counts/counts.go
@@ -11,6 +11,7 @@ import (
 	"github.com/rancher/steve/pkg/attributes"
 	"github.com/rancher/steve/pkg/clustercache"
 	"github.com/rancher/steve/pkg/resources/ownership"
+	"github.com/rancher/steve/pkg/schema"
 	"github.com/rancher/wrangler/v3/pkg/schemas"
 	"github.com/rancher/wrangler/v3/pkg/summary"
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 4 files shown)


pkg/resources/counts/counts.go
```diff
@@ -142,6 +146,64 @@ func (s *Store) Watch(apiOp *types.APIRequest, schema *types.APISchema, w types.
 		gvkToSchema[attributes.GVK(schema)] = schema
 	}
 
+	// Subscribe to global schema changes, then re-filter for this user
+	if s.schemaFactory != nil {
```
joshmeranda: I think there is only 1 branch what this store is initialized so `s.schemaFactory` cannot be `nil`. Have you observed differently or am I missing something?

joshmeranda (approved):
Seems to do what it says it does. LGTM

PR #1354, opened 2026-09-11
Add RBAC compatibility functions to remove fragile k8s.io dep

mallardduck:


tomleb (approved):

tomleb:
@mallardduck do you get the same issue with r/r btw? Or is it steve only?

mallardduck:
> steve only

Yeah just steve

crobby (approved):

PR #1361, opened 2026-09-15
Index requested cert-manager fields [#56867]

ericpromislow:
Ref [#56867](https://github.com/rancher/rancher/issues/56867)

This does the basic indexing and adds tests on certs and acme challenges


brandond:
Curious how this overlaps with: https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/4358-custom-resource-field-selectors#proposal - CustomResourceFieldSelectors have beens stable since Kubernetes 1.32.

Does Steve already support discovering indexed fields via CRD's `spec.versions.selectableFields`? How do the currently selectable fields for these types overlap with the hardcoded indexes that Steve maintains? 

It seems odd to hardcode all these when the entire point of CRDs is that they can be loaded or unloaded at runtime.

crobby (approved):

tomleb:
@brandond Steve doesn't yet support CRD's selectable fields. We probably _could_ add support for it if we find a need for it. The current steve implementation seems to work good enough. We'd probably still need the steve implementation anyway because we do not own all CRDs (eg: cert-manager) but we still want to index / filter some fields.

brandond:
Have you considered making resource indexes configurable at runtime instead of hardcoding them? I know that handling migration for this might be awkward since you may have to drop the existing table and rebuild the cache when the schema changes... but maintaining static configuration for dynamic resources feels fragile.

tomleb:
@brandond it's been a bit so a bit blurry but I remember having this thought. I don't remember if that's something I discussed with others (eric or even michael though that's pretty far so unlikely). IIRC, ended up going with hardcoded fields because that would have added a lot of work for something the UI didn't need (at least at the time :thinking: ).

In the end, this is used by UI and if they see a need or see this as useful thing to add then we could add it. We could reach out to Richard to check.

tomleb:
> maintaining static configuration for dynamic resources feels fragile.

@brandond on the fragility of it, I don't know, I _think_ it's been working fine for us up to now, I'd be curious to hear Richard's thought on this, though maybe in another discussion channel

ericpromislow:
I'd say supporting `CRD:spec.versions[].selectableFields` would give steve more flexibility - removes a need to add fields to the whitelist, but should be done through a separate issue.

Interesting that the only CRDs that currently set `spec.versions[].selectableFields` are the two cert-related CRDs this PR works with and `certificaterequests.cert-manager.io`. However, they don't specify all the fields that the underlying issue here requested. So supporting `selectableFields` won't buy us much.

brandond:
> However, they don't specify all the fields that the underlying issue here requested.

So the apiserver will support indexed retrieval (fieldSelector queries) on fields that steve won't have indexes for? Is that a problem? Feels like the steve indexes should cover at LEAST the same columns as the apiserver allows field-selecting on - iow, steve indexed columns must be equal to or a superset of selectableFields.

tomleb (approved):

ericpromislow:
> So the apiserver will support indexed retrieval (fieldSelector queries) on fields that steve won't have indexes for? Is that a problem? Feels like the steve indexes should cover at LEAST the same columns as the apiserver allows field-selecting on - iow, steve indexed columns must be equal to or a superset of selectableFields.

I punted on this by pushing it to a still-nonexistent separate issue. But I'm not very knowledgeable on the API server -- the only time I had to work on it was to add new entries to the payload steve returns for the summary work. Where exactly are the indexed fields in the apiserver?

PR #1362, opened 2026-09-18
Ensure certificates and acme challenge state names are in sync with the UI

ericpromislow:
This is because the UI used to do the sorting by state name. It's now done by sqlite in the backend.

Ref issue [#57139](https://github.com/rancher/rancher/issues/57139)


crobby (commented):
Seems reasonable, good tests....I see it's targeted at main...is that intentional?

crobby (approved):
Oh...yep, I was thinking of a different issue.  main is the place.

gehrkefc (approved):

